Issue #10, opened 2023-08-07
Implement Memory Management for VM


rodrigo-pino:

rodrigo-pino:
Closing because al the core functionality is in. Some improvements can still be added, but those will come as separate issues.

PR #8, opened 2023-08-07
feat: Add Relocatable logic

ElijahVlasov:

PR #11, opened 2023-08-08
feat: Add an initial memory layout

rodrigo-pino:
This PR is a first attempt on solving #10 by adding the foundation for memory logic. It defines a MemoryManager, a Segment and a Memory Cell. This definitions are very simple right now, but I believe they will get more complex along with the implementation of the VM instructions.

Also, it expands on @ElijahVlasov works on relocatables (#8) since they are a core part of Memory:
* Renamed Relocatable to MemoryAddress, since I feel this name is more meaningful
* Created a new type `MemoryValue` that is going to work as the joining of all the possible values that a Memory Cell can hold: A `MemoryAddress` and a Field `Element`. I didn't find this solution very go like but we can change it in the future without impacting the rest of the code.


pkg/vm/memory/relocatable.go
```diff
@@ -6,89 +6,162 @@ import (
...
+	return &MemoryAddress{SegmentIndex: segment, Offset: offset}
 }
 
-// Adds two relocatables if the second one is a Felt (relocatable of the form 0:off).
-func (r *MemoryAddress) Add(r1 *MemoryAddress, r2 *MemoryAddress) (*MemoryAddress, error) {
-	if r2.SegmentIndex != feltSegment {
-		return nil, fmt.Errorf("cannot add two relocatables")
-	}
+func (address MemoryAddress) String() string {
+	return fmt.Sprintf("Memory Address: segment: %d, offset: %s", address.SegmentIndex, address.Offset.Text(10))
```
ElijahVlasov: A todo for the future, we might need a nice printing function that output -1 instead of (PRIME - 1), fore example

pkg/vm/memory/relocatable.go
```diff
@@ -6,89 +6,162 @@ import (
...
+	fmt.Println(rhs)
+
+	// todo(rodro): is lhs always a memory address?
+	if lhs.isAddress {
+		memVal.isAddress = true
+		memVal.segmentIndex = lhs.segmentIndex
+	}
+	fmt.Println("x2", lhs.value, rhs.value)
+	memVal.value.Add(lhs.value, rhs.value)
+	fmt.Println("x3")
```
ElijahVlasov: Same here

pkg/vm/memory/relocatable.go
```diff
@@ -6,89 +6,162 @@ import (
...
 	}
 
-	r, err := r.Add((*segmentsOffsets)[r1.SegmentIndex], &MemoryAddress{0, r1.Offset})
+	fmt.Println(memVal)
+	fmt.Println("x0")
+	fmt.Println(lhs)
+	fmt.Println("x1")
+	fmt.Println(rhs)
+
+	// todo(rodro): is lhs always a memory address?
```
ElijahVlasov: I think it should be possible to add felt + memory address too

pkg/vm/memory/relocatable.go
```diff
@@ -6,89 +6,162 @@ import (
...
+		// todo(rodro): is lhs always a memory address?
+		if lhs.isAddress {
+			memVal.isAddress = true
+			memVal.segmentIndex = lhs.segmentIndex
+		}
+		memVal.value.Sub(lhs.value, rhs.value)
+		return memVal, nil
+	}
+
+	// todo(rodro): can lhs not be a memory address?
```
ElijahVlasov: I think all operations should be just lifted felt operations if both lhs and rhs are felts

pkg/vm/memory/memory.go
```diff
@@ -8,24 +8,67 @@ import (
...
-	Value    f.Element
+	Value    *MemoryValue
 	Accessed bool
 }
 
-func (cell *Cell) Read() f.Element {
-	cell.Accessed = true
-	return cell.Value
+type Segment struct {
+	Data map[f.Element]Cell
```
ElijahVlasov: Is there a benefit of using a separate type for segments? Rather than modelling memory as ```map[MemoryAddress]*MemoryValue```?

pkg/vm/vm.go
```diff
@@ -18,16 +18,16 @@ type VirtualMachineConfig struct {
...
-	Config               VirtualMachineConfig
+	Context       Context
+	MemoryManager mem.MemoryManager
+	Config        VirtualMachineConfig
 }
 
 // NewVirtualMachine creates a VM from the program bytecode using a specified config.
 func NewVirtualMachine(programBytecode *[]f.Element, config VirtualMachineConfig) *VirtualMachine {
 	return &VirtualMachine{
 		Context{Fp: 0, Ap: 0, Pc: 0},
```
ElijahVlasov, line 36: Shouldn't Fp, Ap, and Pc be MemoryValues now rather than uints?

pkg/vm/memory/memory_segment.go
```diff
@@ -4,26 +4,25 @@ import (
...
+	// BuiltinSegments *[]MemorySegment
 }
 
-func CreateMemorySegmentManager(programBytecode *[]f.Element) MemorySegmentManager {
-	return MemorySegmentManager{
-		ProgramSegment:   &MemorySegment{Cells: *programBytecode},
-		ExecutionSegment: nil,
-		UserSegment:      nil,
-		BuiltinSegments:  nil,
+func CreateMemoryManager(programBytecode *[]f.Element) MemoryManager {
```
ElijahVlasov: How about changing the argument's name? I mean, the memory manager shouldn't care about bytecode at all. That's out of its abstraction layer.

pkg/vm/memory/memory.go
```diff
@@ -41,17 +58,30 @@ func (segment *Segment) Read(index f.Element) *MemoryValue {
...
-	return Memory{
+func InitializeEmptyMemory() *Memory {
+	return &Memory{
 		// size 4 should be enough for the minimum amount of segments
-		Segments: make([]Segment, 4),
+		Segments: make([]*Segment, 4),
 	}
 }
 
+func (memory *Memory) LoadBytecode(bytecode *[]f.Element) error {
```
ElijahVlasov, line 72: Maybe create a general purpose function that fills a segment with an array of felts?

pkg/vm/memory/memory_segment.go
ElijahVlasov: Memory manager just manages memory, y'know

ElijahVlasov (changes requested):
LGTM! Just some minor suggestions here and there

pkg/vm/memory/memory.go
rodrigo-pino: That's a fair point, though I don't know why it shows you the old code. We are currently using an array insted of a map

pkg/vm/memory/relocatable.go
rodrigo-pino: Yes I agree. Although I wonder if we are going to ever add/substract two memory value that hold felts.

pkg/vm/memory/relocatable.go
rodrigo-pino: Yes we can add it no problem, but before implementing, I'll wonder it will be useful. By implementing the first opcode I am going to have a clearer picture

pkg/vm/vm.go
rodrigo-pino: Should them? I am actually not sure that all three need to be MemoryValues. I think we can easily swap them to Memory Address if we require it.

pkg/vm/memory/memory_segment.go
rodrigo-pino: Yep it makes sense

pkg/vm/memory/memory.go
rodrigo-pino: The reason behind Segment was to make it more readable, like the Segment word make it really obvious what you are dealing with. I think it is a zero cost abstraction as well. But we can change it

pkg/vm/vm.go
ElijahVlasov: At least Fp and Ap should be, because for each function call a separate segment created for it's local memory and Fp and Ap are living in that segment during the execution of the func

ElijahVlasov: Pc is also a MemoryValue but in a dummy way, we know its segment is

pkg/vm/memory/relocatable.go
ElijahVlasov: Since a cell in the memory representaion contains a MemmoryValue we won't know beforehand what value it contains. But for execution of assert opcode we'll have to perfrom arithmetic ops on them. I suggest we hide that logic into MemoryValue

pkg/vm/memory/relocatable.go
ElijahVlasov: Same as above

pkg/vm/vm.go
rodrigo-pino: I am still not sure they need to be `MemoryValues` per se. Following the whitepaper recommendations on memory layout, function calls don't create new segments, only a single segment is used to store the stack trace which includes fp and pc.

`ap` is the allocation pointer, and will always point to free memory. We will always know its position, since it starts at 0, and every write operation increase it (assuming valid CASM). Then I believe that by induction we can prove we will always know its fixed location, since we know it is its previous location (which is fixed) + 1 , hence no need to use a `MemoryAddress`

Implementation wise we can see this as a segment where all the "value" types are stored as well as all the pointers to all reference types (this types getting segments of their own). This segment is going to get filled during a program execution at the same rate `ap` increases.

Since `fp` is just a snapshots of `ap` for everytime there is a function call,  we can say that for transitivity the same proof for `ap` applies to `fp`. So no need for `fp` to be a memory address as well.

ElijahVlasov: The old Python vm used to create a separate segment for each call, but ofc a vm is free to choose any method of segmenting it prefers.

ElijahVlasov (approved):
LGTM!

PR #25, opened 2023-08-22
Small enhancements in tests & codebase

kirugan:


rodrigo-pino (approved):
Great!

pkg/vm/vm_test.go
```diff
@@ -105,7 +102,6 @@ func TestGetApCellOp0(t *testing.T) {
 
 	cell, err := vm.getCellOp0(&instruction)
 	require.NoError(t, err)
-	assert.NotNil(t, cell)
```
kirugan, line 108: Common idiom is use `require.NoError(...)` to check that first value is ok. IMO it's useful only in cases where if `err` is nil first value may be nil as well. Also if it's nil due to a bug - we will know about it through panic.

The main purpose for this particular change - DRY in tests.

PR #29, opened 2023-08-27
Feature: constant time overflow checking while computing offsets

ElijahVlasov:


pkg/safemath/safemath.go
```diff
@@ -0,0 +1,33 @@
...
+	res = x + uint64(y)
+	// Why does this work?
+	// Let's proceed by cases on MSB(x).
+	// If MSB(x) == 1 then we don't have any potential overflow issues
+	// if y < 0 (i.e. MSB(y) == 1), meaning that we can consider the case y >= 0.
+	// In that case we can only wrap up by going to the begining of uint64 range making the MSB(res) = 0.
+	// This is the second disjunct of the DNF.
+	// In the same fashion the case MSB(X) == 0 is reasoned about.
+	// Finally, we boil everything down to MSBs by rotating and anding with ...000001.
+	isOverflow = bits.RotateLeft64((^x&uint64(y)&res)|(x & ^uint64(y) & ^res), 1)&0x1 != 0
```
rodrigo-pino, line 31: since `uint64(y)` is repeated several times, would you consider creating a var containing it?

pkg/safemath/safemath.go
```diff
@@ -0,0 +1,33 @@
...
+// This shows better results because the final bytecode
+// doesn't contain any conditional jump instructions
+// making it easier for a processor to pipeline the function.
+func SafeOffset(x uint64, y int16) (res uint64, isOverflow bool) {
+	// I'll leave proving that this is correct as an exercise for the reader :)
+	res = x + uint64(y)
+	// Why does this work?
+	// Let's proceed by cases on MSB(x).
+	// If MSB(x) == 1 then we don't have any potential overflow issues
+	// if y < 0 (i.e. MSB(y) == 1), meaning that we can consider the case y >= 0.
```
rodrigo-pino, line 26: Nitpick: both comments on line `25` and `26` are related but feel disconnected, could you paraphrase or reorganize, so that is more clear that we only need to analize for overflows when `MSB(x) == 1` and `MSB(y) == 0`

pkg/safemath/safemath.go
```diff
@@ -0,0 +1,33 @@
...
+	// In that case we can only wrap up by going to the begining of uint64 range making the MSB(res) = 0.
+	// This is the second disjunct of the DNF.
+	// In the same fashion the case MSB(X) == 0 is reasoned about.
```
rodrigo-pino, line 29: In line `29`, you are referring to the case when `MSB(x) = 0` and `MSB(y) = 1` if am not mistaken. Could you add that second part.

pkg/safemath/safemath.go
```diff
@@ -0,0 +1,33 @@
+package safemath
+
+import "math/bits"
+
+// This is a constant-time version of the following function:
```
rodrigo-pino, line 5: I would add a description at the beginning. Something along the lines of: "this takes a uint64 to an int16 and outputs their addition as well as the ocurrence of an overflow or underflow"

rodrigo-pino (approved):
Lovely! Awesome work!

pkg/safemath/safemath.go
rodrigo-pino: So readers wouldn't need to dive into the explanation to get some context. Very cool explanation nonetheless it makes all the bit handling makes sense!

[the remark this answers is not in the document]
ElijahVlasov: It's become a part of jargon to call such functions constant-time because they're literally constant-time not asymptotically. On every input they perform the same amount of low-level instructions. This package is an example https://pkg.go.dev/crypto/subtle

PR #32, opened 2023-08-29
Feature: Add a hintrunner for the Vm

### LINKED ISSUE #31: Create a Hint Layout for the VM
Implement a layout that makes easy the addition of new hints. This is beneficial because each hint can be implemented individually, allowing the parallelization of their implementation.

rodrigo-pino:
Added a `Hintrunner` as an external component for the VM. The `Hintrunner` is the part of the project that will run the hints by  which would allow for any of the users of the VM to define their own hints if they want.

This will also favor atomic and parallel works on hints, since all the tools for interacting with the VM for hints has been implemented.

`Alloc` and `TestLessThan` hints are implemented as well as use example.

Fixes #31

### DIFF
--- a/pkg/hintrunner/error.go
+++ b/pkg/hintrunner/error.go
@@ -37,6 +37,7 @@ func (e *HintError) Unwrap() error {
 	return e.err
 }
 
+// OperandError is returned when the error is detected during an operand get/resolve execution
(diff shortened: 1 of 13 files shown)


ElijahVlasov (approved):
LGTM!

PR #34, opened 2023-09-01
Feature: Proof mode for the VM

rodrigo-pino:
To prove correct execution, the cairo-runner must show to the verifier the execution proof. This proof consists of two elements: the execution trace (PC_i, AP_i, FP_i) for each instruction and the resulting memory. To accomplish this we added:
1. Trace gathering while running instructions if `PROOF_MODE` is on.
2. Relocation logic to organize segments as continuos memory.
3. A new method `Proof()` which returns this values


pkg/vm/memory/memory_value.go
```diff
@@ -93,6 +93,12 @@ func (address *MemoryAddress) Sub(lhs *MemoryAddress, rhs any) (*MemoryAddress,
 	}
 }
 
+func (address *MemoryAddress) Relocate(segmentsOffset []int) *f.Element {
+	felt := f.Element{}
+	felt.SetUint64(uint64(segmentsOffset[address.SegmentIndex]) + address.Offset)
```
ElijahVlasov, line 98: Should we check for the overflow here?

pkg/vm/memory/memory_manager.go
```diff
@@ -8,14 +8,45 @@ type MemoryManager struct {
...
-	return nil
+// It returns all segments in memory but relocated as a single segment
+// Each element is a pointer to a field element, if the cell was not accessed,
+// nil is stored instead
+func (mm *MemoryManager) RelocateMemory() []*f.Element {
+	maxMemoryUsed := 0
+	// segmentsOffsets[0] =  0
+	// segmentsOffsets[1] = len(segment[0])
+	// segmentsOffsets[N] = len(segment[n - 1]) + sum of segmentsOffsets[n - i] for i in [0, n-1]
+	segmentsOffsets := make([]int, len(mm.Memory.Segments)+1)
```
ElijahVlasov, line 28: What's the advantage of making the offsets slice ```[]int``` not ```[]uint64```?

pkg/vm/vm.go
```diff
@@ -93,6 +96,11 @@ func (vm *VirtualMachine) RunStep(hintRunner HintRunner) error {
 		return fmt.Errorf("cannot decode step at %d: %w", vm.Context.Pc, err)
 	}
 
+	// store the trace before state change
+	if vm.config.ProofMode {
+		vm.Trace = append(vm.Trace, vm.Context)
```
ElijahVlasov, line 101: What's the cost of appending here? Is resizing cheap or should we make the slice big enough at the beginning?

pkg/vm/vm.go
```diff
@@ -30,46 +30,49 @@ type Context struct {
...
+	manager := mem.CreateMemoryManager()
 	// 0 (programSegment) <- segment where the bytecode is stored
-	_, err = manager.Memory.AllocateSegment(programBytecode)
+	_, err := manager.Memory.AllocateSegment(programBytecode)
 	if err != nil {
 		return nil, fmt.Errorf("error loading bytecode: %w", err)
 	}
-
 	// 1 (executionSegment) <- segment where ap and fp move around
 	manager.Memory.Segments = append(manager.Memory.Segments, mem.EmptySegmentWithCapacity(10))
```
ElijahVlasov, line 57: ```AllocateSegment``` probably?

ElijahVlasov (changes requested):
Great job! Some parts are thought-provoking so left some comments :)

pkg/vm/vm.go
```diff
@@ -162,6 +168,26 @@ func (vm *VirtualMachine) RunInstruction(instruction *Instruction) error {
...
+func (vm *VirtualMachine) Proof() ([]Context, []*f.Element, error) {
+	if !vm.config.ProofMode {
+		return nil, nil, fmt.Errorf("cannot get proof if proof mode is off")
+	}
+
+	// update pc, fp and ap to be their real value
+	// that is, pc_i + 0, ap + len(programBytecode), fp + len(programBytecode)
+	totalBytecode := vm.MemoryManager.Memory.Segments[ProgramSegment].Len()
+	for i := range vm.Trace {
+		vm.Trace[i].Ap += totalBytecode
```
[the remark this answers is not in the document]
rodrigo-pino: So you suggest to convert them to address and then use relocation logic, I don't think is zero cost because the of the extra heap allocations. What if I made a`Relocate` function for the `Context` datatype. It would look like:

```go
for i:= range vm.Trace {
    vm.Trace[i].Relocate()
}
```

pkg/vm/vm.go
rodrigo-pino: Resizing cost scales linearly with the size of the array. For each resize you have to:
* Allocate new memory
* Copy the old memory to the new valory

Different compilers use different heuristics for resizing , the the cost becomes an issue when you have to resize bigger arrays. I think the go rules are the following:
* if smaller than 1024, duplicate it's size
* otherwise, increase its size by 25 percent

In this case is hard to predict if a language will be a short or a long run (the halting problem ofc) and I think that initial cost of resizing is negligible enough at the beginning. That does not stopping from giving it an initial value, but I don't think it would help us much

pkg/vm/memory/memory_value.go
rodrigo-pino: For an overflow to occur here the VM should have run with much more than 2**64 bytes of memory (an exabyte of data 😅) I think we need not to worry about this right now

pkg/vm/memory/memory_manager.go
rodrigo-pino: None, changed it to `[]uint64`

pkg/vm/vm.go
ElijahVlasov: Understood! Let's leave it as is

pkg/vm/memory/memory_value.go
ElijahVlasov: Ok, then I don't see the point of having the felt as a returned value. ```SetUint64``` is not a free operation it involves converting into the Montgomery form under the hood which involves quite a few instructions. Can we get away with not calling it here and returning just ```uint64```?

pkg/vm/vm.go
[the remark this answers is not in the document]
ElijahVlasov: Yeah, could you make a func for ```Context``` relocating then please? It just feels like it wouldn't hurt if that was abstracted

pkg/vm/memory/memory_value.go
rodrigo-pino: If we return uint64, at some point we would have to create a `*f.Element` in order to store it in the final relocated memory. Of course we can create a more efficient memory relocation, but I don't see it without sacrificing a lot of current impl simplicity. Let's have an mvp first ready for benchmarking. That will help us better to understand what is lagging behind, maybe this is not among major issues.

ElijahVlasov: Fair enough!

ElijahVlasov (approved):

PR #38, opened 2023-09-05
Use idiomatic error messages for hint errors

joshklop:


rodrigo-pino (approved):
After some discussion with Josh about the existence of custom error types, we've decided against them for the moment in order to minimize project complexity without sacrificing error description

PR #39, opened 2023-09-05
Organize hintrunner/operand.go

joshklop:
Puts methods after the struct types

### DIFF
--- a/pkg/hintrunner/operand.go
+++ b/pkg/hintrunner/operand.go
@@ -28,8 +28,6 @@ type CellRefer interface {
 
 type ApCellRef int16
 
-type FpCellRef int16
-
 func (ap ApCellRef) Get(vm *VM.VirtualMachine) (*memory.Cell, error) {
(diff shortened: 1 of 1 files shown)


pkg/hintrunner/operand.go
```diff
@@ -133,6 +116,23 @@ func (imm Immediate) Resolve(vm *VM.VirtualMachine) (*memory.MemoryValue, error)
...
+type Operator uint8
+
+const (
+	Add Operator = iota
+	Mul
+)
+
+type ResOperander interface {
+	Resolve(vm *VM.VirtualMachine) (*memory.MemoryValue, error)
+}
```
rodrigo-pino, line 128: I think perhaps `ResOperander` should go on top of Deref, since all subsequent type will implement this interface

rodrigo-pino (approved):
Nice 👌

pkg/hintrunner/operand.go
joshklop: Oop, force-pushed without the change. I can fix this in another PR.

PR #40, opened 2023-09-05
Cairo Zero Parser

krafugo:
# Description

- Add Cairo Zero structs
- Add test cases to check marshaling

# Comment

- Field debug-info is still needed to be completed

### DIFF
--- a/pkg/parsers/zero/zero.go
+++ b/pkg/parsers/zero/zero.go
@@ -2,23 +2,45 @@ package zero
 
 import (
 	"encoding/json"
+	starknetParser "github.com/NethermindEth/cairo-vm-go/pkg/parsers/starknet"
 	"os"
 )
 
-type ZeroProgram struct {
-	Attributes []string          `json:"attributes"`
-	Builtins   map[string]int64  `json:"builtins"`
-	Code       string            `json:"code"`
-	DebugInfo  DebugInfo         `json:"debug_info"`
-	Prime      string            `json:"prime"`
-	Version    string            `json:"version"`
-	Hints      map[string][]Hint `json:"hints"`
+type FlowTrackingData struct {
+	ApTracking   ApTracking        `json:"ap_tracking"`
+	ReferenceIds map[string]uint64 `json:"reference_ids"`
+}
+
+type ApTracking struct {
+	Group  int `json:"group"`
+	Offset int `json:"offset"`
+}
+
+type Location struct {
+	EndCol    uint64            `json:"end_col"`
+	EndLine   uint64            `json:"end_line"`
+	InputFile map[string]string `json:"input_file"`
+	StartCol  uint64            `json:"start_col"`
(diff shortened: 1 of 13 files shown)


pkg/parsers/zero/zero_test.go
```diff
@@ -0,0 +1,105 @@
...
+			expected: ZeroProgram{
+				Hints: map[string][]Hint{
+					"0": {{
+						AccessibleScopes: []string{"starkware.cairo.common.math", "starkware.cairo.common.math.assert_not_equal"},
+						Code:             "from starkware.cairo.lang.vm.relocatable import RelocatableValue\nboth_ints = isinstance(ids.a, int) and isinstance(ids.b, int)\nboth_relocatable = (\n    isinstance(ids.a, RelocatableValue) and isinstance(ids.b, RelocatableValue) and\n    ids.a.segment_index == ids.b.segment_index)\nassert both_ints or both_relocatable, \\\n    f'assert_not_equal failed: non-comparable values: {ids.a}, {ids.b}.'\nassert (ids.a - ids.b) % PRIME != 0, f'assert_not_equal failed: {ids.a} = {ids.b}.'",
+					}},
+				},
+			},
+		},
+		// TODO: Add debug-info test case
```
rodrigo-pino, line 91: 👀 todo

pkg/parsers/zero/zero.go
```diff
@@ -0,0 +1,67 @@
+package zero
+
+import (
+	"encoding/json"
+	"os"
+)
+
+type ZeroProgram struct {
+	Attributes []string          `json:"attributes"`
+	Builtins   map[string]int64  `json:"builtins"`
```
rodrigo-pino, line 10: Could you use the Builtin type we defined for parsers/starknet. This way validation already occurs during parsing

rodrigo-pino (commented):
Great work! Some comments here and there

pkg/parsers/zero/zero_test.go
krafugo: This was just to point out the missing debug-info definition that you told me was not important at the moment. Should I remove this message?

pkg/parsers/zero/zero.go
krafugo: I think they are not the same type. The starknet-parser one is uint8.

rodrigo-pino: Yes, but it has it's own `Builtin` type, with its implementation of a marsharller/unmarshaller. The same can be reused here if I am not mistaken.

pkg/parsers/zero/zero_test.go
rodrigo-pino: Let's have a simple test only

rodrigo-pino (approved):
Structure matches the output of a Cairo Zero compiled program and there are unit tests for each field.

PR #44, opened 2023-09-07
Feat: Implement Casm Assembler

### LINKED ISSUE #43: Implement a CASM Assembler
Testing all sort of bytecode is essential to guarantee the correct behaviour of the VM. Writing bytecode though, is not so easy. We need to create an Assembler so that we could write human readable code and transform it into bytecode

rodrigo-pino:
Adds a Cairo Assembler to the project. Fix #43 

Complete implementation with 92.2% test coverage


pkg/assembler/grammar.go
```diff
@@ -0,0 +1,199 @@
...
+
+// Grammar and AST
+
+type CasmProgram struct {
+	Instructions []Instruction `@@`
+}
+
+type Instruction struct {
+	Core      *CoreInstruction `@@`
+	ApPlusOne bool             `( "," @"ap" "+" "+" )? ";" |`
```
joshklop, line 16: `call` cannot be followed by `ap++` according to the whitepaper (page 35).

One idea to fix this: we can remove `CoreInstruction` and "inline" its fields in `Instruction`. Then, we can use the `( ... )` grouping operator to specify that `ap++` can only follow `AssertEq`, `Jnz`, `Jump`, `ApPlus`, and `Ret`.

pkg/assembler/grammar.go
```diff
@@ -0,0 +1,199 @@
...
+	} else if instruction.Core.Jump != nil {
+		return instruction.Core.Jump
+	} else if instruction.Core.Jnz != nil {
+		return instruction.Core.Jnz
+	} else if instruction.Core.Call != nil {
+		return instruction.Core.Call
+	} else {
+		return instruction.Core.Ret
+	}
+}
```
joshklop, line 109: Since all fields are exported (rightfully so imho), I don't think we gain much from the `CoreInstructioner` and `Expressioner` interfaces. If we remove them and the associated methods, we get to eliminate a lot of code with only minor refactors in `assembler.go`.

I tried it out a bit locally and it seemed to work. Let me know if I'm missing something.

pkg/assembler/grammar.go
joshklop, line 16:
```suggestion
	ApPlusOne bool             `( "," @"ap++" )? ";" |`
```
Not sure if this makes a difference with participle's whitespace handling, but it is a bit easier to read for me.

pkg/assembler/grammar.go
joshklop: This also allows us to simplify other things, like removing the `ApPlus` struct, which is nice.

joshklop (commented):
Reading through the grammar and comparing it to the whitepaper right now. Planning to look at the assembler once I understand the grammar better.

pkg/assembler/grammar.go
rodrigo-pino: Yes, we can definitely get rid of `CoreInstructioner`. I initially implemented it thinking I would use it a lot more. 

The purpose of `Expressioner` is to have both expression rules `Expression` and `DerefOrImmediate` as the same type. I think we can remove it as well by putting `DerefOrImmediate` as a child of `Expression` (while removing its `Deref` and `Immediate` children. I am ok with whatever solution, the current implemented, or an updated one.

The purpose of these interfaces was to hide away all of the pattern matching of the ast, since the assembler shouldn't really care about. I wouldn't like it to have it implemented in the assembly code

pkg/assembler/grammar.go
rodrigo-pino: Yep, thanks missed that. `ApPlus` (ap += <smth>), cannot be followed by an `ap+=1` as well. Also I am going to assume that Call instructions include `Ret` as well.

I am not sure how to use the (...) to solve it, if you could expand on that.

Over the top of my head, the updated grammar would look like this. 

 * Instruction => ApPlusInstruction | NonApPlusInstruction
 
With:

* NonApPlusInstruction => (ApPlus | Call | Ret ) ;

And:

* ApPlusInstruction => (Assert | Jmp) (,ap++ | ;)

I don't think we can inline it all into instruction, without to much repetition or making it very hard to read.
Also, this was an attempt by me to solve the `ap+=1` dilema at a parsing level, but we can also solve it at runtime level, although I wouldn't prefer that since it makes no sense for the instruction

rodrigo-pino: Yes, I initially put it like that, but participle uses the same lexer as go, meaning it detects it treats the tokens as a golang tokens. In this case, the lexer detects both tokens separately instead of together. There is no escaping this without implementing our own lexer

joshklop, line 16: Is `ap++` followed by a comma or a semicolon? The whitepaper seems to use a semicolon; has that changed in recent versions?

joshklop: > ApPlus (ap += smth), cannot be followed by an ap++ as well. 

The whitepaper says, "The only instruction that may not use `ap++` is the `call` instruction."

> I am not sure how to use the (...) to solve it, if you could expand on that.

Something along the lines of:

```go
type Instruction struct {
	AssertEq  *AssertEq `(   @@`
	Jnz       *Jnz      `  | @@`
	Jump      *Jump     `  | @@`
	Ret       bool      `  | @ret`
	ApPlus    *ApPlus   `  | @@`
	ApPlusOne bool      `  ( "," @"ap" "+" "+" ";")? )`
	Call      *Call     `| @@`
}
```

pkg/assembler/grammar.go
```diff
@@ -0,0 +1,199 @@
...
+
+type Offset struct {
+	Sign  string `@("+" | "-")`
+	Value *int   `@Int`
+}
+
+type MathOperation struct {
+	Lhs      *Deref      `@@`
+	Operator string      `@("+" | "*")`
+	Rhs      *DerefOrImm `@@`
```
joshklop, line 81: The whitepaper gives the following cases for `Rhs`:

* Immediate
* `Deref`
* `MathOperation` (with `Deref`s for lhs and rhs)
* `MathOperation` (with `Deref` for lhs and Immediate for rhs)
* `DoubleDeref`

pkg/assembler/grammar.go
joshklop: > DerefOrImmediate as a child of Expression (while removing its Deref and Immediate children.

This sounds interesting. The `As<Struct>` interface methods don't feel quite right to me, but maybe we can leave this comment unresolved for now while we discuss the other items. I need a moment to think about it.

pkg/assembler/assembler.go
```diff
@@ -0,0 +1,274 @@
...
+	opcodeAssertEqBit = 62
+
+	// default values
+	biasedZero     uint16 = 0x8000
+	biasedPlusOne  uint16 = 0x8001
+	biasedMinusOne uint16 = 0x7FFF
+	biasedMinusTwo uint16 = 0x7FFE
+)
+
+func encodeCasmProgram(casmAst CasmProgram) ([]*f.Element, error) {
```
joshklop, line 56: Rather than encoding into bytecode, can we encode into our internal representation using `vm.Instruction`? Then we can easily add all of the test cases mentioned in the whitepaper.

rodrigo-pino: Maybe to another internal representation of `vm.Instruction` yes, but it needs to be another, because the vm will use the assembler as well for testing purposes, and ciruclar dependencies can occurs. If reusing the `vm.Instruction`, then I think we can either move the assembler inside VM or move `Instruction`s to the Assembler. I personally like the latter more than the former, and we can discuss it.

 Also you would need to write this `Instruction` representation into bytecode. I didn't go for it because it meant double the work in the testing part, you need to check that both part of the pipelines are working correctly (assembly -> internal_repr) and (internal_repr -> bytecode). I think it is definetly something we can improve upon

rodrigo-pino: Also, I think at some later stages, if required  even to deprecate the current parser (LL(5)) which is also built the grammar in runtime and go for a compiled grammar and a (LALR(1)) parser, which are faster and more powerful

pkg/assembler/grammar.go
rodrigo-pino: Yep, the Rhs for `AssertEq`. For `MathOperation`, the right hand side is only Deref and Immediate.

pkg/assembler/grammar.go
rodrigo-pino: <expr>,ap++; is the standar in current Casm and Cairo 0.11

rodrigo-pino: Making ret bool makes sense, but I thinks make the code harder to understand.

> The whitepaper says, "The only instruction that may not use ap++ is the call instruction."

The whitepaper is not clear enough sometimes. `ApPlus` is not an instruction per se, in the sense that it does not have an opcode (neither does any of the jumps).

But I am telling that they cannot be used because there is a conflict in the flags that are set: 

* `ApPlus` sets a flag to one
* `ApPlusOne` sets the same flag to two

So clearly, an inconsitency, and they cannot be together. If you check the cairo playground or the python compiler code you'd realize that they error as well in this cases

rodrigo-pino: `ApPlusOne` cannot be used with either `Ret` or `Call` because they forced the `Ap` to certain fixed behavior. And with`ApPlus` is a colition.

rodrigo-pino: Ok, so implemented something similar to your suggestion, completely removing `CoreInstruction` and also `CoreInstructioner` in https://github.com/NethermindEth/cairo-vm-go/pull/44/commits/c90023d6d940a55d9360b974093eaf723377e0c3

joshklop (approved):

pkg/assembler/grammar.go
rodrigo-pino: After further talks, we decided to leave as is for now and perhaps improve it later

PR #45, opened 2023-09-08
Chore: Monadic operations for memory values

ElijahVlasov:
This doesn't pretend to be a significant improvement :) I think it's kind of nice to take an implementation of Either type so that we don't need to implement one on our own. Feel free to discard this ofc! Just a suggestion


pkg/vm/memory/memory_value.go
```diff
@@ -99,40 +100,32 @@ func (address MemoryAddress) String() string {
...
 //
 //   - either a Felt value (an `f.Element`),
 //   - or a pointer to another Memory Cell (a `MemoryAddress`)
-//
-// Both members cannot be non-nil at the same time
-type MemoryValue struct {
-	felt    *f.Element
-	address *MemoryAddress
-}
+type MemoryValue memoryValue
```
rodrigo-pino, line 110: You could directly type:
```
type MemoryValue mo.Either[*f.Element, *MemoryAddress]
```
I think the type `memoryValue` is not used anywhere else

rodrigo-pino (approved):
Cool, when I started working with Go, I craved for some fp. @joshklop wanted even to completely remove the existence of MemoryValue by using an interface, I think that is the right call long term wise. Since this doesn't deter us from that final goal, it is good for me!

pkg/vm/memory/memory_value.go
ElijahVlasov: I use its generated cast  function

rodrigo-pino:
I am closing this since MemoryValue has evolved a lot and it is not compatible with current changes

PR #55, opened 2023-09-20
Test: Add extra tests for getCellDst using negative offsets

mmk-1:
Added two unit tests with negative offsets. One using AP & one using FP


rodrigo-pino (approved):
Lgtm!

rodrigo-pino:
Fixes #51

Issue #54, opened 2023-09-19
Test correct calculation of `Op1` Cell


rodrigo-pino:
`Op1` is one of the four auxiliar values of the Virtual Machine. It can obtained from different sources, check the whitepaper sections four and five to know all different ways of you can get `Op1`. 

Currently just one of the sources [have been implemented](https://github.com/NethermindEth/cairo-vm-go/blob/696d38d05dce882ba4a28ef364641b4ee943ddff/pkg/vm/vm_test.go#L106), where `Op1` cell contains an "Immediate". To fix this issue, you should implement unit tests for the remaining three possible `Op1` sources.

rodrigo-pino:
Fixed in #59

PR #59, opened 2023-09-20
Test: Add tests for every possible Op1 source flag

jmjac:
Add tests for cases with op0, ap and fp. Test both negative and positive offsets. Related to #54


rodrigo-pino (approved):
Great work!

PR #61, opened 2023-09-22
added UpdatePc tests 

mmk-1:
I have added 5 tests for UpdatePc function. I could not add tests with negative offsets because the offset should be uint64. In case there is any way to address this, please point it out and I'll add negative offset test variants too.

Also while I was writing tests, I noticed a potential bug. Based on the Cairo whitepaper, in the case of JNZ (conditional relative jump), the next_pc should be `pc + op1` not `pc + res` , so I fixed that too. If I understood incorrectly please lmk.


rodrigo-pino:
> the next_pc should be pc + op1 not pc + res , so I fixed that too. If I understood incorrectly please lmk.

I think they are the same, because when jzn -> res = op1

rodrigo-pino:
> In case there is any way to address this, please point it out and I'll add negative offset test variants too.

Offsets can be negative, but it is express according to Field Elements for example, -1 would be:
```
x = 3618502788666131213697322783095070105623107215331596699973092056135872020480
```
Because operations with felt are made modulus a prime P where P = x + 1. Then x + 3 = 2, for example

pkg/vm/vm.go
```diff
@@ -377,7 +377,7 @@ func (vm *VirtualMachine) updatePc(
 		}
 
 		// todo(rodro): math check when relAddr is negative
-		relAddr, err := res.Uint64()
+		relAddr, err := op1Cell.Value.Uint64()
```
rodrigo-pino, line 380: You shouldn't access a Cell Value field, if you want to have it you should use the `Read()` method

rodrigo-pino: Also, a bug that was existing previously, is that you cannot extract the value as a uint64 but as a field element

rodrigo-pino:
Don't worry about negative offset nonetheless, I am fixing them as part of a feature I am working on

mmk-1:
> > the next_pc should be pc + op1 not pc + res , so I fixed that too. If I understood incorrectly please lmk.
> 
> I think they are the same, because when jzn -> res = op1

Do you mean in the update res part of the pseudo code res = op1? If so, isn't that when pc_update is case 0,1 or 2?

pkg/vm/vm.go
mmk-1: I changed it to Read()

pkg/vm/vm_test.go
```diff
@@ -424,6 +424,101 @@ func TestUpdatePcNextInstrImm(t *testing.T) {
 	assert.Equal(t, mem.NewMemoryAddress(0, 5), nextPc)
 }
 
+func TestUpdatePcJump(t *testing.T) {
+	vm := defaultVirtualMachine()
+
+	vm.Context.Pc = mem.NewMemoryAddress(0, 3)
+	jumpAddr := uint64(10)
+	res := mem.MemoryValueFromMemoryAddress(mem.NewMemoryAddress(0, jumpAddr))
```
rodrigo-pino, line 432: You could use here `MemoryValueFromSegmentAndOffset`

rodrigo-pino:
> Do you mean in the update res part of the pseudo code res = op1? If so, isn't that when pc_update is case 0,1 or 2?

You were right from the beginning, it was a bug, good catch! I think it should be already patched in `main`

PR #67, opened 2023-09-25
feat: uint256divmod hint

### LINKED ISSUE #46: Feat: Implement Uint256DivMod Hint

coburn24:
WIP will add test, but want to share what I have so far for any feedback
resolves #46

### DIFF
--- a/pkg/hintrunner/hint.go
+++ b/pkg/hintrunner/hint.go
@@ -2,6 +2,7 @@ package hintrunner
 
 import (
 	"fmt"
+	"math/big"
 
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
 	"github.com/NethermindEth/cairo-vm-go/pkg/vm/memory"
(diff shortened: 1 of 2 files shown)


pkg/hintrunner/hint.go
```diff
@@ -87,3 +88,155 @@ func (hint TestLessThan) Execute(vm *VM.VirtualMachine) error {
...
+
+	quotientLow := big.NewInt(1)
+	quotientHigh := big.NewInt(1)
+	quotientLow.And(quotient, mask)
+	quotientHigh.Rsh(quotient, 128)
+
+	remainderLow := big.NewInt(1)
+	remainderHigh := big.NewInt(1)
+	remainderLow.And(remainder, mask)
+	remainderHigh.Rsh(remainder, 128)
```
rodrigo-pino, line 192: Could we perform this operations with F.Element directly. Their implementation is incredibely fast. The performance cost comes form converting to felt and viceversa, because they are stored in Montgomery Form, which is very good to perform operations modulus smth, including division, which is in constant time.

pkg/hintrunner/hint.go
```diff
@@ -87,3 +88,155 @@ func (hint TestLessThan) Execute(vm *VM.VirtualMachine) error {
...
+	quotientHighFelt.SetBigInt(quotientHigh)
+
+	remainderLowFelt := &f.Element{}
+	remainderLowFelt.SetBigInt(remainderLow)
+	remainderHighFelt := &f.Element{}
+	remainderHighFelt.SetBigInt(remainderHigh)
+
+	quotientLowCell, err := hint.Quotient0.Get(vm)
+	if err != nil {
+		return fmt.Errorf("get destination cell: %v", err)
```
rodrigo-pino, line 206: I would modify this error messages to include some info like: "get quotient 0 cell:...". Also I would use the %w format operator, because it is the one specifically used for errors.

This comment applies as well for the following error messages. Do you agree?

rodrigo-pino (commented):
Hey, looking great! Some minor issues

pkg/hintrunner/hint.go
coburn24: yes makes sense, i'll make the edits 👍🏼

pkg/hintrunner/hint.go
```diff
@@ -87,3 +88,155 @@ func (hint TestLessThan) Execute(vm *VM.VirtualMachine) error {
...
+}
+
+func (hint Uint256DivMod) Execute(vm *VM.VirtualMachine) error {
+
+	u128MaxFelt := MaxU128Felt()
+
+	dividend0, err := hint.Dividend0.Resolve(vm)
+	if err != nil {
+		return fmt.Errorf("resolve dividend0 operand %s: %v", hint.Dividend0, err)
+	}
```
coburn24, line 125: @rodrigo-pino replace %v with %w for err here too? is that a standard convention?

pkg/hintrunner/hint.go
coburn24: I'm a little confused on this now, f.Element.Div just returns an f.Element and no remainder
and theres no .And or .Rsh so im not sure how to set low and high, do you have any thoughts on this?

pkg/hintrunner/hint.go
rodrigo-pino: Yes, at least we are following it because %w serves specifically for wrapping errors. %v gives you the default variable valuable. Is think %v is better used when you are not sure with what data type you're dealing with

pkg/hintrunner/hint.go
rodrigo-pino: Yes, it makes sense. The thing is `big.Int` is made for arbitrary big numbers, but we know that for this case our numbers are 256 bits in sized. That info can help us gain a little bit of performance.

Could you use instead this [uint256 library](https://github.com/holiman/uint256). There you could see a benchmarks and it beats go native `big.Int`, also I believe it has all the same methods and more. 
If that is too much hassle, please add a `todo` there saying smth along the lines of _"swap big int usage for a more performant equivalent"_ and I'll put it there later

pkg/hintrunner/hint.go
rodrigo-pino: Was tweaking that part just know and I've noticed that we are using %v originally. That is a mistake on my part and I am already fixing it.

rodrigo-pino:
Hi @coburn24, how are doing, need help with anything?

coburn24:
Hey @rodrigo-pino apologize for my absence, looking to jump back in, it says it was merged? where's a good place to jump back in?

rodrigo-pino:
Hey @coburn24, yes it is still open! There have been some changes to hint runner to support more complex hints such as Dictionaries, but that shouldn't affect the core of your code

cicr99:
Closing this as Uint256DivMod hint was implemented in #180

PR #69, opened 2023-09-26
Lazy felts

ElijahVlasov:
Kinda sloppy PR so review with great attention to details please.


rodrigo-pino (commented):
Yes, I like this idea a lot. The thing is we need to be very performance efficient, one of those ways is not using big ints at all.

The way I had thought implementing by hiding all different types behind the MemoryValue type:
1. Memory Address
2. Consensys Felt
3. Your Lazy Felts (using a single uint64)

This would perhaps then allow it to extend to two slots felts and three slots felt, but I don't know if it would be worthwhile yet.

This way we remove part of the conversions from and to Montgomery form

ElijahVlasov:
> Yes, I like this idea a lot. The thing is we need to be very performance efficient, one of those ways is not using big ints at all.
> 
> The way I had thought implementing by hiding all different types behind the MemoryValue type:
> 
> 1. Memory Address
> 2. Consensys Felt
> 3. Your Lazy Felts (using a single uint64)
> 
> This would perhaps then allow it to extend to two slots felts and three slots felt, but I don't know if it would be worthwhile yet.
> 
> This way we remove part of the conversions from and to Montgomery form

So, my lazy felts don't boost the overall performance significantly in a consistent manner. I'll play around with those on the weekend a little bit more but highly likely I'll close the PR :)

rodrigo-pino:
> So, my lazy felts don't boost the overall performance significantly in a consistent manner. I'll play around with those on the weekend a little bit more but highly likely I'll close the PR :)

yes, it seems it is mostly the Garbage Collector, still swapping from and to montogomery form has some cost that we could  alleviate a little

Issue #70, opened 2023-09-26
Restructuring assembler pipeline


mmk-1:
Currently, the pipeline is like this: AST -> bytecode
The task is to restructure the pipeline to generate instruction from AST then use that instruction to generate bytecode. Meaning:
AST -> instruction -> bytecode

PR #75, opened 2023-09-28
Assembler restructure

mmk-1:
- Renamed Instruction to AST in grammar.go to avoid confusion with Instruction struct in pkg/vm
- Changed the pipeline of bytecode generation by the assembler
- With this PR the pipeline will change to AST -> []Instruction -> bytecode
- Will resolve issue #70 
- Changed the enum mapping in `instruction.go` to match the whitepaper convention.
- Added a few tests to `assembler_test.go`


pkg/assembler/grammar.go
```diff
@@ -8,10 +8,10 @@ import (
 // Grammar and AST
 
 type CasmProgram struct {
-	Instructions []Instruction `@@*`
+	Ast []AstNode `@@*`
```
rodrigo-pino, line 11: `Ast` is a very generic name. Instructions, InstructionList, are more descriptive

pkg/assembler/assembler.go
```diff
@@ -12,145 +12,104 @@ var parser *participle.Parser[CasmProgram] = participle.MustBuild[CasmProgram](
...
+	encodeDst(&node, &instr)
+	encodeOp0(&node, &instr, expr)
+	encodeOp1(&node, &instr, expr)
+	encodeFlags(&node, &instr, expr)
+	return instr, nil
 }
 
-func encodeDstReg(instr *Instruction, encode uint64) (uint64, error) {
-	if instr.ApPlus != nil || instr.Jump != nil {
+func encodeDst(node *AstNode, instr *Instruction) {
```
rodrigo-pino, line 63: I am not fond of prefixing the names with `encode...` since there is no encoding happening in this methods. I would suggest `setInstructionDst` or smth along the lines. This comment applies for the other functions with a similar purpose to this one. Tell me what you think

pkg/vm/vm.go
```diff
@@ -3,6 +3,7 @@ package vm
 import (
 	"fmt"
 
+	assembler "github.com/NethermindEth/cairo-vm-go/pkg/assembler"
```
rodrigo-pino, line 6: `assembler` is a very long name, what do you think of using a shortest one, like `a` or `as`?

pkg/assembler/instruction_test.go
```diff
@@ -22,20 +22,20 @@ func TestDecodeInstructionValues(t *testing.T) {
...
+		Op1Source:   1,
+		Res:         0,
+		PcUpdate:    0,
+		ApUpdate:    0,
+		Opcode:      4,
 	}
 
 	decoded, err := DecodeInstruction(
-		new(f.Element).SetBytes([]byte{0x48, 0x06, 0x80, 0x01, 0x7F, 0xFF, 0x80, 0x00}),
+		new(f.Element).SetBytes([]byte{0x40, 0x07, 0x80, 0x01, 0x7F, 0xFF, 0x80, 0x01}),
```
rodrigo-pino, line 38: Why are there changes in the instruction byte representation?

pkg/assembler/instruction_test.go
```diff
@@ -22,20 +22,20 @@ func TestDecodeInstructionValues(t *testing.T) {
...
-		PcUpdate:    NextInstr,
-		ApUpdate:    Add1,
-		Opcode:      AssertEq,
+		DstRegister: 1,
+		Op0Register: 1,
+		Op1Source:   1,
+		Res:         0,
+		PcUpdate:    0,
+		ApUpdate:    0,
+		Opcode:      4,
```
rodrigo-pino, line 34: Why use numbers instead of the constants already defined?

pkg/assembler/instruction.go
```diff
@@ -161,6 +160,9 @@ type Instruction struct {
 
 	// Defines which instruction needs to be executed
 	Opcode Opcode
+
+	// Immediate value for 2 word instructions
+	Imm string
```
rodrigo-pino, line 165: We don't need an Imm as a field element of an instruction

pkg/assembler/instruction.go
```diff
@@ -65,41 +64,41 @@ func (res ResLogic) String() string {
...
+	PcUpdateJnz       PcUpdate = 4
 )
 
 type ApUpdate uint8
 
 func (ap ApUpdate) String() string {
 	switch ap {
-	case AddImm:
+	case AddRes:
 		return "Add Imm"
```
rodrigo-pino, line 102: Update string representation as well

pkg/assembler/assembler_test.go
```diff
@@ -8,6 +8,97 @@ import (
...
+		OffOp0:      -1,
+		OffOp1:      0,
+		DstRegister: 0,
+		Op0Register: 1,
+		Op1Source:   2,
+		Res:         0,
+		PcUpdate:    0,
+		ApUpdate:    2,
+		Opcode:      4,
+	}
```
rodrigo-pino, line 25: Use the constant values instead of integers please. Apply this for all test cases.  Using constant values instead of numbers makes changes in the code easier, also it is easier for people to read. Like `Op0Register: 1` doesn't same the same as `Op0Register: Fp` for example

rodrigo-pino (commented):
Great work overall! The only thing I don't think is good is the need for an `Immediate` as part of an Instruction field

pkg/assembler/instruction_test.go
mmk-1: I wanted to quickly test a different instruction but forgot to revert back. its the original now

pkg/assembler/assembler.go
mmk-1: yeah I agree. I've updated those functions

pkg/assembler/instruction.go
mmk-1: I changed the way we convert AST to instruction. I made an interface Word that keeps either an Instruction or Immediate (alias for string for better readability). so now `astToInstruction` returns a "wordList" and we can use this to encode it into bytecode in `instruction.go`. What do you think?

rodrigo-pino (approved):
Very good job!!! 🔥

PR #79, opened 2023-09-29
Reduce allocations on VM instruction loop

omerfirmak:
before

```      64	  29953358 ns/op	 5816382 B/op	  240129 allocs/op```

after

```            97          15180877 ns/op        11495353 B/op         43 allocs/op ```

almost 0 GC on the hot path and a %100 speed up. The ones that are still there are segment allocations


rodrigo-pino (approved):
Really awesome work!!!

PR #81, opened 2023-10-02
Added-test-#52

### LINKED ISSUE #52: Test correct calculation of Op0 Cell
Each instruction of the Virtual Machine uses four auxiliar values. One of them is the memory Cell that contains the auxiliar value `Op0`. We need to verify that our current state transition function updates it adequately for any piece of registry, and with all posible offsets.

Shubham1769:
Added test to verify that our current state transition function updates it adequately for any piece of registry, and with all possible offsets.

fixes #52


rodrigo-pino:
Superseded by #97

PR #82, opened 2023-10-03
make felt and address in MemoryValue share the same bytes

omerfirmak:
This is essentially a poor mans union.

``` c
struct {
   union {
       felt f.Element
       memoryAddress MemoryAddress
   }
   bool isFelt
   bool isMemoryAddress
}

```

This results in a smaller `MemoryValue`.
Helps us in 2 ways,

Makes segments take less space in memory (less total memory consumption)
and makes copying MemoryValue instances cheaper since there are less bytes to copy (speed up)

Before;

```
BenchmarkVMRun-16    	Loading program at /home/omer/Documents/cairo-vm-go/fib_compiled.json
Running....
      93	  12089788 ns/op	11495034 B/op	      44 allocs/op
PASS
```

After

```
BenchmarkVMRun-16    	Loading program at /home/omer/Documents/cairo-vm-go/fib_compiled.json
Running....
     100	  11596486 ns/op	 8215796 B/op	      44 allocs/op
PASS
```

a marginal speed up and a significant reduction in total memory allocated.

### DIFF
--- a/pkg/vm/memory/memory_value.go
+++ b/pkg/vm/memory/memory_value.go
@@ -3,6 +3,7 @@ package memory
 import (
 	"errors"
 	"fmt"
+	"unsafe"
 
 	f "github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 	"golang.org/x/exp/constraints"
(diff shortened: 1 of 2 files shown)


pkg/vm/memory/memory_value.go
```diff
@@ -93,16 +94,16 @@ func (address MemoryAddress) String() string {
 //   - or a pointer to another Memory Cell (a `MemoryAddress`)
 type MemoryValue struct {
```
rodrigo-pino, line 95: I would add a comment here explaining that memory address will be contained inside a `felt`

rodrigo-pino (approved):
Lol, went into the review hoping the `union` keyword really existed :)

pkg/vm/memory/memory_value.go
omerfirmak: done

PR #84, opened 2023-10-03
make VM take a memory and initial context

omerfirmak:
will make builtin initialization much easier


rodrigo-pino (approved):
I was thinking of removing the MemoryManager since it current task is only memory relocation, that part could be handled elsewhere.

PR #91, opened 2023-10-04
feat: Add the parsing logic for builtins

jmjac:
This parses the builtins and allocates the space on stack for each of them. There were some issues when I tried to running this with range_check due to the vm trying to overwrite memory cell with a new value.  Now the vm runs through the whole program though not sure if it's all correct.

### DIFF
--- a/pkg/runners/zero/zero.go
+++ b/pkg/runners/zero/zero.go
@@ -55,14 +55,21 @@ func NewRunner(program *Program, proofmode bool, maxsteps uint64) (*ZeroRunner, 
 	}, nil
 }
 
-func (runner *ZeroRunner) InitializeBuiltins() {
+func (runner *ZeroRunner) InitializeBuiltins() error {
 	runner.segments()[VM.ExecutionSegment].IncreaseSegmentSize(uint64(len(runner.program.Builtins)))
 	for i, b := range runner.program.Builtins {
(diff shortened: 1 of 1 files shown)


pkg/vm/vm.go
```diff
@@ -156,11 +156,18 @@ func (vm *VirtualMachine) RunInstruction(instruction *Instruction) error {
...
 		res, err = vm.computeRes(instruction, &op0Addr, &op1Addr)
 		if err != nil {
 			return fmt.Errorf("compute res: %w", err)
 		}
+		if !res.Known() {
+			res, err = vm.Memory.PeekFromAddress(&dstAddr)
+			if err != nil {
+				return fmt.Errorf("compute res: %w", err)
+			}
+		}
```
omerfirmak, line 170: Why do we need this?

pkg/vm/memory/memory.go
```diff
@@ -106,6 +106,8 @@ func (segment *Segment) Read(offset uint64) (MemoryValue, error) {
 		if err := segment.BuiltinRunner.InferValue(segment, offset); err != nil {
 			return MemoryValue{}, err
 		}
+		//Needed to keep the cell as unkown when returned
```
omerfirmak, line 109: Can you explain why?

PR #94, opened 2023-10-04
Refactor AP Update Logic to Handle Negative Values and add tests

jkktom:
Minor change on the AddImm case - Refactor AP Update Logic to Handle Negative Values and add tests
Adds tests for with positive, negative res cases


jkktom:
@rodrigo-pino Please review and let me know if anything more to add.

rodrigo-pino (approved):
Good work!!!

PR #96, opened 2023-10-05
Infer Op1 when Dst is known and ResLogic is Op1

omerfirmak:
this came up while testing RangeCheck


rodrigo-pino (approved):
@jmjac I think that this is related to the issue you were having

PR #98, opened 2023-10-06
Add Pedersen & Bitwise & RangeCheck & Output builtins

omerfirmak:


omerfirmak:
added bitwise builtin as well

integration_tests/cairozero_test.go
```diff
@@ -238,3 +238,23 @@ func memoryRepr(memory []*fp.Element) string {
...
+
+func TestBitwise(t *testing.T) {
+	compiledOutput, err := compileZeroCode("./builtin_tests/bitwise_builtin_test.cairo")
+	require.NoError(t, err)
+
+	_, _, err = runVm(compiledOutput)
+	require.NoError(t, err)
+
+	clean("./builtin_tests/")
+}
```
rodrigo-pino, line 260: Why doing a separate `TestBitwise` instead of including it in `TestCairoZeroFiles`? The best way to make sure it is correct is to compare our trace to the python vm

integration_tests/cairozero_test.go
```diff
@@ -238,3 +238,23 @@ func memoryRepr(memory []*fp.Element) string {
...
 }
+
+func TestFailingRangeCheck(t *testing.T) {
+	compiledOutput, err := compileZeroCode("./failing_contracts/range_check.cairo")
+	require.NoError(t, err)
+
+	_, _, err = runVm(compiledOutput)
+	require.ErrorContains(t, err, "range check builtin failed for value")
+
+	clean("./failing_contracts/")
```
rodrigo-pino, line 249: This more of a nitpick, but I wouldn't call the folder "failing_contracts" because they are not technically contracts but simple Cairo files. It could be confusing

rodrigo-pino (commented):
Great work! Project is moving a lot faster thanks to you!

integration_tests/failing_contracts/range_check.cairo
rodrigo-pino, line 1: Could we add a non failing range check test too?

integration_tests/cairozero_test.go
omerfirmak: Sadly we cant do that because the trace/memory dump python vm generates depends on the enabled builtins. 

We only have range_check and bitwise for now, but python vm doesnt allow you to enable just those two so our trace/memory ends up being different.

I think we can think of proof mode and non-proof mode as two separate  milestones. (ie reach parity on non-proof mode first, then go for proof mode parity.)

rodrigo-pino: I see, I think it is a good idea. Currently will show that there is no failure, I think we also need to verify that we are receiving the expected out. Not to be solved here but something to think about. What would be the best approach for this? Where are return values stored? Do we require the Output builtin 🤔? We need to make it hint independent as well

rodrigo-pino: Perhaps the best place for this type of test is in "pkg/runners/zero/...", this way we can test individual features easily and having access to the runner allows us check results in a better way 🤔 , leaving integration tests as a way to test everything alltogether is correct

rodrigo-pino: After putting some thought into it, I think we can do it in a clean way which doesn't seemed forced. Using a custom `Program`, with our own Assembler for the bytecode and setting the builtins we want by hand. It is easy to unit test each part of the builtin and also easy to check the expected result. Tell me what you think? If you agree, we can leave it as it is for now and I'll do it in a future PR

omerfirmak: > It is easy to unit test each part of the builtin and also easy to check the expected result.

Can you elaborate on how it helps with testing?

omerfirmak: > I think we also need to verify that we are receiving the expected out. Not to be solved here but something to think about. What would be the best approach for this? Where are return values stored? Do we require the Output builtin 🤔? We need to make it hint independent as well

The assumption is that `assert` statements work as expected. So if the program executes fine without throwing an exception, it was executed as expected. Maybe we can add a check for expected number of steps to ensure that it ran to completion?

omerfirmak:
Added pedersen and output builtins as well

integration_tests/cairozero_test.go
rodrigo-pino: Yes, but I don't think this is the best place to do the tests. I think since we are testing that the runner properly initializes and uses Builtins, this tests should be in the "pkg/runners/zero/..." file because it will make testing easier. We will have access to all runtime information. Test would be in the following format, writing it loosely:
```go
func TestStepLimitExceeded(t *testing.T) {
	program := createProgramWithBuiltns(`
           // store in [ap] the ptr to range check
           [ap] = [fp - 2]; 
           [ap + 1] = 3;
           [ap + 1] = [[ap]]; // writing to range check segment
           ... // and so on test what you require
           ret;
      `, ["range_check"])

      runner.Run(program)
      // In here we add the code to test that we have the expected output
```

I don't mind doing this myself, as it will serve as a way to study builtins for me

omerfirmak: Feel free to update the tests as you. 👍

omerfirmak: I dont think we should handwrite the tests in assembly. Because that way we risk tests fitting our implementation and end up with tests that dont really represent the behavior of the code that compiler generates.

rodrigo-pino: I don't understand the argument about test fitting, isn't this argument applicable to all unit tests and isn't this another unit test that verifies that runner is initializing each builtin correctly? 

If tests are faulty, obviously the code is faulty too. Assuming that tests are correct, or that they will be correct because we have the means of checking what is the correct behaviour, then I think these tests as part of the runner pkg is the best place because it is controlled environment. There we can try different combinations and make sure the VM can handle them all.

If our handcrafted assembly is not 100% truthfull to the way compiler generates builtin, it also shouldn't be a problem as long as we respect the theory. A successful run with assemblies obtained from different sources only says that the VM is sturdier in my opinion. It is not a VM for a compiler, but for any compiler that produce valid CASM

omerfirmak: Runner initializes the memory the way it does only because it is how compiler expects it to be initialized. If you write both the runner and the tests, you could easily get them both wrong and not realize that. This goes for every unittest like you said. Using compiler generated testcases instead of hand written CASM, eliminates the possibility of test cases being wrong.

If you are %100 sure the handwritten tests will work the same way as compiler generated CASM, go for it :+1: You can update this PR, I dont mind.

pkg/runners/zero/zero.go
```diff
@@ -210,6 +231,21 @@ func (runner *ZeroRunner) steps() uint64 {
 	return runner.vm.Step
 }
 
+func (runner *ZeroRunner) Output() []*fp.Element {
+	output := []*fp.Element{}
+	for _, segment := range runner.segments() {
+		if segment.BuiltinRunner.String() == "output" {
+			for offset := uint64(0); offset < segment.Len(); offset++ {
+				value := segment.Peek(offset)
+				valueFelt, _ := value.FieldElement()
```
rodrigo-pino, line 240: Is it safe to ignore the error on line 240? Does output can only hold felts?

integration_tests/cairozero_test.go
rodrigo-pino: > Using compiler generated testcases instead of hand written CASM, eliminates the possibility of test cases being wrong

I fully agree. I am not saying to remove your tests. I am saying that I am not fully convinced that they prove correct execution right now. They probably do, but since we are not comparing to another VM and it is hard to compare that two execution are the same when proof mode is off.

Assuming we test each feature with proof mode on, I think there is no need to test without it, since the former is superset of the latter. So I think `proof_mode_correctnes` => `non_proof_mode_correctenes` as long as we don't do anything weird with the code.

At the beginning I said that reaching feature parity without proof mode to be a good approach, but after some thought I am not very keen on it.

I will make the changes in a follow up, since I also want to organize the Zero Runner tests

pkg/runners/zero/zero.go
rodrigo-pino: left a todo. we'll solve all this later

integration_tests/failing_contracts/range_check.cairo
rodrigo-pino: Later on the roadmap

rodrigo-pino (approved):
Great PR, a lot of new stuff added! Some minor issues will be done as follow ups

PR #108, opened 2023-10-10
Fix(assembler): increase parser lookahead

### LINKED ISSUE #104: Issue with parsing specific CASM instruction
Currently the instruction `jmp rel [ap + 1] + [fp - 7];` cannot get parsed and there is a panic. Seems like the "rel" keyword is causing the problem becaues `jmp abs [ap + 1] + [fp - 7];` parses without any problems. 
Based on some early inspections, this could be due to (1) the grammar rules that is missing something or (2) the number of lookahead tokens is not enough in `assembler.go` to disambiguate. There can be another reason for it though.

rodrigo-pino:
Increase parser look ahead so it could select the right production to expand. It couldn't correctly decide between these two instructions:

* jmp rel expr
* jmp rel <dst> if <exrp> != 0

Fixes #104


mmk-1 (approved):
lgtm!

PR #109, opened 2023-10-11
Remove memory manager

### LINKED ISSUE #99: Remove Memory Manager
The `MemoryManager` component only contains relocation logic and just one method `RelocateMemory`. This `RelocateMemory` performs two operations:

1. It gets the relocation offsets of each segment: The relocation offset of the first segment is 0. The relocation offset of the second segment is the length of the first segment. The relocation offset of the ith segment is the sum of the length of the previous segments.
2. Given the relocation offsets and the current memory segments, it returns a continuos representation of the current memory.

This two pieces of component should be divided:

1. The relocation offset calculation should be a method of `Memory`
2. The continuos memory representation should be a `VM` methods

jmjac:
Resolves #99. Not sure about the function names, maybe `relocationOffsets` for memory would be better.


pkg/runners/zero/zero.go
```diff
@@ -14,10 +14,10 @@ import (
...
 type ZeroRunner struct {
 	// core components
-	program       *Program
-	vm            *vm.VirtualMachine
-	hintrunner    hintrunner.HintRunner
-	memoryManager *memory.MemoryManager
+	program    *Program
+	vm         *vm.VirtualMachine
+	hintrunner hintrunner.HintRunner
+	memory     *memory.Memory
```
rodrigo-pino, line 20: No need for this type in the runner any more

rodrigo-pino:
> relocationOffsets

I agree, that name is better 👍

pkg/vm/memory/memory.go
```diff
@@ -257,3 +257,21 @@ func (memory *Memory) Peek(segmentIndex uint64, offset uint64) (MemoryValue, err
 func (memory *Memory) PeekFromAddress(address *MemoryAddress) (MemoryValue, error) {
 	return memory.Peek(address.SegmentIndex, address.Offset)
 }
+
+// It returns all segment offsets and max memory used
+func (memory *Memory) SegmentsOffsets() ([]uint64, uint64) {
+	// this begins at one, because the prover expects for max memory used to
```
rodrigo-pino, line 263: This comment is incomplete 👀

rodrigo-pino (approved):
Good work!

PR #119, opened 2023-10-16
Feat: Add linearSplit hint

### LINKED ISSUE #111: Feat: Implement LinearSplit Hint

jmjac:
Resolves #111


pkg/hintrunner/hint.go
```diff
@@ -141,6 +142,85 @@ func (hint TestLessThanOrEqual) Execute(vm *VM.VirtualMachine) error {
...
+	max_xField, err := max_x.FieldElement()
+	if err != nil {
+		return err
+	}
+	scalarBig := &big.Int{}
+	valueBig := &big.Int{}
+	maxBig := &big.Int{}
+	scalarField.BigInt(scalarBig)
+	valueField.BigInt(valueBig)
+	max_xField.BigInt(maxBig)
```
rodrigo-pino, line 184: Why do you use bigint for the operation instead of felts?

pkg/hintrunner/hint.go
```diff
@@ -141,6 +142,85 @@ func (hint TestLessThanOrEqual) Execute(vm *VM.VirtualMachine) error {
 	return nil
 }
 
+type LinearSplit struct {
+	value  ResOperander
+	scalar ResOperander
+	max_x  ResOperander
```
rodrigo-pino, line 148: camelCase

pkg/hintrunner/hint.go
rodrigo-pino: I am sorry but I still don't understand the need for a big int. I don't see what is stopping your felts from doing the following. What we want to do is to find `x` and `y` such that `x * scalar + y = value`.

For `x` we get it by calculating `min(value / scalar, minX)`. Since `value` is a felt252 then the result of it is a felt252. No need for big int here.

For `y` we get it by calculating `value - scalar * x`. The only thing that could go out of range is `scalar * x` but it actually doesn't. The reason behind this is that if:

* `x` is `value / scalar` then `scalar * x` is `value`, which we already know is a `felt252`. No need for big int.
* `x` is `maxX`. We know `maxX` is smaller than `value / scalar` and since `value / scalar` fits inside a `felt252` then `maxX` fits inside one as well.

Please tell me if there is any flaw on my logic. Code would look like this:
```go
// for xValue: xValue = min(maxX, value / scalar)
xValue := f.Element{}
xValue.Div(value, scalar)
if xValue.Eq(maxX) == 1 {
  xValue = *maxX
}

// for yValue = value - xValue * scalar
yValue := f.Element{}
yValue.Mul(xValue, scalar)
yValue.Sub(value, yValue)

// then proceed and store this values
```

jmjac: If we take `value=223344*42+14` and different `maxX` we get the following:
```
With Felt Div:
        value=9380462
        scalar=42
        maxX=9999999999
        x=9999999999
        y=3618502788666131213697322783095070105623107215331596699973092055715881400985
With BigInt Div:
        value=9380462
        scalar=42
        maxX=9999999999
        x=223344
        y=14

With Felt Div:
        value=9380462
        scalar=42
        maxX=223343
        x=223343
        y=56
With BigInt Div:
        value=9380462
        scalar=42
        maxX=223343
        x=223343
        y=56
```
then depending on `maxX` we get different values for `y` and `x` if we use felt division or bigInt division

rodrigo-pino: Ok, yep, let's keep the uint256 behaviour. 

For the record, the difference is due to felt division giving working different than normal division, giving very big values when there is no integer division

rodrigo-pino:
Please include benchmark tests for them as well. Update from main to get the latest update on the hint pkg.

rodrigo-pino (approved):

PR #123, opened 2023-10-17
SquareRoot hint

mmk-1:
This PR adds SquareRoot hint.
Resolves issue #110


pkg/hintrunner/hint.go
```diff
@@ -215,3 +215,49 @@ func (hint WideMul128) Execute(vm *VM.VirtualMachine) error {
...
+	valueFelt, err := value.FieldElement()
+	if err != nil {
+		return err
+	}
+
+	valueU256 := uint256.Int(valueFelt.Bits())
+
+	if valueU256.Gt(&mask) {
+		return fmt.Errorf("lhs operand %s should be u128", valueFelt)
+	}
```
rodrigo-pino, line 244: I don't understand why you need to compare if the value is smaller than u128. Also why do you convert to u256 instead of handling only with felts?

pkg/hintrunner/hint.go
```diff
@@ -215,3 +215,40 @@ func (hint WideMul128) Execute(vm *VM.VirtualMachine) error {
...
+	sqrt := valueFelt.Sqrt(valueFelt)
+
+	dst := f.Element{}
+	dst.Set(sqrt)
+
+	dstAddr, err := hint.dst.Get(vm)
+	if err != nil {
+		return fmt.Errorf("get destination cell: %v", err)
+	}
+	dstVal := memory.MemoryValueFromFieldElement(&dst)
```
rodrigo-pino, line 248: Why do you create a `dst` which is equal to `sqrt` felt and then store `dst` without any further changes. Why not storing `sqrt` directly?

rodrigo-pino (approved):

PR #125, opened 2023-10-18
Tests: Add unit tests for Cairo Zero using Builtins

rodrigo-pino:


pkg/vm/builtins/bitwise.go
```diff
@@ -26,14 +31,14 @@ func (b *Bitwise) InferValue(segment *memory.Segment, offset uint64) error {
 	xOffset := offset - bitwiseIndex
 	yOffset := xOffset + 1
 
-	xValue, err := segment.Read(xOffset)
-	if err != nil {
-		return err
+	xValue := segment.Peek(xOffset)
```
jmjac, line 34: Why the change from `Read` to `Peek`?

rodrigo-pino: To give a more helpful error message, if not the error message looks like:
"bitwise: bitwise: error_msg"

jmjac (approved):
Looks good

PR #127, opened 2023-10-18
added more tests for TestAssertEqualInstruction

Tomi-3-0:
Added more test cases for the TestAssertEqualInstruction function

### DIFF
--- a/pkg/vm/vm_test.go
+++ b/pkg/vm/vm_test.go
@@ -918,6 +918,78 @@ func TestAssertEqualInstruction(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, mem.MemoryValueFromInt(-5), mv)
 	})
+
+	t.Run("multiply", func(t *testing.T){
+		vm:= defaultVirtualMachineWithCode("[ap] = [ap - 1] * [fp];")
+		setInitialReg(vm, 3, 1, 0)
+
+		writeToDataSegment(vm, vm.Context.Fp, 5)
+		writeToDataSegment(vm, vm.Context.Ap-1, 10)
+
+		err:= vm.RunStep(&hintrunner)
+		require.NoError(t, err)
+
+		mv, err := vm.Memory.Read(ExecutionSegment, vm.Context.Ap)
+		require.NoError(t, err)
+		assert.Equal(t, mem.MemoryValueFromInt(50), mv)
+	})
+
+	t.Run("comparison equality", func(t *testing.T){
+		vm:= defaultVirtualMachineWithCode("[ap + 1] = [fp];")
+		setInitialReg(vm, 3, 1, 0)
+
+		writeToDataSegment(vm, vm.Context.Ap+1, 5)
+
+		err:= vm.RunStep(&hintrunner)
+		require.NoError(t, err)
+
+		mv, err := vm.Memory.Read(ExecutionSegment, vm.Context.Fp)
+		require.NoError(t, err)
+		assert.Equal(t, mem.MemoryValueFromInt(5), mv)
+	})
+
+	t.Run("comparison equality 2", func(t *testing.T){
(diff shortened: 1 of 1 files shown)


pkg/vm/vm_test.go
```diff
@@ -918,6 +918,78 @@ func TestAssertEqualInstruction(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, mem.MemoryValueFromInt(-5), mv)
 	})
+
+	t.Run("multiply", func(t *testing.T){
```
rodrigo-pino, line 922: Could you add one for division as well?

pkg/vm/vm_test.go
```diff
@@ -918,6 +918,78 @@ func TestAssertEqualInstruction(t *testing.T) {
...
+		err:= vm.RunStep(&hintrunner)
+		require.NoError(t, err)
+
+		mv, err := vm.Memory.Read(ExecutionSegment, vm.Context.Ap)
+		require.NoError(t, err)
+		assert.Equal(t, mem.MemoryValueFromInt(50), mv)
+	})
+
+	t.Run("comparison equality", func(t *testing.T){
+		vm:= defaultVirtualMachineWithCode("[ap + 1] = [fp];")
```
rodrigo-pino, line 938: There is already a test that does this

pkg/vm/vm_test.go
```diff
@@ -918,6 +918,78 @@ func TestAssertEqualInstruction(t *testing.T) {
...
+		err:= vm.RunStep(&hintrunner)
+		require.NoError(t, err)
+
+		mv, err := vm.Memory.Read(ExecutionSegment, vm.Context.Fp)
+		require.NoError(t, err)
+		assert.Equal(t, mem.MemoryValueFromInt(5), mv)
+	})
+
+	t.Run("comparison equality 2", func(t *testing.T){
+		vm:= defaultVirtualMachineWithCode("[ap - 1] = [fp];")
```
rodrigo-pino, line 952: There is already a test that does this

pkg/vm/vm_test.go
```diff
@@ -918,6 +918,78 @@ func TestAssertEqualInstruction(t *testing.T) {
...
+
+		err:= vm.RunStep(&hintrunner)
+		require.NoError(t, err)
+
+		mv, err := vm.Memory.Read(ExecutionSegment, vm.Context.Fp)
+		require.NoError(t, err)
+		assert.Equal(t, mem.MemoryValueFromInt(7), mv)
+	})
+
+	t.Run("compare register values", func(t *testing.T){
```
rodrigo-pino, line 965: There is already a test that does this

pkg/vm/vm_test.go
```diff
@@ -918,6 +918,78 @@ func TestAssertEqualInstruction(t *testing.T) {
...
+
+		err:= vm.RunStep(&hintrunner)
+		require.NoError(t, err)
+
+		mv, err := vm.Memory.Read(ExecutionSegment, vm.Context.Fp)
+		require.NoError(t, err)
+		assert.Equal(t, mem.MemoryValueFromInt(3), mv)
+	})
+
+	t.Run("compare register values left to right", func(t *testing.T){
```
rodrigo-pino, line 979: There is no need for this test since there is already others

rodrigo-pino (commented):
There are tests with considerable overlap. There is no need to test ap or fp at different sides of assert equal since does individual components are already tested in previous tests like `GetDstCell0` and `GetDstCell1`. Remember that what we are looking for here is to verify that all ways of building an equal instruction works as expected.

pkg/vm/vm_test.go
Tomi-3-0: Noted

rodrigo-pino:
Closing, follow up at #133

PR #129, opened 2023-10-19
Feat: debugPrint hint

### LINKED ISSUE #124: Feat: Add DebugPrint hint

jmjac:
Resolves #124.  As I understand this hint handles printing values between `start` and `end` addresses. I don't like getting underlying type of `ResOperand` but this seems to be necessary to figure out how many values do we need to print. Also this won't work with `Immediate` because it's a value not an address

### DIFF
--- a/pkg/hintrunner/hint.go
+++ b/pkg/hintrunner/hint.go
@@ -215,3 +215,46 @@ func (hint WideMul128) Execute(vm *VM.VirtualMachine) error {
 
 	return nil
 }
+
+type DebugPrint struct {
+	start ResOperander
+	end   ResOperander
+}
+
+func (hint DebugPrint) Execute(vm *VM.VirtualMachine) error {
+	start, err := hint.start.Resolve(vm)
+	if err != nil {
+		return fmt.Errorf("resolve start operand %s: %v", hint.start, err)
+	}
+
+	startAddr, err := start.MemoryAddress()
+	if err != nil {
+		return fmt.Errorf("start memory address: %v", err)
+	}
+
+	end, err := hint.end.Resolve(vm)
+	if err != nil {
+		return fmt.Errorf("resolve rhs operand %s: %v", hint.end, err)
+	}
+	endAddr, err := end.MemoryAddress()
+	if err != nil {
+		return fmt.Errorf("end memory address: %v", err)
+	}
+
+	if startAddr.Offset > endAddr.Offset {
+		return fmt.Errorf("start cannot be greater than end")
+	}
+
+	current := startAddr.Offset
+	for current < endAddr.Offset {
+		fmt.Printf("[DEBUG] %x\n", current)
+		current += 1
+	}
+	//v, err := vm.Memory.ReadFromAddress(startAddr)
+
+	//if err != nil {
+	//	return fmt.Errorf("read memory address %s: %v", err)
+	//}
+
+	return nil
+}
--- a/pkg/hintrunner/hint.go
+++ b/pkg/hintrunner/hint.go
@@ -253,13 +253,9 @@ func (hint DebugPrint) Execute(vm *VM.VirtualMachine) error {
 			return err
 		}
 
-		fmt.Printf("[DEBUG] %s\n", v)
+		fmt.Printf("[DEBUG] %s\n", v.Hex())
 		current += 1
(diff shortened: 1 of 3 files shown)


pkg/hintrunner/hint.go
```diff
@@ -215,3 +215,47 @@ func (hint WideMul128) Execute(vm *VM.VirtualMachine) error {
...
+		return fmt.Errorf("start has to be a pointer")
+	}
+
+	var endAddr memory.MemoryAddress
+	switch v := hint.end.(type) {
+	case Deref:
+		endAddr, _ = v.deref.Get(vm)
+	default:
+		return fmt.Errorf("end has to be a pointer")
+	}
```
rodrigo-pino, line 239: This is not right. You need to use Resolve. It will return a MemoryValue which can be either a Felt or an address. You then extract an address from both of them or error instead

pkg/hintrunner/hint_test.go
```diff
@@ -231,3 +234,38 @@ func TestWideMul128IncorrectRange(t *testing.T) {
...
+	err := hint.Execute(vm)
+
+	w.Close()
+	out, _ := io.ReadAll(r)
+	fmt.Println(string(out))
+	//Restore stdout at the end of the test
+	os.Stdout = rescueStdout
+
+	require.NoError(t, err)
+	require.Equal(t, out, expected)
```
rodrigo-pino, line 270:
```suggestion
        require.Equal(t, expected, out)
```
Expected goes in the right

pkg/vm/memory/memory_value.go
```diff
@@ -88,6 +88,12 @@ func (address MemoryAddress) String() string {
 	)
 }
 
+func (address MemoryAddress) Hex() string {
+	return fmt.Sprintf(
+		"%x:%x", address.SegmentIndex, address.Offset,
+	)
+}
+
```
rodrigo-pino, line 96: No need for a Hex method for memory address

pkg/vm/memory/memory_value.go
```diff
@@ -275,6 +281,13 @@ func (mv MemoryValue) String() string {
 	return mv.felt.String()
 }
 
+func (mv MemoryValue) Hex() string {
+	if mv.IsAddress() {
+		return fmt.Sprintf("%x", mv.addrUnsafe())
+	}
+	return mv.felt.Text(16)
+}
```
rodrigo-pino, line 289: No need for hex method for memory value

pkg/hintrunner/hint.go
```diff
@@ -215,3 +215,47 @@ func (hint WideMul128) Execute(vm *VM.VirtualMachine) error {
...
+		return fmt.Errorf("start cannot be greater than end")
+	}
+
+	current := startAddr.Offset
+
+	for current < endAddr.Offset {
+		v, err := vm.Memory.ReadFromAddress(&memory.MemoryAddress{
+			SegmentIndex: startAddr.SegmentIndex,
+			Offset:       current,
+		})
```
rodrigo-pino, line 251: After reading some, ouptut segment only contains felts, so no need to add hex method to anyone. You can assume your extracting a field element

jmjac: Okay, removed both and used `felt.Text(16)` to it in hex format

pkg/hintrunner/hint_test.go
jmjac: Fixed

pkg/hintrunner/hint.go
jmjac: Makes sense, fixed

pkg/hintrunner/hint_test.go
```diff
@@ -232,11 +234,46 @@ func TestWideMul128IncorrectRange(t *testing.T) {
 	require.ErrorContains(t, err, "should be u128")
 }
 
-func TestSquareRoot(t *testing.T) {
+func TestDebugPrint(t *testing.T) {
```
rodrigo-pino, line 237: The test should use the Output builtin

rodrigo-pino:
Please include benchmark tests for them as well. Update from main to get the latest update on the hint pkg.

rodrigo-pino (approved):

Issue #118, opened 2023-10-16
Test that any type of `[smth] = <smth>` Instruction is correct.


rodrigo-pino:
We should test that run step works flawlessly for any form of `EqualAssertion` instruction. There are already some tests implemented. We should add the remaining ones.

The implementation should be done in the following section:
https://github.com/NethermindEth/cairo-vm-go/blob/8ad5c06813580e736222d830d6d483e1eefa4924/pkg/vm/vm_test.go#L856-L921

PR #133, opened 2023-10-22
More tests for assert equal Instruction

Tomi-3-0:
A test for an  instruction that checks if two values are equal. It is represented by the following syntax:
`<left_hand_op> = <right_hand_op>`
It ensures that both sides are equal and rejects the program execution otherwise.
For example: 
` [reg0 + offop0] ◦ imm`

where  `◦`  is an arithmetic operator (addition or multiplication), and `imm` is any fixed field element. 
Division and subtraction are represented as multiplication and addition (respectively) with a different order of operands.
Issue #118


rodrigo-pino (approved):
Looks good to me!

PR #134, opened 2023-10-23
Implement Uint256SquareRoot

mmk-1:
resolves issue #131


pkg/hintrunner/hint.go
```diff
@@ -250,3 +250,161 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+	valueHighFelt, err := valueHigh.FieldElement()
+	if err != nil {
+		return err
+	}
+
+	// value = {value_low} + {value_high} * 2**128
+	valueLowU256 := uint256.Int(valueLowFelt.Bits())
+	valueHighU256 := uint256.Int(valueHighFelt.Bits())
+	valueHighU256.Lsh(&valueHighU256, 128)
+	value := valueHighU256.Add(&valueHighU256, &valueLowU256)
```
rodrigo-pino, line 293: Through normally I use this pattern myself but since performance is a primary objective I think we could save some memory here and reuse one of the u256 variable

pkg/hintrunner/hint.go
```diff
@@ -250,3 +250,161 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+
+	// root = math.isqrt(value)
+	root := value.Clone()
```
rodrigo-pino, line 296: root can be an empty var, no need to hold anything

pkg/hintrunner/hint.go
```diff
@@ -250,3 +250,161 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+	root.Sqrt(value)
+
+	// remainder = value - root ** 2
+	root2 := root.Clone()
```
rodrigo-pino, line 300: root2 can be an empty var as well

pkg/hintrunner/hint.go
```diff
@@ -250,3 +250,161 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+	root2.Mul(root, root)
+	remainder := value.Clone()
```
rodrigo-pino, line 302: reamainder can be an empty value as well

rodrigo-pino: I'll stop signaling since the patter repeats further down below. It hurts readability initializing variables with values they are not going to use

pkg/hintrunner/hint.go
```diff
@@ -250,3 +250,161 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+	// memory{sqrt0} = root & 0xFFFFFFFFFFFFFFFF
+	// memory{sqrt1} = root >> 64
+	mask64 := uint256.NewInt(0xFFFFFFFFFFFFFFFF)
+	rootMasked := root.Clone()
+	rootMasked.And(root, mask64)
+	rootShifted := root.Rsh(root, 64)
+
+	sqrt0 := f.Element{}
+	sqrt0.SetBytes(rootMasked.Bytes())
+	fmt.Println(rootMasked.Dec())
```
rodrigo-pino, line 314: rogue print

pkg/hintrunner/hint.go
```diff
@@ -250,3 +250,161 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+
+	sqrt1 := f.Element{}
+	sqrt1.SetBytes(rootShifted.Bytes())
+	fmt.Println(rootShifted.Dec())
```
rodrigo-pino, line 318: rogue print

rodrigo-pino (commented):
Initial review. It looks good, just some minor issues

rodrigo-pino:
Please include benchmark tests for them as well. There is a dedicated file

pkg/hintrunner/hint.go
jmjac: You should be able to do 
```
var root = uint256.Int{}
root.Sqrt(&value)
```
What memory errors are you getting?

pkg/hintrunner/hint.go
```diff
@@ -376,3 +376,159 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+
+	// memory{sqrt_mul_2_minus_remainder_ge_u128} = root * 2 - remainder >= 2**128
+	rootMul2 := uint256.Int{}
+	rootMul2.Lsh(&root, 1)
+	lhs := uint256.Int{}
+	lhs.Sub(&rootMul2, &remainder)
+
+	rhs := uint256.NewInt(1)
+	rhs.Lsh(rhs, 128)
+	// rhs.Mul(mask128, mask128)
```
rodrigo-pino, line 511: Does this commented line has any purpose?

pkg/hintrunner/hint.go
```diff
@@ -376,3 +376,159 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+	result := rhs.Gt(&lhs)
+	result = !result
+
+	sqrtMul2MinusRemainderGeU128 := f.Element{}
+	if result {
+		sqrtMul2MinusRemainderGeU128.SetOne()
+	} else {
+		sqrtMul2MinusRemainderGeU128.SetZero()
+	}
```
rodrigo-pino, line 520: No need to set to zero since default value for an uninitialized Element is zero

rodrigo-pino:
@mmk-1 please fix conflicts so we can merge

rodrigo-pino (approved):

Issue #117, opened 2023-10-16
Full tests of VM `RunStep`


rodrigo-pino:
Currently, we test that each sub component of the transition function is correct, but we haven't tested them as a whole. We should add the following tests for `RunStep`:
- [x] #118
- [x] Test `ap += <smth>` is correct.
- [x] Test `jump abs` and `jump rel` jumps are correct.
- [x] Test conditional jumps `jnz` are correct.
- [x] Test `call rel`, `call abs` and `ret` are correct.

rodrigo-pino:
Fixed in #152

PR #141, opened 2023-10-25
Test for unconditional jump instructions

Tomi-3-0:
Tests for relative jumps (where the operand represents an offset
from the current pc) and absolute jumps – represented by the keywords `rel`
and `abs`, respectively

Unconditional jumps:
```
jmp abs <address>
jmp rel <offset>
```
#117

### DIFF
--- a/pkg/vm/vm_test.go
+++ b/pkg/vm/vm_test.go
@@ -960,6 +960,48 @@ func TestAssertEqualInstruction(t *testing.T) {
 
 }
 
+// add test for each different form of jump instructions
+func TestJumpInstructions(t *testing.T) {
+
+	// relative jumps
+	t.Run("jmp rel 5", func(t *testing.T){
+		vm := defaultVirtualMachine()
+
+		vm.Context.Pc = mem.MemoryAddress{SegmentIndex: 1, Offset: 4}
+		relAddr := uint64(5)
+		res := mem.MemoryValueFromInt(relAddr)
+
+		instruction := a.Instruction{
+			PcUpdate: a.PcUpdateJumpRel,
+		}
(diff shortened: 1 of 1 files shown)


pkg/vm/vm_test.go
```diff
@@ -927,8 +927,81 @@ func TestAssertEqualInstruction(t *testing.T) {
...
 }
 
+// add test for each different form of jump instructions
+func TestJumpInstructions(t *testing.T) {
+
+	// relative jumps
+	t.Run("jmp rel", func(t *testing.T){
+		vm := defaultVirtualMachineWithCode("jmp rel [ap + 1] + [ap - 7];")
+
+		vm.Context.Pc = mem.MemoryAddress{SegmentIndex: 1, Offset: 3}
```
rodrigo-pino, line 970: pc should always be on segment index 0

pkg/vm/vm_test.go
```diff
@@ -927,8 +927,81 @@ func TestAssertEqualInstruction(t *testing.T) {
...
+		relAddr := uint64(10)
+		res := mem.MemoryValueFromInt(relAddr)
+
+		instruction := a.Instruction{
+			PcUpdate: a.PcUpdateJumpRel,
+		}
+		nextPc, err := vm.updatePc(&instruction, nil, nil, &res)
+
+		require.NoError(t, err)
+		assert.Equal(t, mem.MemoryAddress{SegmentIndex: 1, Offset: 3 + relAddr}, nextPc)
```
rodrigo-pino, line 980: The test body is not correct and does not match the behaviour of the previous test of the same kind

pkg/vm/vm_test.go
Tomi-3-0: > pc should always be on segment index 0

Corrected

Issue #130, opened 2023-10-19
Feat: Keccak Builtin


jkktom:
#126 Implementing Keccak Builtin in Cairo Go VM
What is Keccak

- **Purpose**: Keccak is the function selected as the SHA-3 (Secure Hash Algorithm 3) standard by NIST (National Institute of Standards and Technology). Keccak is a one-way cryptographic functions that take an input and produce a fixed-size string of characters, typically a sequence of numbers. The Go library ( [golang.org/x/crypto/sha3](url)) provides the NewLegacyKeccak256 function to create a Keccak-256 hasher, which you can use to compute Keccak hashes in Go.

- **Common Use**: Hash functions are used in various applications, including data integrity verification, password hashing, and digital signatures. In blockchain, they are used to create addresses, form blocks, and maintain the integrity and order of data.

- Constants : Constants like `KECCAK_FULL_RATE_IN_BYTES` (136), `KECCAK_FULL_RATE_IN_U64S` (17), and `BYTES_IN_U64_WORD` are declared. **136** : The value `136` bytes equates to `1088` bits (`136 * 8 = 1088` bits). **8**: `136 / 8 = 17`. Hence, `KECCAK_FULL_RATE_IN_U64S` is `17`.

- **Padding before Hashing** `add_padding(ref input, last_input_word, last_input_num_bytes);`

In a hash function, data is processed in blocks of a fixed size. If a block of data is not full, padding is added to fill the block, much like adding book partitions to stabilize and fill out a shelf. This ensures that each block is fully occupied and ready for processing, which in the hash function's case, means ready for hashing.

The padding, like the book partitions, creates a standard organization system. However, unlike the bookshelf scenario, the padding in a hash function also serves other critical purposes such as ensuring unique hash outputs for different inputs, and often encoding the length of the original message to prevent certain types of cryptographic attacks.

- LE (Little Endian) & BE (Big Endian)
0x12345678 would be stored in memory as 78 56 34 12 in LE , 12 34 56 78. in BE
**Different ways** of reading numbers, much like how different languages read text from left to right or right to left
When serializing data to be stored or transmitted, or **deserializing** it back into memory, the endianness needs to be handled correctly to ensure data integrity.
Being able to handle or convert between endianness is _important for cross-platform_ **compatibility**.
It's crucial that all parties agree on the byte order, so data is interpreted correctly.

PR #144, opened 2023-10-26
Builtin keccak

jkktom:
Draft PR for Builtin - Keccak256
Done
[x] add padding
[x] applied keccak as CairoKeccak
[x] test result from the web identical (verified)
[x] LE, BE (Little Endian, Big Endian) way of Keccak Created as Rust VM does
[ ] test for LE, BE Keccak

Side Note : in the Cairo Repo, there's "starknet_keccak" in Rust code. Basically doing the same thing. location : https://github.com/cairo/crates/cairo-lang-starknet/src/contract.rs

### DIFF
--- a/pkg/vm/builtins/keccak.go
+++ b/pkg/vm/builtins/keccak.go
@@ -13,6 +13,14 @@ const (
 	BYTES_IN_U64_WORD         = 8
 )
 
+type uint128 struct {
+	High, Low uint64
+}
+
+type U256 struct {
+	High, Low uint128
+}
+
 // u128ToU64 converts a big.Int (representing a 128-bit integer) to a uint64,
 // assuming the big.Int fits into a uint64.
 func U128ToU64(input *big.Int) (uint64, error) {
@@ -36,6 +44,20 @@ func U128Split(input *big.Int) (high, low uint64, err error) {
 		return 0, 0, err
 	}
 	return high, low, nil
+}
+
+// SplitU256 splits a U256 into four uint64s.
+func SplitU256(value U256) (lowLow, lowHigh, highLow, highHigh uint64) {
+	lowLow, lowHigh = value.Low.Low, value.Low.High
+	highLow, highHigh = value.High.Low, value.High.High
+	return
+}
+
+// AppendU256ToSlice appends the values from a U256 to a slice in the specified order.
+func AppendU256ToSlice(slice []uint64, value U256) []uint64 {
+	lowLow, lowHigh, highLow, highHigh := SplitU256(value)
+	slice = append(slice, lowLow, lowHigh, highLow, highHigh)
+	return slice
 }
 
 // Keccak256 computes the Keccak-256 hash of the input data.
--- a/pkg/vm/builtins/keccak.go
+++ b/pkg/vm/builtins/keccak.go
@@ -1,0 +1,26 @@
+package builtins
+
+import (
+	"fmt"
+
+	"golang.org/x/crypto/sha3"
+)
+
+// Keccak256 computes the Keccak-256 hash of the input data.
+func Keccak256(data []byte) ([]byte, error) {
+	hasher := sha3.NewLegacyKeccak256()
+	if _, err := hasher.Write(data); err != nil {
+		return nil, err // Return the error to the caller.
+	}
+	return hasher.Sum(nil), nil // Return the hash and a nil error.
+}
+
+// Cairo_Keccak computes and prints the Keccak-256 hash of the input data.
+func Cairo_Keccak(data []byte) {
+	hash, err := Keccak256(data)
+	if err != nil {
+		fmt.Println("Error:", err)
+		return
+	}
+	fmt.Printf("%x\n", hash)
+}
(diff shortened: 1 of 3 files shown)


jkktom:
Dear @rodrigo-pino , This PR is ready for review.
Keccak256 works for suitable inputs with appropriate paddings applied. All tested with success.
All the test inputs in test codes verified from Web result : https://emn178.github.io/online-tools/keccak_256.html 
Same inputs, Same results.

Still contains Println and Printf lines for verification and debug. This will be commented out after reveiw.
Thanks.

jkktom:
Resolves issue #130

jmjac:
Spend some time today trying to figure out how this should work and it seems the keccak builtin does the keccak-f1600 permutation to the 8 felt 200bit inputs and writes 8 felt 200bit outputs.

pkg/vm/builtins/keccak.go
```diff
@@ -0,0 +1,257 @@
...
+		if !value.Known() {
+			return fmt.Errorf("cannot infer value: input value at offset %d is unknown", startOffset+i)
+		}
+		v, err := value.FieldElement()
+		if err != nil {
+			return fmt.Errorf("Keccak input has to be felt")
+		}
+		var out [32]byte
+		fp.LittleEndian.PutElement(&out, *v)
+		copy(data[i*25:i*25+25], out[:25]) //25*8 = 200bits
```
rodrigo-pino, line 46: 25 * 8 = 200 bytes ?

pkg/vm/builtins/keccak.go
```diff
@@ -0,0 +1,257 @@
...
+	var output [200]byte
+	for i := 0; i < 25; i++ {
+		v := binary.LittleEndian.AppendUint64(nil, out[i])
+		copy(output[i*8:i*8+8], v)
+	}
+
+	for i := 0; i < inputCellsPerKeccak; i++ {
+		var bytes [32]byte
+		copy(bytes[:], output[i*25:i*25+25])
+		//Only 200 bits so always fits, no need to check error
```
rodrigo-pino, line 65: Only 200 bits, or 200 bytes? I am unsure if it should be bits, or bytes. Since data structures are in bytes but comments are in bits. Shouldn't all be bits?

pkg/vm/builtins/keccak_math.go
```diff
@@ -0,0 +1,411 @@
+package builtins
+
+import "math/bits"
+
+// Slighly modified from: https://github.com/golang/crypto/blob/master/sha3/keccakf.go
```
rodrigo-pino, line 5: What modifications we required?

pkg/vm/builtins/keccak.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+	}
+	return nil
+}
+
+func (k *Keccak) String() string {
+	return KeccakName
+}
+
+const (
```
rodrigo-pino, line 80: Are we using the code below this line?

rodrigo-pino (commented):
@jkktom @jmjac good work! Just a couple of questions, and remember we need to add an integration test too. Just like we have one for the other builtins.

pkg/vm/builtins/keccak.go
jmjac: It's 200 bits. We read 25 bytes from each felt which gives us 200bits per felt. 1600bits in total from the 8 inputs

pkg/vm/builtins/keccak.go
jmjac: Maybe rewording it to `This is 25*8 bits which is smaller than max felt 252 bits so no need to check the error` would make it clearer?

rodrigo-pino: Yes, we are talking about bits, but the data slice for example holds 200 bytes, shouldn't it hold 25 bytes instead?

pkg/vm/builtins/keccak_math.go
jmjac: As I remember I had to change it to work on values instead of pointers during testing. Looking at it now it works the just fine with the original code so I'll change it back to that for easier compatibility if go's keccakf1600 ever gets updated.

pkg/vm/builtins/keccak.go
jmjac: Not as I'm aware though @jkktom may know if it maybe used in the future for cairo 1

pkg/vm/builtins/keccak.go
jmjac: There are two arrays here. The first output holds 200 bytes because keccakf1600 works on [25]uint64 which is also 200 bytes. Then it does the reverse of the first loop and converts each 25bytes to a felt which gives us in total 200bytes converted to 8 felts where each felt is 25bytes=200bits.

pkg/vm/builtins/keccak.go
jkktom: DIscussed with Jacob, to be deleted as it's not necessary.

rodrigo-pino: Let's put this code in `pkg/utils/keccak.go`, to have it separated from the builtin

pkg/vm/builtins/keccak_test.go
rodrigo-pino, line 1: Let's add a test as well

jmjac: Moved and commented out debug statements to not clutter the test output

rodrigo-pino (approved):
Amazing! Thanks @jkktom and @jmjac

Issue #138, opened 2023-10-24
Add Support for Dictionaries


rodrigo-pino:
In order to add dictionaries we need to:

* Implement an execution context for Dictionary Hints
* Implement dictionary hints

rodrigo-pino:
Solved in #146

PR #146, opened 2023-10-28
Feat: Add Dictionaries

rodrigo-pino:
Solves #138 

Add all dictionary related hints. I had to implement them all together to have a good grasp of all the context scopes. I didn't add tests on purpose to avoid re-writing any test in the case I'd misunderstood something down the line.

This means that dictionaries are implemented but tests are lacking, they'll come as follow up, together with benchmark tests.

This includes the following hints for dictionaries:

* AllocFelt252Dict
* Felt5252DictEntryInit
* Felt252DictEntryUpdate
* GetSegmentArenaIndex

And the following hints for squashed dictionaries:

* InitSquashData
* GetCurrentAccessIndex
* ShouldSkipSquashLoop
* GetCurrentAccessDelta
* ShouldContinueSquashLoop
* GetNextDictKey

Both dictionaries and squashed dictionaries come with their own set of execution contexts:

- Dictionaries use the context to keep track of what dictionary is in what segment and the values it holds given a key
- Squashed Dictionaries keeps track of the keys that have been already been checked out

### DIFF
--- a/pkg/hintrunner/constants.go
+++ b/pkg/hintrunner/constants.go
@@ -1,7 +1,0 @@
-package hintrunner
-
-import "github.com/holiman/uint256"
-
-func MaxU128() uint256.Int {
-	return uint256.Int{18446744073709551615, 18446744073709551615, 0, 0}
-}
--- a/pkg/hintrunner/hint.go
+++ b/pkg/hintrunner/hint.go
@@ -474,3 +474,25 @@ func (hint *InitSquashData) Execute(vm *VM.VirtualMachine, ctx *HintRunnerContex
 
 	return vm.Memory.WriteToAddress(&firstKeyAddr, &mv)
 }
+
+type GetCurrentAccessIndex struct {
+	RangeCheckPtr ResOperander
+}
+
+func (hint *GetCurrentAccessIndex) String() string {
+	return "GetCurrentAccessIndex"
+}
+
+func (hint *GetCurrentAccessIndex) Execute(vm *VM.VirtualMachine, ctx *HintRunnerContext) error {
+	rangeCheckPtr, err := ResolveAsAddress(vm, hint.RangeCheckPtr)
+	if err != nil {
+		return fmt.Errorf("resolve range check pointer: %w", err)
+	}
+
+	lastIndex := f.NewElement(
+		ctx.SquashedDictionaryManager.LastIndex(),
+	)
+	mv := mem.MemoryValueFromFieldElement(&lastIndex)
+
+	return vm.Memory.WriteToAddress(&rangeCheckPtr, &mv)
+}
--- a/pkg/hintrunner/hint.go
+++ b/pkg/hintrunner/hint.go
@@ -516,7 +516,7 @@ type GetSegmentArenaIndex struct {
 }
 
 func (hint *GetSegmentArenaIndex) String() string {
-	return "FeltGetSegmentArenaIndex"
+	return "GetSegmentArenaIndex"
 }
 
 func (hint *GetSegmentArenaIndex) Execute(vm *VM.VirtualMachine, ctx *HintRunnerContext) error {
--- a/pkg/hintrunner/hint.go
+++ b/pkg/hintrunner/hint.go
@@ -21,7 +21,7 @@ type AllocSegment struct {
 	dst CellRefer
 }
 
-func (hint AllocSegment) String() string {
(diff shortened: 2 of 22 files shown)


pkg/hintrunner/hint_test.go
```diff
@@ -184,8 +185,14 @@ func TestWideMul128(t *testing.T) {
 	var dstLow ApCellRef = 1
 	var dstHigh ApCellRef = 2
 
-	lhs := Immediate(*big.NewInt(1).Lsh(big.NewInt(1), 127))
-	rhs := Immediate(*big.NewInt(1<<8 + 1))
+	lhsBytes := uint256.NewInt(1).Lsh(uint256.NewInt(1), 127).Bytes32()
```
jmjac, line 188: I think it's clearer as `new(uint256.Int).Lsh(uint256.NewInt(1), 127).Bytes32()` given that `newInt(1)` is instantly overwritten

pkg/hintrunner/hint_test.go
```diff
@@ -220,8 +227,12 @@ func TestWideMul128IncorrectRange(t *testing.T) {
 	var dstLow ApCellRef = 1
 	var dstHigh ApCellRef = 2
 
-	lhs := Immediate(*big.NewInt(1).Lsh(big.NewInt(1), 128))
-	rhs := Immediate(*big.NewInt(1))
+	lhsBytes := uint256.NewInt(1).Lsh(uint256.NewInt(1), 128).Bytes32()
```
jmjac, line 230: `new(uint256.Int).Lsh(uint256.NewInt(1), 128).Bytes32()`

pkg/hintrunner/hint.go
```diff
@@ -287,13 +289,362 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+	return "GetCurrentAccessDelta"
+}
+
+func (hint *GetCurrentAccessDelta) Execute(vm *VM.VirtualMachine, ctx *HintRunnerContext) error {
+	indexDeltaPtr, err := hint.IndexDeltaMinusOne.Get(vm)
+	if err != nil {
+		return fmt.Errorf("get index delta address: %w", err)
+	}
+
+	previousKeyIndex := ctx.SquashedDictionaryManager.PopIndex()
```
jmjac, line 599: This will panic if the keys list is empty. Is that situation possible?

pkg/hintrunner/hintrunner.go
```diff
@@ -4,25 +4,215 @@ import (
...
+		idx:  uint64(len(dm.dictionaries)),
+	}
+	return newDictAddr
+}
+
+// Given a memory address, it looks for the right dictionary using the segment index. If no
+// segment is associated with the given segment index, it errors
+func (dm *DictionaryManager) GetDictionary(dictAddr *mem.MemoryAddress) (Dictionary, error) {
+	dict, ok := dm.dictionaries[dictAddr.SegmentIndex]
+	if !ok {
```
jmjac, line 65: nit: All other functions in this file follow pattern 
```
if dict, ok := ...; ok{
     return value;
}
return error;
```
this one is the only that does the inverse

pkg/hintrunner/hint.go
```diff
@@ -371,13 +373,383 @@ func (hint SquareRoot) Execute(vm *VM.VirtualMachine) error {
...
+
+	value, err := hint.Value.Resolve(vm)
+	if err != nil {
+		return fmt.Errorf("resolve value: %w", err)
+	}
+
+	return ctx.DictionaryManager.Set(&dictPtr, key, &value)
+}
+
+type GetSegmentArenaIndex struct {
```
jmjac, line 513: Why is this struct the only one that's `String()` value is different than struct name?

jmjac (approved):

pkg/hintrunner/hint.go
rodrigo-pino: oops!

pkg/hintrunner/constants.go
jkktom, line 1: @rodrigo-pino Can you please let me know why was this file deleted?

rodrigo-pino: It was moved

rodrigo-pino: It's new location makes all constant accessible to any project pkg. We had some repetition around

Issue #135, opened 2023-10-23
Assembler cannot handle negative immediate values


rodrigo-pino:
Assember cannot parse expressions which have a negative immediate value. For example:
```go
[ap] = -1;
// or
jmp rel -1;
```

PR #148, opened 2023-11-01
Fix bug for parsing negative immediate values

mmk-1:
Fixes issue #135


pkg/assembler/grammar.go
```diff
@@ -126,7 +131,15 @@ func (e *Expression) AsMathOperation() *MathOperation {
...
-	return e.Immediate
+	if e.Immediate == nil {
+		return nil
+	}
+	if e.Immediate.Sign != "" {
+		immediate := signedString(e.Immediate.Sign == "-", *e.Immediate.Value)
+		return &immediate
+	}
+	immediate := signedString(false, *e.Immediate.Value)
+	return &immediate
```
rodrigo-pino, line 142: This all can be refactored into:
```go
imm := signedString(e.Immediate.Sign == "-", e.Immediate.Value)
return imm;
```

After this refactor there is no need for a `signedString` function as well. You can just put in `AsImmediate` instead:

```go
if e.Immediate.Sign == "-" {
   return the "-%s"
}
return the immediate
```

pkg/assembler/grammar_test.go
rodrigo-pino, line 1: There should be a test for a negative immediate as well, on both grammar_test.go and assembler_test.go

rodrigo-pino, line 1: We should add a test as well on assembler_test.go, just to make sure the negative immediate is behaving as expected

rodrigo-pino (approved):
Good job! Just an extra test and it is good to go

rodrigo-pino:
Also, please be sure to merge from `main` before merging to `main`

PR #149, opened 2023-11-01
Refactor: Safemath is now the Util pkg

rodrigo-pino:
Change name to a more fitting one. There we have tools for math, arrays and everything else required along the lines.


jmjac (approved):
lgtm

PR #150, opened 2023-11-02
Fix bench Mul128 error

jmjac:
Benchmark for Mul128 was failing when `randomFeltElement` was bigger than `u128`. This should fix it. Also there is `(&f.Element{}).SetRandom()` to create felts between `[0,q)`. I think we could use it instead of current `randomFeltElement` implementation unless I'm missing something.


rodrigo-pino:
> Benchmark for Mul128 was failing when randomFeltElement was bigger than u128. This should fix it. Also there is (&f.Element{}).SetRandom() to create felts between [0,q). I think we could use it instead of current randomFeltElement implementation unless I'm missing something.

@jmjac  the reason behind not using it is that I was looking for certain determinism in benchmark test. There are lot of variables during benchmarks already, just to make sure we are testing the same thing. I've also noticed that `randomFeltElement` is not entirely correct since it could output invalid felt numbers, something to fix in the near future

jmjac:
Makes sense. For the `randomFeltElement` we could use the similar approach as for u128 and limit the last z[3] to 50bits like 
``` 
v = binary.BigEndian.AppendUint64(nil, rand.Uint64()>>14)
copy(b[16:24], v)
``` 
so that at max we would be using 250bits which is lower than 251 bits of felt.

rodrigo-pino:
> so that at max we would be using 250bits which is lower than 251 bits of felt

I agree, would you like to add this fix as part of this PR?

jmjac:
@rodrigo-pino added the fix and simplified the function

rodrigo-pino (approved):

PR #151, opened 2023-11-03
Implement ECDSA builtin

### LINKED ISSUE #145: ECDSA

jmjac:
Resolves #145. It seems that ecdsa builtin takes the X part of the public key and hash of the msg. Builtin should recover Y, check if it's on the curve and then verify the signature. The part that is missing is when `addSignature` should be called to put the sig in the memory


jmjac:
Because VM doesn't  yet support running hints the integration test will fail

pkg/vm/builtins/ecdsa.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// recoverY recovers the y and -y coordinate of x. True y can be either y or -y
+func recoverY(x *fp.Element) (*fp.Element, *fp.Element, error) {
+	ALPHA := fp.NewElement(1)
+	BETA := fp.Element{}
+	_, _ = BETA.SetString("3141592653589793238462643383279502884197169399375105820974944592307816406665")
+	// y_squared = (x * x * x + ALPHA * x + BETA) % FIELD_PRIME
+	x2 := new(fp.Element).Mul(x, x)
+	x3 := x2.Mul(x2, x)
```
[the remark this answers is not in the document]
rodrigo-pino: Also please don't use `new` since they perform heap allocations which we don't want

pkg/vm/builtins/ecdsa.go
```diff
@@ -0,0 +1,160 @@
...
+	a := new(fp.Element).Mul(&ALPHA, x)
```
rodrigo-pino, line 151: Please don't use new. As a rule of a thumb we always create `Element`s as value types

pkg/vm/builtins/ecdsa.go
```diff
@@ -0,0 +1,160 @@
...
+	x3.Add(x3, a)
+	x3.Add(x3, &BETA)
+	y := x3.Sqrt(x3)
```
rodrigo-pino, line 154: no need to store a new `y` as well, you could reuse `x3`. I understand that this could feel restrictive and not good practice, but this is code that should be as fast/cheap as possible. You could aid yourself from comments to mitigate this

pkg/vm/builtins/ecdsa.go
```diff
@@ -0,0 +1,160 @@
...
+	e.signatures[pubOffset] = sig
+	return nil
+}
+
+func (e *ECDSA) String() string {
+	return ECDSAName
+}
+
+// recoverY recovers the y and -y coordinate of x. True y can be either y or -y
+func recoverY(x *fp.Element) (*fp.Element, *fp.Element, error) {
```
rodrigo-pino, line 144: If we return pointers to elements, that means we are forcing heap allocations because those variables outlive their scope (the current function). We want to avoid that as much as possible and return value types. (there are still allocations, but this time on the stack, which is way cheaper)

pkg/vm/builtins/ecdsa.go
```diff
@@ -0,0 +1,160 @@
...
+	                    "starkware.cairo.common.signature.verify_ecdsa_signature.public_key": 1,
+	                    "starkware.cairo.common.signature.verify_ecdsa_signature.signature_r": 2,
+	                    "starkware.cairo.common.signature.verify_ecdsa_signature.signature_s": 3
+	                }
+	            }
+	        }
+	    ]
+	},
+*/
+func (e *ECDSA) AddSignature(pubOffset uint64, r, s fp.Element) error {
```
rodrigo-pino, line 119: `r` and `s` should be pointers(`*fp.Element`). The general style is that we receive references and return value types. This way we avoid copying the full Felt, and just a pointer to it. This pointer references the stack and not the heap if this technique is done well (go compiler perform escape analysis)

pkg/vm/builtins/ecdsa.go
```diff
@@ -0,0 +1,160 @@
...
+	starkcurve "github.com/consensys/gnark-crypto/ecc/stark-curve"
+	ecdsa "github.com/consensys/gnark-crypto/ecc/stark-curve/ecdsa"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+)
+
+const ECDSAName = "ecdsa"
+const cellsPerECDSA = 2
+
+type ECDSA struct {
+	signatures map[uint64]ecdsa.Signature
```
rodrigo-pino, line 16: What hint would add signatures here?

rodrigo-pino (commented):
Looks good. Minor optimization and style comments

pkg/vm/builtins/ecdsa.go
jmjac: Here we store y because `sqrt` returns nil if the square root doesn't exist. Otherwise we would need check if x2 changed which would require another variable either way

pkg/vm/builtins/ecdsa.go
jmjac: From compiler output it's this hint  `ecdsa_builtin.add_signature(ids.ecdsa_ptr.address_, (ids.signature_r, ids.signature_s))`

pkg/vm/builtins/ecdsa.go
jmjac: Fixed

pkg/vm/builtins/ecdsa.go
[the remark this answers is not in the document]
jmjac: Sure, removed additional allocations. This was written this way originally to be easier to read the math

rodrigo-pino (approved):
Great!!!

pkg/vm/builtins/ecdsa.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		if !valid {
+			return fmt.Errorf("signature is not valid")
+		}
+	}
+	return nil
+}
+
+func (e *ECDSA) InferValue(segment *memory.Segment, offset uint64) error {
+	return fmt.Errorf("Can't infer value")
```
rodrigo-pino, line 84: lower case here

PR #172, opened 2023-11-22
Python hints integration

### LINKED ISSUE #87: Integration of Python hints

cicr99:
Closes #87

This PR creates the structure that allows to identify the pythonic hints and initialize the hinter map used by the runner

Summary of the pipeline:
- The hints need to be extracted from the CairoZeroProgram, so that's done before initializing the runner with the method `GetZeroHints`
- The hints structure that is extracted from the zero program is `map[string][]Hint` where the string represents the pc, we assume here that there is only one hint in the array
- The type of hint can be identified using the code of the hint which is always the same
- The parameters for the hint need to be extracted as well, for that we use the references in the hint structure and the identifiers of the program. The identifier might have several references, so the one we need is the closest one whith pc less than the pc of the hint.
- Once the reference was found, the value needs to be parsed in order to get the actual parameter. The grammar defiend in `hintparser` file is the one used for the parsing
- Having identified the parameters and parse the parameters, its just a matter of building the corresponding Hinter


pkg/runners/zero/zero_benchmark_test.go
```diff
@@ -214,12 +217,22 @@ func BenchmarkRunnerWithFibonacci(b *testing.B) {
     `)
 
 	for i := 0; i < b.N; i++ {
-		program, err := LoadCairoZeroProgram(compiledJson)
+        cairoZeroJson, err := zero.ZeroProgramFromJSON(compiledJson)
```
Tomi-3-0, line 220: I think it might be a good idea to reset the timer for this benchmark test since we already agreed with that convention.

Tomi-3-0, line 220: I think it might be a good idea to reset the timer for this benchmark test since we already agreed with that convention.

Tomi-3-0 (changes requested):

Tomi-3-0 (approved):

pkg/hintrunner/hintparser.go
rodrigo-pino, line 1: Loved the code. What do you think of having all pythonic hints logic inside their own package (a sub pkg of hintrunner), to have that logic separated.

pkg/hintrunner/zerohint.go
rodrigo-pino, line 1: I would add this file as well inside the sub pkg, would making a pythonic hintrunner make things easier?

cicr99: I like the idea of adding these files in a sub pkg of the hintrunner, but I don't think it would be necessary to make a specific hintrunner for this. The only thing changing is the way the hinter map is initialized, but the logic after that is the same for both

cicr99: After some consideration, turns out having different folders for this also means moving other code around, else you get cyclic dependencies. The pieces of code that need to be moved, although they are not very big, they are quite a few and not  all of them are related to this PR content. I think reorganizing these files should go on a separate PR so as to make the review process easier.

pkg/runners/zero/program.go
```diff
@@ -20,12 +20,7 @@ type Program struct {
 	Builtins []sn.Builtin
 }
 
-func LoadCairoZeroProgram(content []byte) (*Program, error) {
-	cairoZeroJson, err := zero.ZeroProgramFromJSON(content)
-	if err != nil {
-		return nil, err
-	}
-
+func LoadCairoZeroProgram(cairoZeroJson *zero.ZeroProgram) (*Program, error) {
```
quasilyte, line 23: Ah, I like this change.
I would want to use this new API in my PR as well because I need to have a "raw" json object in addition to the runnerzero-processed `program`.

pkg/hintrunner/hintparser.go
```diff
@@ -0,0 +1,297 @@
...
+	case expression.DerefCastExp != nil:
+		return expression.DerefCastExp.Evaluate()
+	case expression.CastExp != nil:
+		return expression.CastExp.Evaluate()
+	default:
+		return nil, fmt.Errorf("unexpected identifier value")
+	}
+}
+
+func (expression DerefCastExp) Evaluate() (interface{}, error) {
```
quasilyte, line 101: I'm almost sure that we use both `any` and `interface{}` in our code base.
What do you think about sticking to `any`?
It's a minor thing, but `interface{}` looks less pretty even if it was the only option before `any` was introduced. :D

pkg/runners/zero/zero_benchmark_test.go
```diff
@@ -3,6 +3,9 @@ package zero
 import (
 	"math"
 	"testing"
+
+    hintrunner "github.com/NethermindEth/cairo-vm-go/pkg/hintrunner"
```
quasilyte, line 7: Does your editor/IDE auto-apply the gofmt on save?
I noticed a couple of lines that look like they're indented with spaces.

pkg/hintrunner/zerohint_test.go
```diff
@@ -0,0 +1,52 @@
+package hintrunner
+
+import (
+	// "fmt"
```
quasilyte, line 4: This is probably a debug leftover.
With things like goimports (and alike), I think we can safely remove a line like that.

pkg/hintrunner/zerohint.go
```diff
@@ -0,0 +1,119 @@
...
+func GetZeroHints(cairoZeroJson *zero.ZeroProgram) (map[uint64]Hinter, error) {
+	hints := make(map[uint64]Hinter)
+	for counter, rawHints := range cairoZeroJson.Hints {
+		pc, err := strconv.ParseUint(counter, 10, 64)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(rawHints) != 1 {
+			return nil, fmt.Errorf("expected only 1 hint but got  %d", len(rawHints))
```
quasilyte, line 23: Since you probably have the best understanding regarding the multi-hints per single PC, could you document it here in form of a comment?
e.g. when they occur and what would they look like in (cairo0?) code.

pkg/hintrunner/zerohint_test.go
cicr99: Oh, it was. I thought I had removed them all

pkg/hintrunner/hintparser.go
cicr99: We mostly use `interface{}` in the code base. But I agree with using `any` instead of `interface{}`. I've never liked `interface{}` in the first place. However, if we are sticking to `any` we should probably change it everywhere so as to keep the same format throughout the project.

pkg/runners/zero/zero_benchmark_test.go
cicr99: I'll check that

pkg/hintrunner/zerohint.go
cicr99: I had a discussion with @rodrigo-pino about this when I first started coding this PR. I'll have to check the code of the compiler again to write the comment. I was thinking of adding an explanation of this file in a documentation page anyway, and it won't hurt to have a comment here as well

quasilyte (approved):

pkg/hintrunner/zerohint.go
cicr99: I've been going through the cairo 0 compiler but there are lots of details I'm not certain about, this one is one of them. All of the examples I've worked with have only one hint per pc, but I haven't found the place in the compiler's code that assures that's always the case when using only whitelisted hints. I'm going to leave a TODO comment to continue studying this, and in the case it was an incorrect assumption, then I'll create another PR to fix it. The structure of the code shouldn't change much.

cicr99: I've been going through the cairo 0 compiler and there are lots of details I'm not certain about. Most of the examples I've worked with only have one hint per pc, but I haven't been able to find in the compiler code how to be certain it can't be more than that. I'll leave a TODO comment to continue studying this and in the case it was an incorrect assumption I'll create another PR to fix it. I think for the first hints to be implemented we can work as it is.

PR #177, opened 2023-11-29
RandomEcPoint Hint Implementation

mmk-1:
Random Felt generation is used in this hint. I thought to move the random generator functions, into `hintrunner/utils.go` so it can be used outside of tests. But if you have a better approach let me know.

### DIFF
--- a/pkg/hintrunner/hint.go
+++ b/pkg/hintrunner/hint.go
@@ -1289,7 +1289,10 @@ func (hint *RandomEcPoint) String() string {
 
 func (hint *RandomEcPoint) Execute(vm *VM.VirtualMachine) error {
 	// Keep sampling a random field element `X` until `X^3 + X + beta` is a quadratic residue.
+
+	// Starkware's elliptic curve Beta value https://docs.starkware.co/starkex/crypto/stark-curve.html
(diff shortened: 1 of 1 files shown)


pkg/hintrunner/hint.go
```diff
@@ -1278,3 +1277,58 @@ func (hint *AssertLeIsSecondArcExcluded) Execute(vm *VM.VirtualMachine, ctx *Hin
...
+	y CellRefer
+}
+
+func (hint *RandomEcPoint) String() string {
+	return "RandomEc"
+}
+
+func (hint *RandomEcPoint) Execute(vm *VM.VirtualMachine) error {
+	// Keep sampling a random field element `X` until `X^3 + X + beta` is a quadratic residue.
+	betaFelt := f.Element{3863487492851900874, 7432612994240712710, 12360725113329547591, 88155977965380735}
```
cicr99, line 1292: Where do these values come from?
If they are from an external source it would be nice to add it here as a comment

cicr99 (commented):
I do agree on moving the functions to the utils file btw

cicr99 (approved):

Issue #178, opened 2023-11-30
Implement FieldSqrt hint


cicr99:
- Params: `val: ResOperand, sqrt: CellRef`
- Description: Computes the square root of `val`, if `val` is a quadratic residue, and of `3 * val` otherwise. Since 3 is not a quadratic residue, exactly one of `val` and `3 * val` is a quadratic residue (unless `val` is 0). This allows proving that `val` is not a quadratic residue.

The implementation of the hint should go in `pkg/hintrunner/hint.go` following a similar format to the ones already implemented there. Once this is done, tests and benchmarks should also be implemented for it in `pkg/hintrunner/hint_test.go` and `pkg/hintrunner/hint_bechmark_test.go` respectively.

PR #179, opened 2023-12-05
Implement FieldSqrt hint

mmk-1:
Resolves issue #178


cicr99 (commented):
Other than the comment about the test, looks good to me

mmk-1:
@cicr99 I added a new test cases for when the input is quadratic non-residue.
I also added a comments for the `Legendre` method with a link that gives the mathematical reasoning how it works. What do you think?

PR #180, opened 2023-12-06
Feat/uint256 div mod

Tomi-3-0:
Uint256DivMod Implementation:
Divides dividend (represented by 2 128bit limbs) by divisor (represented by 2 128bit limbs). Returns the quotient (represented by 2 128bit limbs) and the remainder (represented by 2 128bit limbs)
Tied to #46 
Values  for test gotten from [here](https://github.com/lambdaclass/cairo-vm_in_go/blob/31c3628bc10ebc1628685b3cdfa72d0e938b533e/pkg/hints/uint256_hints_test.go#L401)


pkg/hintrunner/hint_test.go
```diff
@@ -268,6 +268,110 @@ func TestWideMul128(t *testing.T) {
...
+	var quotient0 ApCellRef = 1
+	var quotient1 ApCellRef = 2
+	var remainder0 ApCellRef = 3
+	var remainder1 ApCellRef = 4
+
+	dividend0Felt := f.NewElement(89)
+	dividend1Felt := f.NewElement(72)
+
+	divisor0Felt := f.NewElement(3)
+	divisor1Felt := f.NewElement(7)
```
cicr99, line 285: It would also be nice to test an edge case as well, for example the division between 256 bit numbers, I mean, so they cannot fit in a felt once you put the two parts together

pkg/hintrunner/hint_bechmark_test.go
```diff
@@ -130,6 +130,44 @@ func BenchmarkWideMul128(b *testing.B) {
...
+	vm.Context.Ap = 0
+	vm.Context.Fp = 0
+
+	rand := defaultRandGenerator()
+
+	var quotient0 ApCellRef = 1
+	var quotient1 ApCellRef = 2
+	var remainder0 ApCellRef = 3
+	var remainder1 ApCellRef = 4
+
```
cicr99, line 144: Check here if you need to reset the timer before the for-loop

cicr99 (approved):

PR #185, opened 2024-01-22
pkg/assembler: allow JNZ with ap_change

### LINKED ISSUE #184: Error while decoding "0xa0680017fff8000" from Cairo1-produced CASM
While decoding `0xa0680017fff8000` with `assembler.DecodeInstruction`:
```
flags: jnz opcode must have unconstrained res logic, no opcode, and no ap change
```
It could be the case that CASM changed a bit since Cairo0 times or our assembler needs to relax some of its restrictions.
I'll take a look at the Cairo1 compiler code to know what this new combination of flags should signify.

This issue will probably resolved in the same context as #183, but I may send a separate PR just for this fix (if any).

`thoth` disassembles it into an ordinary JNZ:
```
offset 0:          JNZ                 7                   # JMP 7
```
The input bytecode looks like this (we only care about the first value in this array, so the output is truncated):
```
{
  "prime": "0x800000000000011000000000000000000000000000000000000000000000001",
  "compiler_version": "2.4.1",
  "bytecode": [
    "0xa0680017fff8000",
    "0x7",
```
Instruction fields:
```
uint64 value: 722405534170316800
off_dst: -32768 (1000000000000000)
off_op0: 32767 (111111111111111)
off_op1: -32767 (1000000000000001)
dst_reg: 0 (0)
op0_reg: 1 (1)
op1_src: 1 (1)
res_logic: 0 (0)
pc_update: 4 (100)
ap_update: 2 (10)
opcode: 0 (0)
bits: 1000000000000000 111111111111111 1000000000000001 0 1 1 0 100 10 0
```

quasilyte:
The cairo1 compiler generates jnz with ap++ for some range-check related code. To support this code in our assembler, the ap_change check should be removed.

See the referenced issue for more context.

Fixes #184

### DIFF
--- a/pkg/assembler/instruction.go
+++ b/pkg/assembler/instruction.go
@@ -332,13 +332,15 @@ func decodeInstructionFlags(instruction *Instruction, flags uint16) error {
 	}
 	instruction.Opcode = Opcode(opcode)
 
-	// for pc udpate Jnz, res should be unconstrainded, no opcode, and ap should update with Imm
+	// for pc udpate Jnz, res should be unconstrainded, no opcode;
+	// it used to have an ap update check, but new Cairo1 compiler
+	// emits Jnz with ap_add1 sometimes
+	// See #184
 	if instruction.PcUpdate == PcUpdateJnz &&
 		(instruction.Res != Unconstrained ||
-			instruction.Opcode != OpCodeNop ||
-			instruction.ApUpdate != SameAp) {
+			instruction.Opcode != OpCodeNop) {
 		return fmt.Errorf(
-			"jnz opcode must have unconstrained res logic, no opcode, and no ap change",
+			"jnz opcode must have unconstrained res logic and no opcode",
 		)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/assembler/instruction.go
```diff
@@ -332,11 +332,13 @@ func decodeInstructionFlags(instruction *Instruction, flags uint16) error {
...
+	// it used to have an ap update check, but new Cairo1 compiler
+	// emits Jnz with ap_add1 sometimes
+	// See #184
 	if instruction.PcUpdate == PcUpdateJnz &&
 		(instruction.Res != Unconstrained ||
-			instruction.Opcode != OpCodeNop ||
-			instruction.ApUpdate != SameAp) {
+			instruction.Opcode != OpCodeNop) {
 		return fmt.Errorf(
 			"jnz opcode must have unconstrained res logic, no opcode, and no ap change",
```
cicr99, line 343: Probably the part of "... and no ap change" should be removed from the error message now

pkg/assembler/instruction_test.go
```diff
@@ -214,7 +214,6 @@ func TestInvalidOpcode(t *testing.T) {
 func TestPcUpdateJnzInvalid(t *testing.T) {
 	instructionInvalidRes := new(f.Element).SetBytes([]byte{0x06, 0x27, 0x80, 0x01, 0x80, 0x01, 0x80, 0x00})
 	instructionInvalidOpcode := new(f.Element).SetBytes([]byte{0x16, 0x07, 0x80, 0x01, 0x80, 0x01, 0x80, 0x00})
-	instructionInvalidApUpdate := new(f.Element).SetBytes([]byte{0x06, 0x07, 0x80, 0x01, 0x80, 0x01, 0x80, 0x00})
 
 	expectedError := "jnz opcode must have unconstrained res logic, no opcode, and no ap change"
```
cicr99, line 218: Same over here if we change the other one

pkg/assembler/instruction.go
quasilyte: Good catch! Fixed

cicr99 (approved):

PR #188, opened 2024-01-24
assembler: parse negative values in Offset properly

### LINKED ISSUE #186: assembler can't parse negative immediate values
Given this line:
```
jmp rel 6 if [fp + -3] != 0;
```
We get this error:
```
1:20: unexpected token "-" (expected <int>)
```
It looks like it can't handle negative immediate values.

quasilyte:
Fixes #186

The idea is taken from here:
https://github.com/alecthomas/participle/issues/112#issuecomment-703019668

### DIFF
--- a/pkg/assembler/grammar.go
+++ b/pkg/assembler/grammar.go
@@ -68,7 +68,7 @@ type DoubleDeref struct {
 
 type Offset struct {
 	Sign  string `@("+" | "-")`
-	Value *int   `@Int`
+	Value *int   `@("-"? Int)`
 }
 
 type MathOperation struct {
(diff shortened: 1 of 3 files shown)


pkg/assembler/grammar.go
```diff
@@ -68,7 +68,7 @@ type DoubleDeref struct {
 
 type Offset struct {
 	Sign  string `@("+" | "-")`
-	Value *int   `@Int`
+	Value *int   `@('-'? Int)`
```
cicr99, line 71: Nice fix! I wasn't aware you could do that.
I think you can use double quotes in `"-"` to use the same format as the  rest of the code, but other than that, it looks good

quasilyte: Good call.
Fixed.

cicr99 (approved):

PR #192, opened 2024-01-29
Hintrunner package structure

cicr99:
The idea is to separate the logic of cairo 0 hints from the cairo 1 hints, so a subpackage was created for each. The hintrunner should remain the same in spite of the version that's being used.

Refer to: https://github.com/NethermindEth/cairo-vm-go/pull/172#discussion_r1428114543


quasilyte (approved):
Can we merge the gofmt linter PR and then merge this one?

PR #195, opened 2024-01-30
pkg/parsers/zero: improve Identifiers typing

quasilyte:
This allows us to do less redundant type assertions while handling the zero prog identifiers.

We may go even further, but this step is enough for now to make zero hints handler less awkward.

The `.([]zero.Reference)` assertion was not working as the slice contained `[]any` instead.
Instead of rewriting the code to handle even more
weakly typed `any` values, this patch improves the Identifier objects typing.
This makes type assertions and slice conversions unnecessary.


cicr99 (approved):
LGTM

PR #196, opened 2024-02-02
pkg/hintrunner/zero: allow multiplication binary ops in references

quasilyte:
References may include the multiplication inside them.

Let's take this like of code for the example:

    https://github.com/starkware-libs/cairo-lang/blob/caba294d82eeeccc3d86a158adb8ba209bf2d8fc/src/starkware/cairo/common/math.cairo#L193

It will produce a reference like this:
```json
    {
        "cairo_type": "felt",
        "full_name": "starkware.cairo.common.math.assert_le_felt.arc_prod",
        "references": [
            {
                "ap_tracking_data": {
                    "group": 1,
                    "offset": 8
                },
                "pc": 14,
                "value": "cast([ap + (-5)] * [ap + (-1)], felt)"
            }
        ],
        "type": "reference"
    }
```

### DIFF
--- a/pkg/hintrunner/zero/hintparser.go
+++ b/pkg/hintrunner/zero/hintparser.go
@@ -3,6 +3,7 @@ package zero
 import (
 	"fmt"
 
+	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
 	op "github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
 	"github.com/alecthomas/participle/v2"
 )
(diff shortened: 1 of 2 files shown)


pkg/hintrunner/zero/hintparser.go
```diff
@@ -238,6 +244,11 @@ func (expression BinOpExp) Evaluate() (any, error) {
 		return nil, err
 	}
 
+	operation, err := parseOperator(expression.Operator)
+	if err != nil {
+		return nil, err
+	}
+
 	switch lResult := leftExp.(type) {
 	case op.CellRefer:
```
cicr99, line 253: I have a question regarding this case. I extracted the possible cases from https://github.com/lambdaclass/cairo-vm_in_go/blob/main/pkg/hints/hint_utils/hint_reference.go#L41, so this would be when you have: `reg - off1 + off2` for example. The problem here is that this could be a BinaryOperation, but the left side of the BinaryOperation operand should always be a Deref. So the solution we found for this was to add up the offsets and turn this into a CellRefer. My question would be, is it possible to get a multiplication here as well? And if so, how should we handle this?

cicr99: The solution for adding the multiplication implemented in this PR looks good.

I was just wondering if it was possible to also get a multiplication in this specific case of the `switch`  where the left side is a CellRefer, something like `(reg + off1) * off2` for example. If that was possible I'm not sure how it should be handled, in the sense that you should get some kind of operand from this expression and a BinaryOperation operand is not possible atm as you need a Deref as the left side, and the other possibility I see is try to create a CellRefer instead, but not sure how it would work with the multiplication.

quasilyte: I added a comment to reflect that. Let me know if it makes sense.

PR #200, opened 2024-02-12
Manage references and their ids

cicr99:


pkg/hintrunner/zero/hint_reference_resolver.go
```diff
@@ -1,56 +1,45 @@
...
-		operander: v,
-	})
-	m.numResOperander++
+func (m *hintReferenceResolver) AddReference(name string, v hinter.Reference) error {
+	shortName := shortSymbolName(name)
+	_, ok := m.refs[shortName]
+	if ok {
+		return fmt.Errorf("cannot overwrite reference %s (%s)", shortName, name)
+	}
+	fmt.Printf("hola1")
```
quasilyte, line 25: Probably a debug leftover.

quasilyte:
Otherwise, LGTM.

quasilyte (approved):

pkg/hintrunner/zero/hint_reference_resolver.go
quasilyte: I'm not sure being careful all the time is easier than removing those from time to time from the codebase. :D
I'm just trying to notice at least something useful while doing a review. :D

Issue #202, opened 2024-02-13
Investigate error in proof_mode when executing the specified code


quasilyte:
Just a reminder that this case could be worth investigating.

Our VM executed the given code without an error with and without a proof mode.
`cairo-run` doesn't want to execute it in proof mode.
It could be either an error on our side, or maybe an implementation issue on cairo0 side; or perhaps the code itself is invalid (it still needs an explanation).

Tested on:
* cairo-lang 0.11 (as recommended in our README)
* cairo-lang 0.12.2
* cairo-lang 0.13.0

```
quasilyte@lispbook2:~/CODE/cairo-vm-go2$ cairo-compile --version
cairo-compile 0.12.2

quasilyte@lispbook2:~/CODE/cairo-vm-go2$ cat test.cairo 
%builtins range_check

from starkware.cairo.common.math import assert_le_felt

func main{range_check_ptr}() {
    alloc_locals;
    local v1 = 543;
    local v2 = 657;

    assert_le_felt(v1, v2);

    ret;
}

quasilyte@lispbook2:~/CODE/cairo-vm-go2$ cairo-compile test.cairo --no_debug_info --proof_mode --output ./test_compiled.json

quasilyte@lispbook2:~/CODE/cairo-vm-go2$ cairo-run --version
cairo-run 0.12.2

quasilyte@lispbook2:~/CODE/cairo-vm-go2$ cairo-run --proof_mode --program test_compiled.json
Traceback (most recent call last):
  File "/home/quasilyte/.local/bin/cairo-run", line 10, in <module>
    sys.exit(main())
  File "/home/quasilyte/.local/lib/python3.10/site-packages/starkware/cairo/lang/vm/cairo_run.py", line 199, in main
    res = cairo_run(args)
  File "/home/quasilyte/.local/lib/python3.10/site-packages/starkware/cairo/lang/vm/cairo_run.py", line 326, in cairo_run
    runner.end_run(disable_trace_padding=disable_trace_padding)
  File "/home/quasilyte/.local/lib/python3.10/site-packages/starkware/cairo/lang/vm/cairo_runner.py", line 382, in end_run
    self.segments.compute_effective_sizes(allow_tmp_segments=allow_tmp_segments)
  File "/home/quasilyte/.local/lib/python3.10/site-packages/starkware/cairo/lang/vm/memory_segments.py", line 91, in compute_effective_sizes
    raise SecurityError(
[...]

PR #203, opened 2024-02-13
implement cairo0 hints needed for assert_le_felt

quasilyte:
In turn, `assert_le_felt` is needed for some other functions that use cairo0 hints, like `is_nn`.

### DIFF
--- a/pkg/hintrunner/zero/hint_reference_resolver.go
+++ b/pkg/hintrunner/zero/hint_reference_resolver.go
@@ -33,6 +33,20 @@ func (m *hintReferenceResolver) GetReference(name string) (hinter.Reference, err
 	return nil, fmt.Errorf("missing reference %s", name)
 }
 
+// GetResOperander returns the result of GetReference type-asserted to ResOperander.
+// If reference is not found or it's not of ResOperander type, a non-nil error is returned.
+func (m *hintReferenceResolver) GetResOperander(name string) (hinter.ResOperander, error) {
+	ref, err := m.GetReference(name)
+	if err != nil {
+		return nil, err
+	}
+	op, ok := ref.(hinter.ResOperander)
+	if !ok {
+		return nil, fmt.Errorf("expected %s to be ResOperander (got %T)", name, ref)
+	}
+	return op, nil
+}
+
 // shortSymbolName turns a full symbol name like "a.b.c" into just "c".
 func shortSymbolName(name string) string {
 	i := strings.LastIndexByte(name, '.')
--- a/pkg/hintrunner/zero/hintcode.go
+++ b/pkg/hintrunner/zero/hintcode.go
@@ -1,10 +1,16 @@
 package zero
 
 const (
-	AllocSegmentCode string = "memory[ap] = segments.add()"
+	allocSegmentCode string = "memory[ap] = segments.add()"
 
 	// This is a very simple Cairo0 hint that allows us to test
 	// the identifier resolution code.
 	// Depending on the context, ids.a may be a complex reference.
-	TestAssignCode string = "memory[ap] = ids.a"
+	testAssignCode string = "memory[ap] = ids.a"
+
+	// assert_le_felt() hints.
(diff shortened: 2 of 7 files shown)


quasilyte:
The tests are probably going to fail now, see #202 
Perhaps we need to use some other testing strategy for the cases like this.
Or maybe the test file's code itself should be fixed.

pkg/hintrunner/zero/hint_reference_resolver.go
```diff
@@ -33,6 +33,20 @@ func (m *hintReferenceResolver) GetReference(name string) (hinter.Reference, err
 	return nil, fmt.Errorf("missing reference %s", name)
 }
 
+// GetResOperander returns the result of GetReference type-asserted to ResOperander.
+// If reference is not found or it's not of ResOperander type, a non-nil error is returned.
+func (m *hintReferenceResolver) GetResOperander(name string) (hinter.ResOperander, error) {
```
cicr99, line 38: It'll probably be good to add the same check for getting CellRef types

pkg/hintrunner/zero/hintcode.go
```diff
@@ -7,4 +7,10 @@ const (
 	// the identifier resolution code.
 	// Depending on the context, ids.a may be a complex reference.
 	TestAssignCode string = "memory[ap] = ids.a"
+
+	// assert_le_felt() hints.
+	assertLeFelt          string = "import itertools\n\nfrom starkware.cairo.common.math_utils import assert_integer\nassert_integer(ids.a)\nassert_integer(ids.b)\na = ids.a % PRIME\nb = ids.b % PRIME\nassert a <= b, f'a = {a} is not less than or equal to b = {b}.'\n\n# Find an arc less than PRIME / 3, and another less than PRIME / 2.\nlengths_and_indices = [(a, 0), (b - a, 1), (PRIME - 1 - b, 2)]\nlengths_and_indices.sort()\nassert lengths_and_indices[0][0] <= PRIME // 3 and lengths_and_indices[1][0] <= PRIME // 2\nexcluded = lengths_and_indices[2][1]\n\nmemory[ids.range_check_ptr + 1], memory[ids.range_check_ptr + 0] = (\n    divmod(lengths_and_indices[0][0], ids.PRIME_OVER_3_HIGH))\nmemory[ids.range_check_ptr + 3], memory[ids.range_check_ptr + 2] = (\n    divmod(lengths_and_indices[1][0], ids.PRIME_OVER_2_HIGH))"
```
cicr99, line 12: I would like to follow some formatting with the name of these constants. The first two are public (which I think we should change to private since they are only supposed to be used in this package) and have the suffix "_Code" (which I would like to add to the others as I think is more self-explanatory)

integration_tests/cairo_files/math.cairo
cicr99, line 1: Maybe we should remove this file until issue #202 is decided.
You could add instead some unit tests such as the ones in `hintrunner/core/hint_test.go` for the moment

pkg/hintrunner/zero/hint_reference_resolver.go
quasilyte: I haven't yet needed to handle CellRef types yet, so I wasn't in a hurry to add such a method.
Do you want it to be added now?

integration_tests/cairo_files/math.cairo
quasilyte: There is ~nothing new to test in terms of the `hint_test` level.
The cairo0 hint is implemented in terms of already implemented hints (they're tested in `hint_test`).

The only new layers are: hint identification (code matching), operands resolution via hint reference manager. I don't think that we can cover these using the same strategy as in `hint_test` file.

I removed the file though since making this kind of testing work may need some planning and discussions. We can go forward without addressing it right away.

pkg/hintrunner/zero/hint_reference_resolver.go
cicr99: Yes, you are right. I haven't had to use it yet either. We can add it when needed

cicr99 (approved):

Issue #204, opened 2024-02-14
DecodeMemory used in tests can panic on some inputs


quasilyte:
Right now it seems like the memory trace doesn't always have the highest memory address (index) in its last memory trace element.
This is an assumption we make when decoding the memory:
```
lasMemIndex := binary.LittleEndian.Uint64(content[lastContentInd : lastContentInd+addrSize])
```
This seems to be working with `layout=plain`, but it fails to decode the `layout=small` memory trace.

(Or maybe we're reading whole 8 bytes that can contain some bits for the "segment". Since the segment is always 0 in the simplest cases, it could work correctly. But I don't know the exact memory encoding yet.)

Here is an example:
```
addr=80
addr=81
addr=82
addr=1635
addr=1634
addr=1637
addr=1636
addr=83
addr=84
```
I would assume that values like 1635 come from a range_check builtin segment (my test uses it).

quasilyte:
There are 2 ways for the address to be encoded:
```
    def to_bytes(value: "MaybeRelocatable", n_bytes: int, byte_order: Endianness) -> bytes:                                             
        """                                                                                                                             
        Serializes RelocatableValue as:                                                                                                 
        1bit |   SEGMENT_BITS |   OFFSET_BITS                                                                                           
        1    |     segment    |   offset                                                                                                
        Serializes int as                                                                                                               
        1bit | num                                                                                                                      
        0    | num                                                                                                                      
        """
```
But for the use cases I care about right now this is not an issue, because all values are encoded using the second schema.

quasilyte:
In this particular case, it all happens due to the different segment relocation between the two runners.
Here is the cairo-run segments relocation table for this test:
```
Segment relocation table:
0     1
1     68
2     98
3     98
4     1634
5     2146
```
1634 is a range_check builtin.

har777:
Fixed in https://github.com/NethermindEth/cairo-vm-go/pull/397

Issue #164, opened 2023-11-14
Add Cairo Zero Hints


rodrigo-pino:
# Cairo Zero Hints

The following is a (big!) list of hints required by Cairo Zero. They might be an overlap between Cairo Zero hints and Cairo One hints #36 .

The list was extracted from sister VM in Go by LambdaClass where they [gathered all Pythonic hints](https://github.com/lambdaclass/cairo-vm_in_go/tree/main/pkg/hints/hint_codes).

:eyes: For specific instructions on how to implement these hints you can check out the [readme](https://github.com/NethermindEth/cairo-vm-go/blob/main/pkg/hintrunner/zero/README.md#implementing-new-cairo-0-hints) file in the zero package

## Uint256 Operations

- [x] #187
- [x] #219
- [x] #221
- [x] #223
- [x] #224
- [x] #226
- [x] #227
- [x] #561 
- [x] #563 
- [x] #562 
- [x] #614 

## Math Operations

- [x] IsNn: Checks if a value is non-negative within a specific range.
- [x] IsNnOutOfRange: Checks if the negation of a value (minus one) is non-negative within a specific range.
- [x] IsLeFelt: Determines if one value is less than or equal to another within a finite field.
- [x] AssertLtFelt: Asserts that one value is strictly less than another within a finite field.
- [x] #238
- [x] IsPositive: Checks if a value is positive within a specific range and prime context.
- [x] #239
- [x] SplitInt: Splits an integer based on a base and asserts the result is within a certain range.
- [x] SplitIntAssertRange: Asserts that the value for `SplitInt` is within the expected range.
- [x] AssertNotZero: Asserts that a value is not zero.
- [x] AssertNn: Asserts that a given integer is non-negative within a specified range.
- [x] AssertNotEqual: Asserts that two given values are not equal.
- [x] Assert250Bits: Asserts that a value is within the range of 250 bits.
- [x] AssertLeFelt: Asserts that one value is less than or equal to another within a finite field.
- [x] AssertLeFeltExcluded0: Custom assertion related to `AssertLeFelt`.
- [x] AssertLeFeltExcluded1: Another custom assertion related to `AssertLeFelt`.
- [x] AssertLeFeltExcluded2: Final custom assertion related to `AssertLeFelt`.
- [x] #240
- [x] #241
- [x] #242
- [x] #243
- [x] #544 
- [x] #554 
- [x] #558 
- [x] #559 
- [x] #560 
- [x] #564 
[...]

PR #207, opened 2024-02-16
implement is_nn Cairo0 hint

quasilyte:
This hint uses the assert_le_felt beneath it, but
it was implemented beforehand.

This PR has no tests included since #204 is not solved yet. I used a couple of Cairo0 scripts to test this functionality with a set of different arguments to cover both hints that are a part of `is_nn` function.
(One of them handles negatives while another is for the non-negatives.)

Refs #164

### DIFF
--- a/pkg/hintrunner/zero/zerohint.go
+++ b/pkg/hintrunner/zero/zerohint.go
@@ -7,7 +7,10 @@ import (
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/core"
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
 	zero "github.com/NethermindEth/cairo-vm-go/pkg/parsers/zero"
+	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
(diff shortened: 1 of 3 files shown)


quasilyte:
I added a couple of extra cairo0 hints to the PR.

quasilyte:
Refs #208

pkg/hintrunner/zero/zerohint.go
```diff
@@ -73,6 +84,95 @@ func CreateAllocSegmentHinter(resolver hintReferenceResolver) (hinter.Hinter, er
 	return &core.AllocSegment{Dst: hinter.ApCellRef(0)}, nil
 }
 
+func createIsLeFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
```
cicr99, line 87: This file is going to get big too fast. Other than what we had discussed to do Unit testing, we could also consider having separated files to group set of hints (as they are separated in the issue #164 for example). In this case, they could go in a `math.go` file.

cicr99 (approved):

pkg/hintrunner/zero/zerohint.go
quasilyte: I'll move it in the upcoming PR.

Issue #187, opened 2024-01-23
Implement Uint256Add Cairo 0 Hint


cicr99:
Adds two uint256 integers, calculating the sum of the lower and higher parts separately, and includes overflow checks.

Resources:
- CairoLang implementation (Python): https://github.com/starkware-libs/cairo-lang/blob/ab6d079f0265f903630aac9287e93d9e7cfb6fcf/src/starkware/cairo/common/uint256.cairo#L70
- LC implementation: https://github.com/lambdaclass/cairo-vm_in_go/blob/main/pkg/hints/uint256_hints.go

PR #210, opened 2024-02-19
Get address for ResOperands

### LINKED ISSUE #206: Handle addresses for ResOperands and Uint256 references
I encountered a few issues regarding the work on Uint256 cairo 0 hints.
Usually, when we were to create a new Hinter type (for Cairo 1) you got ResOperanders in the parameters and CellRefs where you were suppose to write the results of the operation, for example Uint256_add hint has the following variables involved:
```
"reference_ids": {
                        "starkware.cairo.common.uint256.uint256_add.a": 3,
                        "starkware.cairo.common.uint256.uint256_add.b": 4,
                        "starkware.cairo.common.uint256.uint256_add.carry_high": 8,
                        "starkware.cairo.common.uint256.uint256_add.carry_low": 7,
                        "starkware.cairo.common.uint256.uint256_add.range_check_ptr": 5,
                        "starkware.cairo.common.uint256.uint256_add.res": 6
                    }
```
`a` and `b`  are the parameters you get in uint256_add method, so one would expect them to be resOperands as you would like to extract their value. However `carry_high` and `carry_low`  both are supposed to be assigned in the hint implementation, so one would expect to get a CellRefer to write to. But in the cairo zero hint format all of them are Derefs which are ResOperanders. The problem with this is that ResOperanders only implement `Resolve` function which returns the accessed value instead of the address being accessed, so we would need to make some hack to get the address and write to it.

The other thing I noticed is that `a` and `b` are both uint256 variables. In cario 1 hints you would get uint256 values separated into low and high, but here they just give you one address, which would correspond to the low part, the (address + 1) would correspond to high part. Since both parameters are ResOperanders, getting their value only would yield the low part, so we still have the issue I mentioned before, we need to access the address to compute the address of the high part and get the complete result.

Proposed solution:
Add a method Get as well for ResOperanders, it would return the address in the case of Derefs and DoubleDerefs, and an error on Binary Operations and Immediate values, as you can't get an address from them. For the uint256 values, we could also implement something like GetUint256, which would handle all the address computation.

cicr99:
Fix #206 
The handling of addresses for Uint256 elements will be solved as part of #187

### DIFF
--- a/pkg/hintrunner/hinter/operand.go
+++ b/pkg/hintrunner/hinter/operand.go
@@ -55,6 +55,7 @@ type ResOperander interface {
 	fmt.Stringer
 
 	ApplyApTracking(hint, ref zero.ApTracking) Reference
+	GetAddress(vm *VM.VirtualMachine) (mem.MemoryAddress, error)
 	Resolve(vm *VM.VirtualMachine) (mem.MemoryValue, error)
 }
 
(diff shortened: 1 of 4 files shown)


quasilyte (approved):

cicr99:
After the first commit some test were failing: Adding a method `Get` for the ResOperanders causes them to implement the CellRefer interface as well, which is a problem for all the logic in the project that assumes CellRefers and ResOperanders are separated. So, decided to just change the name of the method to be `GetAddress`

PR #211, opened 2024-02-20
pkg/hintrunner/zero: add unit tests for the existing Cairo0 hints

quasilyte:
Also split hint constructor and Cairo0 hint wrapper code to make unit testing easier.


pkg/hintrunner/zero/zerohint_test.go
```diff
@@ -0,0 +1,385 @@
...
+			switch o.kind {
+			case apRelative, fpRelative:
+				o.memoryOffset = vm.Context.Ap
+				runnerutil.WriteTo(vm, VM.ExecutionSegment, vm.Context.Ap, memory.MemoryValueFromFieldElement(o.felt))
+				vm.Context.Ap++
+
+			case immediate:
+				// Nothing to do.
+
+			default:
```
cicr99, line 332: It appears you are not including DoubleDeref or BinaryOperation types, could you explain the issue with those?

cicr99 (commented):
Looks very good.

pkg/hintrunner/zero/zerohint_test.go
quasilyte: I wasn't sure whether to go on with that or not.
It's good if we want to test the addressing modes along the way, but perhaps it's easier to have these tests to be separated.

Or we can select the addressing mode from a table with all combinations for every test, just to increase the harness. This would remove the operander kind from the test case description object.

pkg/hintrunner/zero/zerohint_test.go
```diff
@@ -0,0 +1,385 @@
...
+
+		if tc.check == nil {
+			// Allow omitting a VM state check if errCheck is present.
+			if tc.errCheck == nil {
+				t.Fatal("check and errCheck can't both be nil")
+			}
+		} else {
+			tc.check(t, testCtx)
+		}
+	}
```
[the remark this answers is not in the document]
quasilyte: I don't know yet if there is any merit in doing both, but this code runs both functions if they're present.
The `both nil` test is a sanity check.
It's possible to only allow one of the two and make the code a little simpler.

pkg/hintrunner/zero/zerohint_test.go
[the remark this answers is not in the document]
quasilyte: Sounds good to me. I'm on it.

pkg/hintrunner/zero/zerohint_test.go
quasilyte: These operander types can be important for ones where you need to get their address. It won't work with immediate, etc.

pkg/hintrunner/zero/zerohint_test.go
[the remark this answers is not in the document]
cicr99: It doesn't do any harm to leave it as it is now. I was just asking in case there was no point on running both of them, but maybe it could be of interest in some case

pkg/hintrunner/zero/zerohint_test.go
cicr99: I don't think trying all the combinations in every test is the best idea. For these tests is ok to only use the current types (as it is), I actually haven't dealt with any different types in my examples yet. 
I was only wondering if there was any other reason not to add them here. Let's leave it like this and have separate ones to check we are getting the address correctly.

cicr99 (approved):

PR #212, opened 2024-02-22
Uint256Add hinter

### LINKED ISSUE #187: Implement Uint256Add Cairo 0 Hint

cicr99:
Fixes #187

### DIFF
--- a/pkg/hintrunner/hinter/operand.go
+++ b/pkg/hintrunner/hinter/operand.go
@@ -229,3 +229,27 @@ func (v Immediate) ApplyApTracking(hint, ref zero.ApTracking) Reference {
 	// Nothing to do
 	return v
 }
+
+func GetConsecutiveValues(vm *VM.VirtualMachine, ref ResOperander, off int16) ([]mem.MemoryValue, error) {
+	addr, err := ref.GetAddress(vm)
+	if err != nil {
+		return nil, err
+	}
+
+	var values []mem.MemoryValue
+	for i := int16(0); i < off; i++ {
+		nAddr, err := addr.AddOffset(i)
+		if err != nil {
+			return nil, err
+		}
+
+		v, err := vm.Memory.ReadFromAddress(&nAddr)
+		if err != nil {
+			return nil, err
+		}
+
+		values = append(values, v)
+	}
+
+	return values, nil
+}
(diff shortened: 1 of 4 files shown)


cicr99:
I added a few modifications to the implementation of Uint256Add hint in the last commit, so we could use the same one for Uint256AddLow hint, as most of the logic is the same for both

quasilyte:
Doesn't the current 1 field per operander approach work in this case?

cicr99:
> Doesn't the current 1 field per operander approach work in this case?

Unfortunately no. The hint uses `a` and `b` as the parameters for the addition, and they are both of type `Uint256`, so it's actually two felts each one.

pkg/hintrunner/zero/zerohint_uint256_test.go
```diff
@@ -0,0 +1,87 @@
+package zero
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+	// "github.com/NethermindEth/cairo-vm-go/pkg/utils"
```
quasilyte, line 9: A non-blocking nit: this can be removed later if it's easier to merge right away.

cicr99: I was using it at some point, then it wasn't necessary anymore but thought I would have to use it again. Of course, I ended up forgetting about it.

pkg/hintrunner/hinter/operand.go
```diff
@@ -229,3 +229,27 @@ func (v Immediate) ApplyApTracking(hint, ref zero.ApTracking) Reference {
...
 	return v
 }
+
+func GetConsecutiveValues(vm *VM.VirtualMachine, ref ResOperander, off int16) ([]mem.MemoryValue, error) {
+	addr, err := ref.GetAddress(vm)
+	if err != nil {
+		return nil, err
+	}
+
+	var values []mem.MemoryValue
```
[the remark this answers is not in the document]
cicr99: You are right, I'll change it

PR #214, opened 2024-02-26
pkg/hintrunner/zero: add is_positive hint

quasilyte:
Also add an unitialized operander kind for testing.


quasilyte:
I suggest refactoring the repeating pattern of `resolve + FieldElement` combination in the next PR.

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -225,3 +225,149 @@ func createIsNNOutOfRangeHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+
+			value, err := value.Resolve(vm)
+			if err != nil {
+				return err
+			}
+			valueFelt, err := value.FieldElement()
+			if err != nil {
+				return err
+			}
+			if valueFelt.IsZero() {
```
cicr99, line 287: Shouldn't we check the opposite here? If `valueFelt` is zero then the assertion is correct, so no need to return this error.

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -225,3 +225,149 @@ func createIsNNOutOfRangeHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+			bound, err := bound.Resolve(vm)
+			if err != nil {
+				return err
+			}
+			boundFelt, err := bound.FieldElement()
+			if err != nil {
+				return err
+			}
+
+			result := utils.FeltMod(valueFelt, baseFelt)
```
cicr99, line 343: Quick question: I saw you omitted this step `int(ids.value) % PRIME`, is it safe to do so?
I understand what is stored in `value` should fit in a felt, in that case it wouldn't be necessary to do it. Should we add a comment clarifying this?

pkg/hintrunner/zero/zerohint_test.go
```diff
@@ -77,9 +77,18 @@ const (
 	// $value | Immediate($value)
 	// Requires {Name, Kind=immediate, Value}
 	immediate
+
+	// A memory cell that is allocated, but not written to yet.
+	// It's allowed to write to this address once.
+	// Requires {Name, Kind=unitialized}
+	unitialized
```
cicr99, line 84: Typo here _uninitialized_?

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -193,5 +193,154 @@ func TestZeroHintMath(t *testing.T) {
...
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newIsPositiveHint(ctx.operanders["value"], ctx.operanders["is_positive"])
+				},
+				check: varValueEquals("is_positive", feltInt64(0)),
+			},
+		},
+
+		"SplitIntAssertRange": {
+			// Zero value - assertion failed, any other - nothing.
```
cicr99, line 266: If the comment I made above about the implementation of this hint is correct, then these tests should be fixed as well

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -225,3 +225,149 @@ func createIsNNOutOfRangeHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+			fmt.Printf("result=%v value=%v base=%v bound=%v\n", &result, valueFelt, baseFelt, boundFelt)
```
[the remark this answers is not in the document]
quasilyte: Yup :D

pkg/hintrunner/zero/zerohint_test.go
quasilyte: True

pkg/hintrunner/zero/zerohint_math.go
quasilyte: I commented on it in one of the places and decided that it could be excessive here.
All operations on felts we use are module prime - it's stated in the felt comments.

pkg/hintrunner/zero/zerohint_math.go
quasilyte: Good catch!

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -225,3 +225,149 @@ func createIsNNOutOfRangeHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+
+func createIsPositiveHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
+	// This hint is executed for the negative values.
+	// If the value was non-negative, it's usually handled by the IsNN hint.
+
+	a, err := resolver.GetResOperander("a")
+	if err != nil {
+		return nil, err
+	}
+	return newIsNNOutOfRangeHint(a), nil
```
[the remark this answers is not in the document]
quasilyte: Fixed. Since tests use newIsPositiveHint directly, it was easy to overlook this issue. :(

pkg/hintrunner/zero/zerohint_math_test.go
quasilyte: Done

cicr99 (approved):

PR #215, opened 2024-02-28
pkg/hintrunner/zero: refactor Resolve+FieldElement idiom

quasilyte:
Since we use this combination that often, it's worth moving it to a new function so we can handle the error exactly once per every operander (instead of twice).


cicr99 (approved):
lgtm

PR #218, opened 2024-02-28
replace resolveFieldElement with hinter.ResolveAsFelt

quasilyte:
Based on a hint from @cicr99

### DIFF
--- a/pkg/hintrunner/hinter/utils.go
+++ b/pkg/hintrunner/hinter/utils.go
@@ -8,32 +8,32 @@ import (
 	f "github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
-func ResolveAsAddress(vm *VM.VirtualMachine, op ResOperander) (mem.MemoryAddress, error) {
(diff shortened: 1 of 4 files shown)


pkg/hintrunner/hinter/utils.go
```diff
@@ -8,32 +8,32 @@ import (
 	f "github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
-func ResolveAsAddress(vm *VM.VirtualMachine, op ResOperander) (mem.MemoryAddress, error) {
+func ResolveAsAddress(vm *VM.VirtualMachine, op ResOperander) (*mem.MemoryAddress, error) {
 	mv, err := op.Resolve(vm)
 	if err != nil {
-		return mem.UnknownAddress, fmt.Errorf("%s: %w", op, err)
+		return &mem.UnknownAddress, fmt.Errorf("%s: %w", op, err)
```
cicr99, line 14: why not use `nil` here as well?

cicr99 (approved):

pkg/hintrunner/hinter/utils.go
quasilyte: Let's make it nil then.

PR #229, opened 2024-03-01
pkg/hintrunner/zero: implement assert250bits Cairo0 hint

quasilyte:
This hint required a builtins support from the test runner.

I tested this hint with our VM as well.
The results for tests are gathered from the cairo-run executions for the same values.

### DIFF
--- a/pkg/hintrunner/zero/zerohint_math_test.go
+++ b/pkg/hintrunner/zero/zerohint_math_test.go
@@ -4,7 +4,9 @@ import (
 	"testing"
 
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	"github.com/NethermindEth/cairo-vm-go/pkg/parsers/starknet"
 	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
 func TestZeroHintMath(t *testing.T) {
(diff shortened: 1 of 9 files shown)


pkg/utils/math_test.go
```diff
@@ -33,3 +36,47 @@ func TestOffsetRightNoOverflow(t *testing.T) {
...
+		var b big.Int
+		var div big.Int
+		var rem big.Int
+		test.a.BigInt(&a)
+		test.b.BigInt(&b)
+		div.DivMod(&a, &b, &rem)
+
+		wantDiv := new(fp.Element).SetBigInt(&div)
+		wantRem := new(fp.Element).SetBigInt(&rem)
+
```
cicr99, line 76: Even though I like the idea of checking random values, I don't think this is very helpful in this case, as the way to calculate the expected result is the same code you want to test. I rather just go with the fixed cases

pkg/hintrunner/zero/zerohint_test.go
```diff
@@ -179,6 +220,22 @@ func runHinterTests(t *testing.T, tests map[string][]hintTestCase) {
 
 			case immediate:
 				testCtx.operanders[o.Name] = hinter.Immediate(*o.Value.(*fp.Element))
+
+			case reference:
+				switch value := o.Value.(type) {
```
cicr99, line 225: Should we add a `default` case for this switch?

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -479,5 +481,94 @@ func TestZeroHintMath(t *testing.T) {
...
+					{Name: "value", Kind: apRelative, Value: feltString("348329493943842849393124453993999999231222222222")},
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newAssert250bitsHint(ctx.operanders["low"], ctx.operanders["high"], ctx.operanders["value"])
+				},
+				check: allVarValueEquals(map[string]*fp.Element{
+					"low":  feltString("302658603189151847763334509038790380942"),
+					"high": feltInt64(1023648380),
+				}),
+			},
```
cicr99, line 559: This test case is the same as the one right above right?

pkg/utils/math.go
```diff
@@ -96,3 +96,25 @@ func FeltMod(a, b *fp.Element) fp.Element {
 	result.SetBigInt(&tmpResult)
 	return result
 }
+
+func FeltDivRem(a, b *fp.Element) (div fp.Element, rem fp.Element) {
+	// It would be possible to compute the mod as `a - div*b`,
```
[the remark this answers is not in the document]
quasilyte: Yes. I think I used both of these terms to provide a synonym, but perhaps it creates unnecessary confusion.

pkg/hintrunner/zero/zerohint_math_test.go
quasilyte: I will add a small sanity check that would protect us from such mistakes in the future.

cicr99 (approved):

PR #236, opened 2024-03-07
Implement Uint256SignedNN hint

MaksymMalicki:


pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -190,3 +190,36 @@ func createSplit64Hinter(resolver hintReferenceResolver) (hinter.Hinter, error)
 
 	return newSplit64Hint(a, low, high), nil
 }
+
+func newUint256SignedNN(a hinter.ResOperander) hinter.Hinter {
```
cicr99, line 194: Just to follow the same format as the others, the name of the function should be `new<HintName>Hint`

pkg/hintrunner/zero/zerohint_uint256_test.go
```diff
@@ -131,5 +131,37 @@ func TestZeroHintUint256(t *testing.T) {
 				}),
 			},
 		},
+		"Uint256SignedNN": {
+			{
+				operanders: []*hintOperander{
+					{Name: "a.low", Kind: fpRelative, Value: feltUint64(0)},
+					{Name: "a.high", Kind: fpRelative, Value: feltInt64(-5)},
```
cicr99, line 138: These values are not valid if you want to construct a Uint256 as it is in this case. The Uint256 structure takes two felts but both of them should be smaller than 2**128. The idea is that the result of `high << 128 + low` fits in 256 bits.
If you use -5, the felt value for that would probably be around the 252 bits

pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -190,3 +190,36 @@ func createSplit64Hinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+			// memory[ap] = 1 if 0 <= (ids.a.high % PRIME) < 2 ** 127 else 0
+			apAddr := vm.Context.AddressAp()
+
+			_, aHigh, err := GetUint256AsFelts(vm, a)
+			if err != nil {
+				return err
+			}
+			var v memory.MemoryValue
+			felt127 := new(fp.Element).SetBigInt(new(big.Int).Lsh(big.NewInt(1), 127))
```
cicr99, line 206: Maybe we could add this to `utils/constant.go` file, as it is used in other places as well

cicr99 (changes requested):

pkg/hintrunner/zero/zerohint_uint256.go
MaksymMalicki: corrected

pkg/hintrunner/zero/zerohint_uint256.go
MaksymMalicki: I think it's a good idea, I have seen it in some tests. I will add felt127 const in this PR, but maybe a new PR with replacements.

cicr99 (approved):

cicr99:
Once you address the conflicts with `main` branch, it should be good to go

pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -190,3 +190,36 @@ func createSplit64Hinter(resolver hintReferenceResolver) (hinter.Hinter, error)
 
 	return newSplit64Hint(a, low, high), nil
 }
+
+func newUint256SignedNNHint(a hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "Uint256SignedNN",
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+			// memory[ap] = 1 if 0 <= (ids.a.high % PRIME) < 2 ** 127 else 0
```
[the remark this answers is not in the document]
MaksymMalicki: I have corrected that. I agree, that I should stick with the convention

Issue #237, opened 2024-03-08
Starknet contracts execution and blockifier-like API discussion


quasilyte:
This is a work-in-progress document. It may be extended further, or perhaps it will become redundant after some point (and will be replaced by a proper design document).
Right now it's more of a TODO list.

To create an alternative for a Blockifier, a current state of a Blockifier needs to be analyzed in the context of its usage inside Juno.

## Dependencies

The API surface is small, but since the respective documentation is lacking on the both sides, it's hard to get going without describing a high-level picture first.

Juno uses blockifier as a transaction execution component. Since Juno is written in Go and blockifier in Rust, some FFI interfaces are used to make it work.
Juno repository contains a small amount of Rust code that implements the 2-function API for the Go side:

* cairoVMCall
* cairoVMExecute

These two functions are then called via CGo. These functions use blockifier inside.
Blockifier symbols used from Juno (Rust side):

* `blockifier::block::BlockInfo`
* `blockifier::block::BlockNumberHashPair`
* `blockifier::block::GasPrices`
* `blockifier::context::BlockContext`
* `blockifier::context::ChainInfo`
* `blockifier::context::FeeTokenAddresses`
* `blockifier::execution::contract_class::ClassInfo`
* `blockifier::execution::entry_point::CallEntryPoint`
* `blockifier::execution::entry_point::CallType`
* `blockifier::execution::entry_point::EntryPointExecutionContext`
* `blockifier::state::cached_state::CachedState`
* `blockifier::state::cached_state::GlobalContractCache`
* `blockifier::state::state_api::State`
* `blockifier::transaction::errors::TransactionExecutionError::ContractConstructorExecutionFailed`
* `blockifier::transaction::errors::TransactionExecutionError::ExecutionError`
* `blockifier::transaction::errors::TransactionExecutionError::ValidateTransactionError`
* `blockifier::transaction::objects::DeprecatedTransactionInfo`
* `blockifier::transaction::objects::HasRelatedFeeType`
* `blockifier::transaction::objects::TransactionInfo`
* `blockifier::transaction::transaction_execution::Transaction`
* `blockifier::transaction::transactions::ExecutableTransaction`
* `blockifier::versioned_constants::VersionedConstants`
* `blockifier::block::pre_process_block`
* `blockifier::fee::fee_utils::calculate_tx_fee`

It also uses these symbols from the cairo_vm crate:

* `cairo_vm::vm::runners::cairo_runner::ExecutionResources`

[...]

PR #272, opened 2024-03-12
pkg/runners/zero: add a way to use PC offset as an entrypoint

quasilyte:
The Blockifier-style API uses the PC-based contract class entry point.

See #237 for more context


cicr99 (approved):
Looks good!
Should we add some kind of test for this? What could be a good one?

quasilyte:
We can probably duplicate all normal runs with PC-based executions.
The compiled file contains the main function PC which can be used as an argument.

PR #304, opened 2024-03-13
Cairo0 hint split felt

MaksymMalicki:


pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -501,3 +502,99 @@ func createSplitIntHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+func createSplitFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
+	maxHigh, err := resolver.GetResOperander("MAX_HIGH")
+	if err != nil {
+		return nil, err
+	}
+
+	maxLow, err := resolver.GetResOperander("MAX_LOW")
+	if err != nil {
+		return nil, err
+	}
```
cicr99, line 582: These two variables are constants, if you check the compiled output of a file that's using this hint, you'll see that they are not actually parameters you need to pass to the hint. The json looks like this:
```
   "15": [
            {
                "accessible_scopes": [
                    "starkware.cairo.common.math",
                    "starkware.cairo.common.math.split_felt"
                ],
                "code": "from starkware.cairo.common.math_utils import assert_integer\nassert ids.MAX_HIGH < 2**128 and ids.MAX_LOW < 2**128\nassert PRIME - 1 == ids.MAX_HIGH * 2**128 + ids.MAX_LOW\nassert_integer(ids.value)\nids.low = ids.value & ((1 << 128) - 1)\nids.high = ids.value >> 128",
                "flow_tracking_data": {
                    "ap_tracking": {
                        "group": 4,
                        "offset": 0
                    },
                    "reference_ids": {
                        "starkware.cairo.common.math.split_felt.high": 10,
                        "starkware.cairo.common.math.split_felt.low": 9,
                        "starkware.cairo.common.math.split_felt.range_check_ptr": 11,
                        "starkware.cairo.common.math.split_felt.value": 7
                    }
                }
            }
        ]
```
I would suggest you define these variables inside the hint implementation instead. Fixing this will require you to modify also the constructor of this hint and the test cases

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -501,3 +502,99 @@ func createSplitIntHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
 	}
 	return newSplitIntHint(output, value, base, bound), nil
 }
+
+func newSplitFeltHint(maxHigh, maxLow, low, high, value hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "SplitFelt",
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+			//> from starkware.cairo.common.math_utils import assert_integer\nassert ids.MAX_HIGH < 2**128 and ids.MAX_LOW < 2**128\nassert PRIME - 1 == ids.MAX_HIGH * 2**128 + ids.MAX_LOW\nassert_integer(ids.value)\nids.low = ids.value & ((1 << 128) - 1)\nids.high = ids.value >> 128
```
cicr99, line 510: Could you actually use the enters instead of having `\n` in the string. It would be a lot more readable

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -555,5 +555,79 @@ func TestZeroHintMath(t *testing.T) {
 				errCheck: errorTextContains("outside of the range [0, 2**250)"),
 			},
 		},
+
+		"SplitFelt": {
+			{
+				operanders: []*hintOperander{
+					{Name: "MAX_HIGH", Kind: apRelative, Value: &utils.FeltMax128},
+					{Name: "MAX_LOW", Kind: apRelative, Value: &utils.FeltZero},
```
cicr99, line 563: The values for `MAX_HIGH` and `MAX_LOW`, as mentioned in the previous comment are always the same:
```
// Note: the following code works because PRIME - 1 is divisible by 2**128.
const MAX_HIGH = (-1) / 2 ** 128;
const MAX_LOW = 0;
```
Same for the other test cases. Anyway, when you solve the previous comment this should be removed from the test case.

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -555,5 +555,79 @@ func TestZeroHintMath(t *testing.T) {
...
+					{Name: "low", Kind: uninitialized},
+					{Name: "high", Kind: uninitialized},
```
cicr99, line 565: `low` and `high` are supposed to be references to `range_check_ptr` builtin. In the code of the compiler it looks like this:
```
 // Guess the low and high parts of the integer.
let low = [range_check_ptr];
let high = [range_check_ptr + 1];
let range_check_ptr = range_check_ptr + 2;
```

See `Assert250bits` implementation to see how this is handled

cicr99 (changes requested):

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: Modified the code, so that each line is in the new line.

pkg/hintrunner/zero/zerohint_math_test.go
MaksymMalicki: As previously mentioned, removed it from testcases and fixed in methods

pkg/hintrunner/zero/zerohint_math_test.go
MaksymMalicki: I have noticed, that in the testcase above I did it correctly, so that must be some copypaste typo. Didn't catch that however.

cicr99 (approved):
lgtm

PR #305, opened 2024-03-13
Add felt127 const

MaksymMalicki:


cicr99 (commented):
It would be good that you also modify the code in the places that are creating this value, to use this variable instead

PR #306, opened 2024-03-13
Cairo0 hint pow

MaksymMalicki:


pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -555,5 +555,39 @@ func TestZeroHintMath(t *testing.T) {
 				errCheck: errorTextContains("outside of the range [0, 2**250)"),
 			},
 		},
+		"Pow": {
+			{
+				operanders: []*hintOperander{
+					{Name: "locs.bit", Kind: uninitialized},
```
cicr99, line 561: I'd like it more if you passed this operander as the last one if you are not passing the entire struct. As the memory  in the test is initialized in the same order these operanders are given, then right after `locs.bit` slot you would write the `prev_locs` struct, which is not exactly accurate as the memory for the rest of the members of `locs` should also have been reserved. I understand that for testing purposes this is not that relevant, but I would like it to be as much realistic as possible. Hence, putting it at the end sounds better to me, even if you do not declare the entire struct. Does this make sense?

MaksymMalicki: Absolutely. So if I understand correctly, if I put "locs" instead of "locs.bit", I would allocate the entire necessary space for the struct?
Changed that, so "locs.bit" is last

cicr99: Each element on the list `[]*hintOperander` will allocate only one memory slot, that's why when we want to allocate a struct we pass one for each field. The name you use won't matter as long as you pass the correct address to the constructor of the hint. 
What I meant with my comment above, is that if `locs` memory was reserved before the `prev_locs`, there would be five slots on memory between the first field of `locs` (`locs.bit`) and the first of `prev_locs` (`prev_locs.bit`), and that was not happening in your test.
The fix you made is ok

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -505,6 +505,59 @@ func createSplitIntHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+	if err != nil {
+		return nil, err
+	}
+	prev_locs, err := resolver.GetCellRefer("prev_locs")
+	if err != nil {
+		return nil, err
+	}
+	locsRes := hinter.Deref{Deref: locs}
+	prevLocsRes := hinter.Deref{Deref: prev_locs}
+	return newPowHint(locsRes, prevLocsRes), nil
```
cicr99, line 558: This part I don't like so much. The `newPowHint` method does not have to comply with a specific interface so in reality you could pass whichever parameter types you need. The problem is with the testing interface, right? Because we create only `Deref` structs for the operanders.
I would prefer a solution where the testing suit allows us to pass the types we need, we shouldn't have to wrap the operand just to unwrap it later. Also our decisions regarding implementation shouldn't be determined by the way we do testing, but the other way around.
However, this would imply we have to modify the current testing structure, so let's leave this as it is and implement a solution in another PR.

Just created an issue for it https://github.com/NethermindEth/cairo-vm-go/issues/332

cicr99 (approved):

cicr99 (approved):

PR #307, opened 2024-03-13
Cairo0 hint uint256 sqrt

MaksymMalicki:


pkg/hintrunner/zero/hintcode.go
```diff
@@ -38,4 +38,5 @@ const (
 	uint256AddCode    string = "sum_low = ids.a.low + ids.b.low\nids.carry_low = 1 if sum_low >= ids.SHIFT else 0\nsum_high = ids.a.high + ids.b.high + ids.carry_low\nids.carry_high = 1 if sum_high >= ids.SHIFT else 0"
 	uint256AddLowCode string = "sum_low = ids.a.low + ids.b.low\nids.carry_low = 1 if sum_low >= ids.SHIFT else 0"
 	split64Code       string = "ids.low = ids.a & ((1<<64) - 1)\nids.high = ids.a >> 64"
+	uint256Sqrt       string = "from starkware.python.math_utils import isqrt\nn = (ids.n.high << 128) + ids.n.low\nroot = isqrt(n)\nassert 0 <= root < 2 ** 128\nids.root.low = root\nids.root.high = 0"
```
quasilyte, line 41: Missing the Code suffix in the const name.

pkg/utils/arrays.go
```diff
@@ -5,3 +5,10 @@ func Reverse[T any](a []T) {
 		a[i], a[j] = a[j], a[i]
 	}
 }
+
```
quasilyte, line 8: I would suggest using the stdlib function for that.
https://pkg.go.dev/slices#Reverse

MaksymMalicki: Didn't notice the stdlib function. Replaced

pkg/hintrunner/zero/hintcode.go
MaksymMalicki: Added code suffix

pkg/hintrunner/zero/zerohint_uint256_test.go
```diff
@@ -131,5 +131,67 @@ func TestZeroHintUint256(t *testing.T) {
...
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newUint256SqrtHint(ctx.operanders["n.low"], ctx.operanders["root.low"])
+				},
+				check: allVarValueEquals(map[string]*fp.Element{
+					"root.low":  feltUint64(4),
+					"root.high": feltUint64(0),
+				}),
+			},
+		},
 	})
```
cicr99, line 196: I think this last test case is not necessary as you are already checking with another value that is simple (the previous case). I'd rather you add a test for a bigger number, one that doesn't fit in a felt either, just in a uint256. Other test case that could be useful would be one in which `n` is not a square.

MaksymMalicki: Added test cases for bigger numbers, this was a very good call, as I have found critical error, so TY

pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -190,3 +192,64 @@ func createSplit64Hinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+				return err
+			}
+			//> ids.root.low = root
+			rootLowValue := memory.MemoryValueFromFieldElement(&calculatedFeltRoot)
+			err = vm.Memory.WriteToAddress(&rootAddr, &rootLowValue)
+			if err != nil {
+				return err
+			}
+			//> ids.root.high = 0
+			rootHighValue := memory.MemoryValueFromFieldElement(new(fp.Element).SetInt64(0))
```
cicr99, line 239: Could you use FeltZero constant from utils here?

pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -190,3 +192,64 @@ func createSplit64Hinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+			return hinter.WriteToNthStructField(vm, root, rootHighValue, 1)
```
cicr99, line 240: If we merge #308 first then you can use `WriteUint256ToAddress` utility here for writing `root` to memory

MaksymMalicki: yes, absolutely

pkg/hintrunner/zero/zerohint_uint256.go
MaksymMalicki: done

pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -190,3 +192,64 @@ func createSplit64Hinter(resolver hintReferenceResolver) (hinter.Hinter, error)
 
 	return newSplit64Hint(a, low, high), nil
 }
+
+func newUint256SqrtHint(n hinter.ResOperander, root hinter.ResOperander) hinter.Hinter {
+	// from starkware.python.math_utils import isqrt\nn = (ids.n.high << 128) + ids.n.low\nroot = isqrt(n)\nassert 0 <= root < 2 ** 128\nids.root.low = root\nids.root.high = 0
```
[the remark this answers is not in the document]
MaksymMalicki: done

cicr99 (approved):

PR #308, opened 2024-03-13
Implement Uint256UnsignedDivRem hint and test cases

MaksymMalicki:


pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -223,3 +223,99 @@ func createUint256SignedNNHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+	return &GenericZeroHinter{
+		Name: "Uint256UnsignedDivRem",
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+			//> a = (ids.a.high << 128) + ids.a.low
+			//div = (ids.div.high << 128) + ids.div.low
+			//quotient, remainder = divmod(a, div)
+			//ids.quotient.low = quotient & ((1 << 128) - 1)
+			//ids.quotient.high = quotient >> 128
+			//ids.remainder.low = remainder & ((1 << 128) - 1)
+			//ids.remainder.high = remainder >> 128
```
cicr99, line 237: Just for format - Missing `>` on the hint code lines

pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -223,3 +223,99 @@ func createUint256SignedNNHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+			}
+			x := memory.MemoryValueFromFieldElement(lowQuot)
+			err = vm.Memory.WriteToAddress(&quotientAddr, &x)
+			if err != nil {
+				return err
+			}
+			err = hinter.WriteToNthStructField(vm, quotient, memory.MemoryValueFromFieldElement(highQuot), 1)
+			if err != nil {
+				return err
+			}
```
cicr99, line 286: You've had to use this logic several times, I'd suggest you create a utility for uint256 values, something like `WriteUint256ToAddress()`, that does this

pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -223,3 +223,99 @@ func createUint256SignedNNHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+			var divLowBig big.Int
+			divLow.BigInt(&divLowBig)
+			var divHighBig big.Int
+			divHigh.BigInt(&divHighBig)
+
+			a := new(big.Int).Add(new(big.Int).Lsh(&aHighBig, 128), &aLowBig)
+			div := new(big.Int).Add(new(big.Int).Lsh(&divHighBig, 128), &divLowBig)
+			quot, rem := utils.FeltDivRem(new(fp.Element).SetBigInt(a), new(fp.Element).SetBigInt(div))
```
cicr99, line 258: This should not be working properly, a uint256 number (256 bits) doesn't have to fit in a felt (252 bits). I think they do some compression when the number doesn't fit in a felt, but not sure how that works or how can it affect the outcome. See if you can use instead uint256 library as in #307, or if you can perform the operation as big.Int

pkg/hintrunner/zero/hintcode.go
```diff
@@ -35,10 +35,11 @@ const (
...
-	uint256AddLowCode string = "sum_low = ids.a.low + ids.b.low\nids.carry_low = 1 if sum_low >= ids.SHIFT else 0"
-	split64Code       string = "ids.low = ids.a & ((1<<64) - 1)\nids.high = ids.a >> 64"
-	uint256SignedNN   string = "memory[ap] = 1 if 0 <= (ids.a.high % PRIME) < 2 ** 127 else 0"
+	uint256AddCode        string = "sum_low = ids.a.low + ids.b.low\nids.carry_low = 1 if sum_low >= ids.SHIFT else 0\nsum_high = ids.a.high + ids.b.high + ids.carry_low\nids.carry_high = 1 if sum_high >= ids.SHIFT else 0"
+	uint256AddLowCode     string = "sum_low = ids.a.low + ids.b.low\nids.carry_low = 1 if sum_low >= ids.SHIFT else 0"
+	split64Code           string = "ids.low = ids.a & ((1<<64) - 1)\nids.high = ids.a >> 64"
+	uint256SignedNN       string = "memory[ap] = 1 if 0 <= (ids.a.high % PRIME) < 2 ** 127 else 0"
+	uint256UnsignedDivRem string = "a = (ids.a.high << 128) + ids.a.low\ndiv = (ids.div.high << 128) + ids.div.low\nquotient, remainder = divmod(a, div)\nids.quotient.low = quotient & ((1 << 128) - 1)\nids.quotient.high = quotient >> 128\nids.remainder.low = remainder & ((1 << 128) - 1)\nids.remainder.high = remainder >> 128"
```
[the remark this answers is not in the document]
MaksymMalicki: done

pkg/hintrunner/zero/zerohint_uint256.go
MaksymMalicki: done

cicr99 (approved):

pkg/hintrunner/zero/zerohint_uint256.go
cicr99: Yes, those are good. Another way was to use `DivMod()` function from bigInt library, but it probably does the exact same thing

Issue #284, opened 2024-03-13
Unexpected "compute res: cannot multiply memory addresses" error while running the code


quasilyte:
While implementing a `pow` hint, I found that our VM can't execute some of the code outside of the hint itself.

Given this code:

```
%builtins output range_check

from starkware.cairo.common.serialize import serialize_word
from starkware.cairo.common.pow import pow

func main{output_ptr: felt*, range_check_ptr}() {
    alloc_locals;
    let (local v) = pow(2, 10);
    serialize_word(v);

    return ();
}
```
We'll get this error from this VM:

> runtime error: pc 0:51 step 40: running instruction: compute res: cannot multiply memory addresses

The cairo-run executes it just fine.

Let's disassemble this function and see what's going on at PC=51. The entire pow function listing is:

```
// <F4> (pc=19)
[ 19]     jmp rel 6 if [fp-3] != 0; // targets L0
[ 21]     [ap] = [fp-5], ap++;
[ 22]     [ap] = 1, ap++;
[ 24]     ret;
// <L0> (pc=25)
[ 25]     [fp] = 1, ap++;
[ 27]     [fp+1] = [fp-4], ap++;
[ 28]     [fp+2] = [fp-3], ap++;
// <L2> (pc=29)
[ 29]     [ap+3] = [ap-2] * [ap-2], ap++;
[ 30]     jmp rel 7 if [ap-1] != 0, ap++; // targets L1
[ 32]     [ap+2] = [ap-3] * 1809251394333065606848661391547535052811553607665798349986546028067936010241, ap++; // div 2
[ 34]     [ap-1] = [ap-6], ap++;
[ 35]     jmp rel -6, ap++; // targets L2
// <L1> (pc=37)
[ 37]     [ap-1] = [ap-3] + -1;
[ 39]     [ap+2] = [ap-1] * 1809251394333065606848661391547535052811553607665798349986546028067936010241, ap++; // div 2
[ 41]     [ap-1] = [ap-6] * [ap-5], ap++;
[ 42]     jmp rel -13 if [ap] != 0, ap++; // targets L2
[ 44]     call rel -43; // calls F2; ap += 2
[ 46]     call rel -46; // calls F0; ap += 2
[ 48]     [ap] = [ap-2] + -2, ap++;
[ 50]     [ap-4] = [ap] + [ap-1], ap++;
[...]

PR #310, opened 2024-03-14
pkg/vm/memory: fix sub operation result type for addr-addr

quasilyte:
Refs #284

### DIFF
--- a/pkg/vm/memory/memory_value.go
+++ b/pkg/vm/memory/memory_value.go
@@ -259,7 +259,20 @@ func (mv *MemoryValue) Add(lhs, rhs *MemoryValue) error {
 // Subs two memory values if they're in the same segment or the rhs is a Felt.
 func (mv *MemoryValue) Sub(lhs, rhs *MemoryValue) error {
 	if lhs.IsAddress() {
-		return mv.addrUnsafe().Sub(lhs.addrUnsafe(), rhs.Any())
+		asAddr := mv.addrUnsafe()
+		if rhs.IsAddress() {
+			// The result is the offset value, felt-typed.
+			// See #284
+			mv.isFelt = true
+			mv.isAddress = false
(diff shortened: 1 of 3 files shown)


pkg/vm/memory/memory_value.go
```diff
@@ -259,7 +259,20 @@ func (mv *MemoryValue) Add(lhs, rhs *MemoryValue) error {
...
+			mv.isFelt = true
+			mv.isAddress = false
+			if err := asAddr.Sub(lhs.addrUnsafe(), rhs.Any()); err != nil {
+				return err
+			}
+			mv.felt.SetUint64(asAddr.Offset)
+			return nil
+		}
+		// This code is executed for felt-typed rhs.
+		return asAddr.Sub(lhs.addrUnsafe(), rhs.Any())
```
cicr99, line 275: Shouldn't we then remove the `case` for rhs being a MemoryAddress in this function? or is still possible that this method is called from another place and that part gets executed anyway? Is it correct in that case to return an address instead of a felt?

quasilyte: > Is it correct in that case to return an address instead of a felt?

It's hard for me to tell what are the original intents of the `MemoryAddress.Sub`.
I think it was intended to be a general purpose implementation of all supported overloadings for the addresses, but since it's implemented inside `MemoryAddress` and it does everything in-place, there is no way to make the operation produce `felt`.

If we would return the value instead, it could solve the problems, so I agree on that part. It looks like a better long-term solution. Let me try that right now.

quasilyte:
If we move the subtraction logic from MemoryAddress to MemoryValue, there is no typing problem as MemoryValue can hold both felt-typed and addr-typed values. Implementing a subtraction as a MemoryAddress method is cumbersome: it can't store the result directly and returning as a result would mean either using `any` or `MemoryValue`.

cicr99 (approved):
Looks a lot better

Issue #241, opened 2024-03-11
UnsignedDivRem


cicr99:
Performs an unsigned division and modulus operation on a given dividend and divisor.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/math.cairo#L301
> Note: `divmod` function used inside the hint is already used in other hints, so check the code to see how it is solved

har777:
@cicr99 I would like to work on this!

cicr99:
@har777 I've assigned you to the issue!

PR #311, opened 2024-03-16
Implement UnsignedDivRem hint

har777:
Implements #241


pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -501,3 +501,78 @@ func createSplitIntHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+			if err != nil {
+				return err
+			}
+			rAddr, err := r.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+
+			// (PRIME // range_check_builtin.bound)
+			divUpperBound := new(big.Int).Div(fp.Modulus(), utils.Uint256Max128.ToBig())
```
har777, line 534: I think this should be just calculated once and stored as a constant somewhere. 

Something like: 
```
//> (PRIME // range_check_builtin.bound)
var divUpperBound big.Int
divUpperBound.SetString("10633823966279327296825105735305134080", 10)
```

But I'm not sure what it should be called.

har777:
Should I write an integration test for this?

pkg/hintrunner/zero/zerohint_math.go
cicr99: Do you know if it's used somewhere else in the code? If it's not, I would say to leave it just like this for now.

cicr99: Also I have a question here. Where did you get the value from  range_check_builtin.bound?
I've seen it being used other times and its value is 1<<128, and you can use `utils.FeltMax128` for it, but is not the same for `utils.Uint256Max128`, this one represents `(1<<128) -1`  if I'm not mistaken

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -501,3 +501,78 @@ func createSplitIntHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+			qValue := memory.MemoryValueFromFieldElement(&q)
+			if err := vm.Memory.WriteToAddress(&qAddr, &qValue); err != nil {
+				return err
+			}
+			rValue := memory.MemoryValueFromFieldElement(&r)
+			if err := vm.Memory.WriteToAddress(&rAddr, &rValue); err != nil {
+				return err
+			}
+
+			return nil
```
cicr99, line 554: you can just finish with `return vm.Memory.WriteToAddress(&rAddr, &rValue)`

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -555,5 +555,106 @@ func TestZeroHintMath(t *testing.T) {
 				errCheck: errorTextContains("outside of the range [0, 2**250)"),
 			},
 		},
+		"UnsignedDivRem": {
+			{
+				operanders: []*hintOperander{
+					{Name: "value", Kind: fpRelative, Value: feltUint64(100)},
+					{Name: "div", Kind: fpRelative, Value: feltUint64(6)},
+					{Name: "q", Kind: uninitialized},
+					{Name: "r", Kind: uninitialized},
```
cicr99, line 564: See in the code that `q` and `r` are references to the `range_check_ptr` so you'll need to handle that differently here. Check `Assert250bits` test implementation as an example. You'll need to fix this on the remaining test cases as well

cicr99 (changes requested):

pkg/hintrunner/zero/zerohint_math.go
har777: Yep you're right it should be 2**128. I'll be more careful.

har777: I just made it:

```
// (PRIME // range_check_builtin.bound)
// 800000000000011000000000000000000000000000000000000000000000001 // 2**128
var divUpperBound big.Int
divUpperBound.SetString("8000000000000110000000000000000", 16)
```

cicr99 (approved):
lgtm

PR #314, opened 2024-03-19
Cairo0 hint uint256 mul div mod

MaksymMalicki:


pkg/hintrunner/zero/zerohint_uint256_test.go
cicr99, line 1: I would also like to add a case where `quotient_high` is different from 0, even try to get `quotient_high.high` different from 0

cicr99 (approved):

PR #316, opened 2024-03-19
Add: Cairo0 Sqrt hint

Tomi-3-0:
### feat: Implement Cairo0 "Sqrt" hint 

This commit introduces a cairo0 hint named "Sqrt" to calculate the square root of a given value while ensuring it falls within a specified range ([0, 2**250)). 

The calculation utilizes the [Sqrt()](https://github.com/holiman/uint256/blob/v1.2.3/uint256.go#L1307) function from the `"github.com/holiman/uint256"` library, setting z (root) to ⌊√x⌋ (value), the largest integer such that z² ≤ x, and returns z. 

To ensure the value remains within the specified range, the `FeltLt()` function from `"github.com/NethermindEth/cairo-vm-go/pkg/utils"` is utilized.

Accompanying tests cover various scenarios:
- Testing the square root of 25, which is expected to be 5.
- Handling zero input value, which is expected to be 0
- Verifying the square root of 50, expected to be 7.xxx.... (rounded to 7).
- Validating the square root of -127, which is expected to fail as it falls outside of the range [0, 2**250)

Relevant link:
[https://github.com/holiman/uint256/blob/v1.2.3/uint256.go#L1307](https://github.com/holiman/uint256/blob/v1.2.3/uint256.go#L1307)


cicr99 (approved):
lgtm!

PR #317, opened 2024-03-19
Replace old felt127 generation with constant

MaksymMalicki:


pkg/hintrunner/zero/zerohint_uint256_test.go
```diff
@@ -1,25 +1,24 @@
...
 	"testing"
 
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
 func TestZeroHintUint256(t *testing.T) {
 	// Values used in the test cases
 	// 1 << 127
```
cicr99, line 13: You can remove these two lines of comments as well, as there are no variables being declared here anyway

cicr99 (approved):

Issue #242, opened 2024-03-11
SignedDivRem


cicr99:
Performs a signed division and modulus operation on a given value, ensuring the quotient is within a specified range.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/math.cairo#L329

> Note: This hint uses `divmod` function and `range_check_builtin`, both have been already implemented in this hint section so just check the code for clarification

Tomi-3-0:
Hello @cicr99 , I would like to work on this if it isn't assigned yet

PR #318, opened 2024-03-20
Implement SignedDivRemHint

MaksymMalicki:


har777:
Implements #242

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -10,6 +11,8 @@ import (
 )
 
 func TestZeroHintMath(t *testing.T) {
+	felt127 := new(fp.Element).SetBigInt(new(big.Int).Lsh(big.NewInt(1), 127))
```
har777, line 14: Why not use `&utils.Felt127`?

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -586,7 +589,63 @@ func TestZeroHintMath(t *testing.T) {
...
+					{Name: "value", Kind: apRelative, Value: feltString("0")},
+					{Name: "div", Kind: apRelative, Value: &utils.FeltMax128},
+					{Name: "bound", Kind: apRelative, Value: felt127},
+					{Name: "r", Kind: reference, Value: addrBuiltin(starknet.RangeCheck, 0)},
+					{Name: "biased_q", Kind: reference, Value: addrBuiltin(starknet.RangeCheck, 1)},
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newSignedDivRemHint(ctx.operanders["value"], ctx.operanders["div"], ctx.operanders["bound"], ctx.operanders["r"], ctx.operanders["biased_q"])
+				},
+				errCheck: errorTextContains("div="),
```
har777, line 604: Probably better to check the exact error string 🤔 
Same for other `errCheck` below.

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -590,6 +590,106 @@ func createSplitFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error
...
+			divFelt, err := hinter.ResolveAsFelt(vm, div)
+			if err != nil {
+				return err
+			}
+			//> assert 0 < ids.div <= PRIME // range_check_builtin.bound, f'div={hex(ids.div)} is out of the valid range.'
+			upperBoundBig := new(big.Int).Div(fp.Modulus(), new(big.Int).Lsh(big.NewInt(1), 128))
+			upperBound := new(fp.Element).SetBigInt(upperBoundBig)
+			if !utils.FeltLe(divFelt, upperBound) {
+				return fmt.Errorf("div=%v is out of the valid range", divFelt)
+			}
```
har777, line 617: Check for `0 > ids.div` too?

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -590,6 +590,106 @@ func createSplitFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error
...
+			// assert -ids.bound <= q < ids.bound, f'{int_value} / {ids.div} = {q} is out of the range [{-ids.bound}, {ids.bound}).'
+			// ids.biased_q = q + ids.bound
+
+			//> assert_integer(ids.div)
+			divFelt, err := hinter.ResolveAsFelt(vm, div)
+			if err != nil {
+				return err
+			}
+			//> assert 0 < ids.div <= PRIME // range_check_builtin.bound, f'div={hex(ids.div)} is out of the valid range.'
+			upperBoundBig := new(big.Int).Div(fp.Modulus(), new(big.Int).Lsh(big.NewInt(1), 128))
```
har777, line 613: This works but from a perf perspective its probably just better to say:
```
// (PRIME // range_check_builtin.bound)
// 800000000000011000000000000000000000000000000000000000000000001 // 2**128
var upperBoundBig big.Int
upperBoundBig.SetString("8000000000000110000000000000000", 16)
```

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -590,6 +590,106 @@ func createSplitFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error
...
+				return fmt.Errorf("div=%v is out of the valid range", divFelt)
+			}
+
+			//> assert_integer(ids.bound)
+			boundFelt, err := hinter.ResolveAsFelt(vm, bound)
+			if err != nil {
+				return err
+			}
+			//> assert ids.bound <= range_check_builtin.bound // 2, f'bound={hex(ids.bound)} is out of the valid range.'
+			felt127 := new(fp.Element).SetBigInt(new(big.Int).Lsh(big.NewInt(1), 127))
```
har777, line 625: There is a `Felt127` constant in the `constant.go` file.

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -590,6 +590,106 @@ func createSplitFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error
...
+			var valueBig, divBig, boundBig big.Int
+			valueFelt.BigInt(&valueBig)
+			divFelt.BigInt(&divBig)
+			boundFelt.BigInt(&boundBig)
+			qBig, rBig := new(big.Int).DivMod(&valueBig, &divBig, new(big.Int))
+			qFelt, rFelt := new(fp.Element).SetBigInt(qBig), new(fp.Element).SetBigInt(rBig)
+			//> assert -ids.bound <= q < ids.bound, f'{int_value} / {ids.div} = {q} is out of the range [{-ids.bound}, {ids.bound}).'
+			if new(big.Int).Abs(&boundBig).Cmp(new(big.Int).Abs(qBig)) == -1 {
+				return fmt.Errorf("%v / %v = %v is out of the range [-%v, %v]", valueFelt, divFelt, qBig, boundFelt, boundFelt)
+			}
```
har777, line 645: Isn't this only raising an error if `ids.bound < q`? Shouldn't there be an error for `q == ids.bound` too? 🤔

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: done

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: done, you are right haha

pkg/hintrunner/zero/zerohint_math_test.go
MaksymMalicki: I forgot to replace it after merging main haha

pkg/hintrunner/zero/zerohint_math_test.go
MaksymMalicki: done

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: Thats good point. Since we are using that value in multiple places, maybe we should move the PRIME // range_check_builtin.bound to the consts?

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: You are right, it should compare ids.bound <= q

pkg/hintrunner/zero/zerohint_math.go
har777: Yeah thats probably a good idea. Not sure what to call it though 😆

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -590,6 +590,106 @@ func createSplitFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error
...
+			//> assert_integer(ids.bound)
+			boundFelt, err := hinter.ResolveAsFelt(vm, bound)
+			if err != nil {
+				return err
+			}
+			//> assert ids.bound <= range_check_builtin.bound // 2, f'bound={hex(ids.bound)} is out of the valid range.'
+			if !utils.FeltLe(boundFelt, &utils.Felt127) {
+				return fmt.Errorf("bound=%v is out of the valid range.", boundFelt)
+			}
+			//> int_value = as_int(ids.value, PRIME)
```
har777, line 629: I believe `as_int` is something we need to implement and add. Lets confirm with @cicr99

pkg/hintrunner/zero/zerohint_math.go
cicr99: Yes please, move this to the `constants` file, it's being used already in a few places in the code.
You can call it `primeHigh` or something similar, as it is the high part of the field prime.
When you move it, make sure to add a comment in the code saying what this number is, I mean:
```
// (PRIME // range_check_builtin.bound)
// 800000000000011000000000000000000000000000000000000000000000001 // 2**128
```

pkg/hintrunner/zero/zerohint_math.go
cicr99: I don't think using just one comparison will solve this as the range limits are inclusive and exclusive respectively. So you'll have to check for the equality in just one case.
I see two choices:
- You can do the checks separately as in `-ids.bound <= q && q < ids.bound`
- You can continue to use `Abs()` method, but you'll need to add another condition as in `Abs(q) < Abs(ids.bound) || q == -ids.bound`

MaksymMalicki: I dont think the first equation will work in the field arithmetics, as -ids.bound (PRIME - ids.bound) is greater than ids.bound, when ids.bound is greater then PRIME // 2. The range would have to look more like q < ids.bound || q >= -ids.bound

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: moved to separate PrimeHigh constant

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: Added the proper check, handling that using big ints, not felts

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: Implemented that in the zerohint_utils.go. Not really sure if that is correct place for that functionality

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -590,6 +590,106 @@ func createSplitFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error
...
+			if err != nil {
+				return err
+			}
+
+			var divBig, boundBig big.Int
+			divFelt.BigInt(&divBig)
+			boundFelt.BigInt(&boundBig)
+
+			// int_value = as_int(ids.value, PRIME)
+			intValueBig := AsInt(valueFelt)
```
cicr99, line 637: I would move these lines up, right after the declaration of valueFelt, so as not to repeat the same comment twice. I also thinks it looks more organized that way

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -590,6 +590,106 @@ func createSplitFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error
...
+
+			rAddr, err := r.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+			rValue := memory.MemoryValueFromFieldElement(rFelt)
+			err = vm.Memory.WriteToAddress(&rAddr, &rValue)
+			if err != nil {
+				return err
+			}
```
cicr99, line 654: Same as before, I'd move these lines up, right after the declaration of `rFelt`, as writing the value to memory is still something you have to do from that step

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -590,6 +590,106 @@ func createSplitFeltHinter(resolver hintReferenceResolver) (hinter.Hinter, error
...
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+			//> from starkware.cairo.common.math_utils import as_int, assert_integer
+			// assert_integer(ids.div)
+			// assert 0 < ids.div <= PRIME // range_check_builtin.bound, f'div={hex(ids.div)} is out of the valid range.'
+			// assert_integer(ids.bound)
+			// assert ids.bound <= range_check_builtin.bound // 2, f'bound={hex(ids.bound)} is out of the valid range.'
+			// int_value = as_int(ids.value, PRIME)
+			// q, ids.r = divmod(int_value, ids.div)
+			// assert -ids.bound <= q < ids.bound, f'{int_value} / {ids.div} = {q} is out of the range [{-ids.bound}, {ids.bound}).'
+			// ids.biased_q = q + ids.bound
```
cicr99, line 605: Missing `>` at the beginning of these commented code lines

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -586,7 +589,105 @@ func TestZeroHintMath(t *testing.T) {
 				}),
 			},
 		},
-
+		"SignedDivRem": {
+			{
+				operanders: []*hintOperander{
+					{Name: "value", Kind: apRelative, Value: feltString("0")},
```
cicr99, line 595: I would rather use `feltString()` for bigger values, those you can't represent as uint64 for example. Passing the value as a string surely makes the process of getting the felt less efficient. You can use `feltUint64` or `feltInt64` instead or direct values from the `constant` file. In this case you could directly use `FeltZero`

Check the rest of the felts you passed in these tests because this comment applies to most of them

pkg/hintrunner/zero/zerohint_utils.go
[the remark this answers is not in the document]
MaksymMalicki: renamed to utils

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: you're right

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: done

pkg/hintrunner/zero/zerohint_math.go
MaksymMalicki: added

pkg/hintrunner/zero/zerohint_math_test.go
MaksymMalicki: Applied that to the tests

pkg/hintrunner/zero/utils.go
```diff
@@ -0,0 +1,20 @@
...
+	valueFelt.BigInt(&valueBig)
+	boundBig := new(big.Int).Div(fp.Modulus(), big.NewInt(2))
+
+	// val if val < prime // 2 else val - prime
+	if valueBig.Cmp(boundBig) == -1 {
+		return valueBig
+	}
+	negativeValueBig := new(big.Int).Sub(&valueBig, fp.Modulus())
+	return *negativeValueBig
+}
```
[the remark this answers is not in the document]
har777: I also like the idea of passing in a `result *big.Int` argument. So kind of like how the normal `big.Int` api's work. And use the passed `result` to calculate and return the final value. Any thoughts?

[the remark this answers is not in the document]
MaksymMalicki: I think, that this method directly transforms a value from the field arithmetics to the integer. If the value is negative, lets say -1, it gets transformed into PRIME - 1. That is why we subtract the PRIME from the number, to get it's actual value. Which would also mean, that all values >=PRIME // 2 are in fact negative in values, represented by felts, which I'm not sure is correct.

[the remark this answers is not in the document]
cicr99: I agree with having  `*fp.Element` as argument instead of `*big.Int `, as the idea would be to make the conversion in this method. 

Also, I haven't seen this method used with a different prime, so for now I think we can leave it as it is. Let's leave a comment at the top regarding this.

I would rather for this method to return a pointer to bigInt instead, most of the operations at the end are done using it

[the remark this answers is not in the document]
har777: I feel having a `*fp.Element` is less flexible. If the first thing the function does is converting the argument to `big.int` then might as well make the parent pass it in right? That way the parent still has a reusable element and big.int it can use downstream 🤔

Issue #243, opened 2024-03-11
IsQuadResidue


cicr99:
Determines if a given number is a quadratic residue and calculates its square root accordingly.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/math.cairo#L455

The main problem here is to implement `is_quad_residue` function from the `starkware.python` package. Maybe `Legendre()` function from the `Element` library used in the project will help with that

Tomi-3-0:
Hi @cicr99 , I will like to work on this

Tomi-3-0:
Hello @cicr99, I opened a [PR](https://github.com/NethermindEth/cairo-vm-go/pull/322) in regards to this issue.
Please review. Thank you!

cicr99:
Addressed in Pr #322

PR #322, opened 2024-03-24
Add: Cairo0 IsQuadResidue Hint

Tomi-3-0:
## Implement Cairo0 isQuadResidue hint #243 

This hint checks if an input value `x` is a quadratic residue and calculates its square root. If x is not a quadratic residue, it computes the square root of the result of dividing `x` by 3 i.e., `√(x/3)` or `3 * y² = x;`

The hint utilizes [Legendre()](https://github.com/Consensys/gnark-crypto/blob/da0317fd013308db6ce847bc9c3d506a2a3ae0ff/ecc/stark-curve/fp/element.go#L1153) method to check if x is a quadratic residue.

There is a challenge in choosing between the two possible square roots in the Stark field returned by the `ModSqrt()` method, as it tends to return the higher of the root values. This implementation compares the calculated square root `y` with `P\2` to determine whether to use ` y ` or `P - y`. If ` y` is greater than `P\2`, the implementation sets `y` to `P - y`, ensuring the smaller root of the two possible roots is returned.

This ensures that the smaller square root is always returned, thus providing an output consistent with the other VM implementations.

 **Methods Used**:
    - **[ModSqrt](https://cs.opensource.google/go/go/+/go1.22.3:src/math/big/int.go;l=1084)** from the `big/math` library which employs the Tonelli-Shanks algorithm or faster methods when applicable.
    - **Element library**'s **[sqrt](https://github.com/consensys/gnark-crypto/blob/v0.12.1/ecc/stark-curve/fp/element.go#L1172)** method for calculating square roots.


pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -711,4 +711,72 @@ func createUnsignedDivRemHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+				return err
+			}
+
+			if x.Legendre() == 1 { // If x.Legendre() returns 1 i.e x is a quadratic residue (z ≡ x² (mod q))
+				// calculates the square root
+				xU256 := uint256.Int(x.Bits())
+				xU256.Sqrt(&xU256)
+
+				value := fp.Element{}
+				value.SetBytes(xU256.Bytes())
```
cicr99, line 745: Is it possible to use `Sqrt()` function from the `"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"` library? It works directly with the felt values, so you wouldn't need to do the conversion

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -711,4 +711,72 @@ func createUnsignedDivRemHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+				}
+			} else {
+				//calculates the square root of the result of dividing x by 3 or 3 * y_squared = x;
+				xU256 := uint256.Int(x.Bits())
+				divResult := xU256.Div(&xU256, uint256.NewInt(3))
+
+				divResult.Sqrt(divResult)
+
+				value := fp.Element{}
+				value.SetBytes(divResult.Bytes())
```
cicr99, line 759: Same thing for this part. You can use the `Div` function from the `Element` library, which already does the correct modular division and then use the `Sqrt` function from the same library.

I'm not sure if Uint256 methods do these operations using modular arithmetic, but if you check out the python codes for the functions they are using in the hint, you'll see they are. Hence, using the `felt` library in this case is probably the best way to go. Also you don't have to do any conversions either

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -731,5 +731,74 @@ func TestZeroHintMath(t *testing.T) {
 				errCheck: errorTextContains("div=0x8000000000000110000000000000001 is out of the valid range."),
 			},
 		},
+
+		"IsQuadResidue": {
```
cicr99, line 735: It would be good to check using `cairo-run` that the result is actually the one you are expecting. I created an issue to implement a script that does this automatically #325, but in the mean time it would be great if you could test it locally

cicr99 (changes requested):

rodrigo-pino:
@Tomi-3-0 can you explain what is the issue your having here. I don't understand why you cannot fix it?

Converting them to uint256 and finding the sqrt of it good enough? Why is it that using the fp.Element sqrt doesn't work

Tomi-3-0:
> @Tomi-3-0 can you explain what is the issue your having here. I don't understand why you cannot fix it?
> 
> Converting them to uint256 and finding the sqrt of it good enough? Why is it that using the fp.Element sqrt doesn't work
> 

The sqrt function from the uint256 library does not perform modular arithmetic as required by the hint

rodrigo-pino:
What do you mean modular arithmetic? Maybe I am missing something, but what operation do you need to do exactly that cannot be composed by the existing arithmetic operations of the uint256/fp library

Tomi-3-0:
> What do you mean modular arithmetic? Maybe I am missing something, but what operation do you need to do exactly that cannot be composed by the existing arithmetic operations of the uint256/fp library

@rodrigo-pino The `fp` library's `Sqrt` function computes the square root modulo a prime `q`. It calculates the integer value `z` such that `z² ≡ x (mod q)`. 
For example, 55 ≡ 16 (mod 13), but 55 ≇ 16 (mod 11) since 39 is a multiple of 13 but not of 11
let's say `x = 81` and `q = 0x800000000000011000000000000000000000000000000000000000000000001`. If we calculate the square root of `81` modulo `q` using the `fp` library's `Sqrt` function, it will provide the result according to modular arithmetic rules.

On the other hand, the `uint256` library's `sqrt` function computes the square root in a traditional mathematical sense, disregarding modular arithmetic. For example, `sqrt(81)` would return `9`.

The issue here arises when the results obtained for some values differ from those obtained from the same function in other contexts, such as Lambdaclass's VM

rodrigo-pino:
Hey @Tomi-3-0 I understand how the `uint256` and the `felt` arithmetic operations work. What I am asking is what operation you want to do?

What is the arithmetic operation that should be performed to get to the right value?

I don't think it is a blocker, unless you require an operation that re-defines the math operators (`+`, `-`, etc...) , it should be achieved by a finite combination of your classic arithmetic operators. So please tell me what is the operation and maybe the solution is for us to define it.

Tomi-3-0:
> Hey @Tomi-3-0 I understand how the `uint256` and the `felt` arithmetic operations work. What I am asking is what operation you want to do?
> 
> What is the arithmetic operation that should be performed to get to the right value?
> 
> I don't think it is a blocker, unless you require an operation that re-defines the math operators (`+`, `-`, etc...) , it should be achieved by a finite combination of your classic arithmetic operators. So please tell me what is the operation and maybe the solution is for us to define it.

Hi @rodrigo-pino ,Thanks for your time.

The arithmetic operation to be performed is to - find the sqrt of an integer divided by 3 i.e `√(x÷3)` if the integer isn't a quadratic residue
OR
- find the sqrt of an integer `(√x)` if it's a quadratic residue.

I use the `Legendre()` method to check if an integer is a quadratic residue

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -711,4 +711,59 @@ func createUnsignedDivRemHinter(resolver hintReferenceResolver) (hinter.Hinter,
 	}
 
 	return newUnsignedDivRemHinter(value, div, q, r), nil
+
+}
+
+func newIsQuadResidueHint(x, y hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "IsQuadResidue",
```
rodrigo-pino, line 719: Is this the right name for the function? Looking at other VMs, it seems it should be called `recoverY` from which given a known x coordinante of a point in the elliptic curve you find the y corrdinate.

pkg/hintrunner/zero/zerohint_math.go
```diff
@@ -711,4 +711,59 @@ func createUnsignedDivRemHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+			//> from starkware.crypto.signature.signature import FIELD_PRIME
```
rodrigo-pino, line 721: nitpick: this import tab space doesn't match the rest of the code

rodrigo-pino:
> > Hey @Tomi-3-0 I understand how the `uint256` and the `felt` arithmetic operations work. What I am asking is what operation you want to do?
> > What is the arithmetic operation that should be performed to get to the right value?
> > I don't think it is a blocker, unless you require an operation that re-defines the math operators (`+`, `-`, etc...) , it should be achieved by a finite combination of your classic arithmetic operators. So please tell me what is the operation and maybe the solution is for us to define it.
> 
> Hi @rodrigo-pino ,Thanks for your time.
> 
> The arithmetic operation to be performed is to
> 
>     * find the sqrt of an integer divided by 3 i.e `√(x÷3)` if the integer isn't a quadratic residue
>       OR
> 
>     * find the sqrt of an integer `(√x)` if it's a quadratic residue.
> 
> 
> I use the `Legendre()` method to check if an integer is a quadratic residue

@Tomi-3-0 thanks for answering. I still didn't understand the issue so I looked into other VMs Using `felt` sqrt is a must from what I've seen. From the failing test I get that from the possible values you're not getting the one that you're expected. In this case "-10" which is not technically wrong though.

pkg/hintrunner/zero/zerohint_math.go
rodrigo-pino: I am mistaken, after looking more, it is called like this.

Tomi-3-0:
> > > Hey @Tomi-3-0 I understand how the `uint256` and the `felt` arithmetic operations work. What I am asking is what operation you want to do?
> > > What is the arithmetic operation that should be performed to get to the right value?
> > > I don't think it is a blocker, unless you require an operation that re-defines the math operators (`+`, `-`, etc...) , it should be achieved by a finite combination of your classic arithmetic operators. So please tell me what is the operation and maybe the solution is for us to define it.
> > 
> > Hi @rodrigo-pino ,Thanks for your time.
> > 
> > The arithmetic operation to be performed is to
> > 
> >     * find the sqrt of an integer divided by 3 i.e `√(x÷3)` if the integer isn't a quadratic residue
> >       OR
> > 
> >     * find the sqrt of an integer `(√x)` if it's a quadratic residue.
> > 
> > 
> > I use the `Legendre()` method to check if an integer is a quadratic residue
> 
> @Tomi-3-0 thanks for answering. I still didn't understand the issue so I looked into other VMs Using `felt` sqrt is a must from what I've seen. From the failing test I get that from the possible values you're not getting the one that you're expected. In this case "-10" which is not technically wrong though.

Exactly. 
And for some other values too the results are different too.

TAdev0:
Hi @Tomi-3-0 , 
do you have any update on this?

I'd like to understand what the issue is exactly.
You don't get the same result as the python cairo vm for that hint,  because of `sqrt` function in the hint. Currently, your last commit calculates sqrt on uint256, so the result will obviously be different as it doesnt calculate the square root in a finite field. 
square roots in finite fields can produce multiple results. I suppose then that we just need to find a matching implementation in Go ?

Tomi-3-0:
> Hi @Tomi-3-0 , 
> do you have any update on this?
> 
> I'd like to understand what the issue is exactly.
> You don't get the same result as the python cairo vm for that hint,  because of `sqrt` function in the hint. Currently, your last commit calculates sqrt on uint256, so the result will obviously be different as it doesnt calculate the square root in a finite field. 
> square roots in finite fields can produce multiple results. I suppose then that we just need to find an implementation in Go that matches the one used by the python vm ?

Hello @TAdev0 .
Thanks for checking this out. My previous implementation calculates the square root in a finite field, but there were still differences in the results.

I also suppose that finding a matching implementation/or creating our own implementation might be the solution

TAdev0:
@Tomi-3-0  ok it's clear! Do you want to investigate and try with one or 2 other libraries? I can also work on it, as you prefer. If you find a library that matches, then we can probably spot the difference in the implementation

Which cases exactly had mismatch in values? Did you also have test that matched correctly when testing the whole hint?

Tomi-3-0:
> @Tomi-3-0  ok it's clear! Do you want to investigate and try with one or 2 other libraries? I can also work on it, as you prefer. If you find a library that matches, then we can probably spot the difference in the implementation
> 
> Which cases exactly had mismatch in values? Did you also have succcessful tests for some values that matched correctly when testing the whole hint?

@TAdev0  yes most values matched correctly.
I would appreciate some help. Thanks

TAdev0:
@Tomi-3-0  ok great, will look into it tomorrow or the day after :) 
Could you just indicate me (if you remember) some values that didnt work?

Tomi-3-0:
> @Tomi-3-0  ok great, will look into it tomorrow or the day after :) 
> Could you just indicate me (if you remember) some values that didnt work? 

@TAdev0  `151461` is a good test value

Tomi-3-0:
Great job @TAdev0 . Give me a minute to check this out. 
On first glance, seems like you're only computing ` P - result` when the input `x` is a quadratic residue.

Did you compare your results with the result of let's say the python VM?

Because calling the `sqrt()` funtion alone from the felt library on an input e.g `x.sqrt()` gave results that differ. You can test this with input values of `9` and `100` for example.

TAdev0:
i have not extensively test with the python vm for now. will do it in the weekend.

yeah i added P-result computation to get the lowest valid result for the square root.

Tomi-3-0:
> i have not extensively test with the python vm for now. will do it in the weekend.
> 
> yeah i added P-result computation to get the lowest valid result for the square root. 

Cool. I would also be working on this. 
Thank you @TAdev0 💥

TAdev0:
here is a cairo file you can use to test : 

```
%builtins output
from starkware.cairo.common.serialize import serialize_word
from starkware.cairo.common.alloc import alloc
from starkware.cairo.common.bool import FALSE, TRUE


// Returns TRUE if `x` is a quadratic residue modulo the STARK prime. Returns FALSE otherwise.
// Returns TRUE on 0.
@known_ap_change
func is_quad_residue(x: felt) -> felt {
    alloc_locals;
    local y;
    %{
        from starkware.crypto.signature.signature import FIELD_PRIME
        from starkware.python.math_utils import div_mod, is_quad_residue, sqrt

        x = ids.x
        if is_quad_residue(x, FIELD_PRIME):
            ids.y = sqrt(x, FIELD_PRIME)
            print(ids.y)
        else:
            ids.y = sqrt(div_mod(x, 3, FIELD_PRIME), FIELD_PRIME)
            print(ids.y)
    %}
    // Relies on the fact that 3 is not a quadratic residue modulo the prime, so for every field
    // element x, either:
    //   * x is a quadratic residue and there exists y such that y^2 = x.
    //   * x is not a quadratic residue and there exists y such that 3 * y^2 = x.
    tempvar y_squared = y * y;
    if (y_squared == x) {
        ap += 1;
        return TRUE;
    } else {
        assert 3 * y_squared = x;
        return FALSE;
    }
}

func fill_array(array_start: felt*, iter: felt) -> () {
    if (iter == 10) {
        return ();
    }
    assert array_start[iter] = iter;
    return fill_array(array_start, iter + 1);
}

func check_quad_res{output_ptr: felt*}(inputs: felt*, expected: felt*, iter: felt) {
    if (iter == 10) {
        return ();
    }
    serialize_word(inputs[iter]);
    serialize_word(expected[iter]);

    assert is_quad_residue(inputs[iter]) = expected[iter];
    return check_quad_res(inputs, expected, iter + 1);
}

func main{output_ptr: felt*}() {
    alloc_locals;
    let (inputs: felt*) = alloc();
    fill_array(inputs, 0);

    let (expected: felt*) = alloc();
    assert expected[0] = 1;
    assert expected[1] = 1;
    assert expected[2] = 1;
    assert expected[3] = 0;
    assert expected[4] = 1;
    assert expected[5] = 1;
    assert expected[6] = 0;
    assert expected[7] = 1;
    assert expected[8] = 1;
    assert expected[9] = 1;

    check_quad_res(inputs, expected, 0);

    return ();
}
```

I added `print(ids.y)` to print the result of the sqrt. Indeed, with the python VM, sqrt(9) returns 3 , not `P-3` . So my modification works correctly in this situation. We should check all situations / edge cases to make sure its ok like this

TAdev0:
just check with the value 2 

```
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(2)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltString("1120755473020101814179135767224264702961552391386192943129361948990833801454")),
			},
```
`1120755473020101814179135767224264702961552391386192943129361948990833801454` is what the python vm returns for 2 

here , we fail, and we have `2497747315646029399518187015870805402661554823945403756843730107145038219027` 

this means here, we dont need to do `P - result`
I think we just need to a check that `result < P - result` , and if not, apply P - result , and then we're good

TAdev0:
final implementation: 

```

func newIsQuadResidueHint(x, y hinter.ResOperander) hinter.Hinter {
	return &GenericZeroHinter{
		Name: "IsQuadResidue",
		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
			//> from starkware.crypto.signature.signature import FIELD_PRIME
			//>	from starkware.python.math_utils import div_mod, is_quad_residue, sqrt
			//>	x = ids.x
			//>	if is_quad_residue(x, FIELD_PRIME):
			//>		ids.y = sqrt(x, FIELD_PRIME)
			//>	else:
			//>		ids.y = sqrt(div_mod(x, 3, FIELD_PRIME), FIELD_PRIME)

			x, err := hinter.ResolveAsFelt(vm, x)
			if err != nil {
				return err
			}

			yAddr, err := y.GetAddress(vm)
			if err != nil {
				return err
			}

			xBigInt := math_utils.AsInt(x)

			var value = memory.MemoryValue{}

			if x.IsZero() || x.IsOne() {
				value = memory.MemoryValueFromFieldElement(x)

			} else {
				var result *fp.Element = new(fp.Element)

				if x.Legendre() == 1 {
					// result = x.Sqrt(x)

					var primeBigInt = new(big.Int)
					primeBigInt, ok := primeBigInt.SetString("3618502788666131213697322783095070105623107215331596699973092056135872020481", 10)
					if !ok {
						panic("")
					}

					halfPrimeBigInt := new(big.Int).Rsh(primeBigInt, 1)

					var tempResult = new(big.Int)
					tempResult.ModSqrt(&xBigInt, primeBigInt)

					if tempResult.Cmp(halfPrimeBigInt) > 0 {
						tempResult.Sub(primeBigInt, tempResult)
					}

					result.SetBigInt(tempResult)

				} else {
					result = x.Sqrt(new(fp.Element).Div(x, new(fp.Element).SetUint64(3)))
				}

				value = memory.MemoryValueFromFieldElement(result)
			}

			return vm.Memory.WriteToAddress(&yAddr, &value)
		},
	}
}
```

and the tests, with value from the python vm : 

```
		{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(0)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltInt64(0)),
			},
			// Test case: x is 1
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(1)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltInt64(1)),
			},
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(2)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltString("1120755473020101814179135767224264702961552391386192943129361948990833801454")),
			},
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(3)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltString("1")),
			},
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(4)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltString("2")),
			},
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(5)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltString("703125680814918687568381966029303999302590701505467922907419583337579557417")),
			},
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(6)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltString("1120755473020101814179135767224264702961552391386192943129361948990833801454")),
			},
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(7)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltString("209880913220488156313834499058596184255382000503803956865172539718143124843")),
			},
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(8)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltString("1376991842625927585339051248646540699700002432559210813714368158154204417573")),
			},
			{
				operanders: []*hintOperander{
					{Name: "y", Kind: uninitialized},
					{Name: "x", Kind: fpRelative, Value: feltInt64(9)},
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
				},
				check: varValueEquals("y", feltInt64(3)),
			},
```

everything passes :)

Tomi-3-0:
Great job! 
I will test on a few other corner values and give a feedback asap

Tomi-3-0:
> final implementation:
> 
> ```
> 
> func newIsQuadResidueHint(x, y hinter.ResOperander) hinter.Hinter {
> 	return &GenericZeroHinter{
> 		Name: "IsQuadResidue",
> 		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
> 			//> from starkware.crypto.signature.signature import FIELD_PRIME
> 			//>	from starkware.python.math_utils import div_mod, is_quad_residue, sqrt
> 			//>	x = ids.x
> 			//>	if is_quad_residue(x, FIELD_PRIME):
> 			//>		ids.y = sqrt(x, FIELD_PRIME)
> 			//>	else:
> 			//>		ids.y = sqrt(div_mod(x, 3, FIELD_PRIME), FIELD_PRIME)
> 
> 			x, err := hinter.ResolveAsFelt(vm, x)
> 			if err != nil {
> 				return err
> 			}
> 
> 			yAddr, err := y.GetAddress(vm)
> 			if err != nil {
> 				return err
> 			}
> 
> 			xBigInt := math_utils.AsInt(x)
> 
> 			var value = memory.MemoryValue{}
> 
> 			if x.IsZero() || x.IsOne() {
> 				value = memory.MemoryValueFromFieldElement(x)
> 
> 			} else {
> 				var result *fp.Element = new(fp.Element)
> 
> 				if x.Legendre() == 1 {
> 					// result = x.Sqrt(x)
> 
> 					var primeBigInt = new(big.Int)
> 					primeBigInt, ok := primeBigInt.SetString("3618502788666131213697322783095070105623107215331596699973092056135872020481", 10)
> 					if !ok {
> 						panic("")
> 					}
> 
> 					halfPrimeBigInt := new(big.Int).Rsh(primeBigInt, 1)
> 
> 					var tempResult = new(big.Int)
> 					tempResult.ModSqrt(&xBigInt, primeBigInt)
> 
> 					if tempResult.Cmp(halfPrimeBigInt) > 0 {
> 						tempResult.Sub(primeBigInt, tempResult)
> 					}
> 
> 					result.SetBigInt(tempResult)
> 
> 				} else {
> 					result = x.Sqrt(new(fp.Element).Div(x, new(fp.Element).SetUint64(3)))
> 				}
> 
> 				value = memory.MemoryValueFromFieldElement(result)
> 			}
> 
> 			return vm.Memory.WriteToAddress(&yAddr, &value)
> 		},
> 	}
> }
> ```
> 
> and the tests, with value from the python vm :
> 
> ```
> 		{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(0)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltInt64(0)),
> 			},
> 			// Test case: x is 1
> 			{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(1)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltInt64(1)),
> 			},
> 			{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(2)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltString("1120755473020101814179135767224264702961552391386192943129361948990833801454")),
> 			},
> 			{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(3)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltString("1")),
> 			},
> 			{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(4)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltString("2")),
> 			},
> 			{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(5)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltString("703125680814918687568381966029303999302590701505467922907419583337579557417")),
> 			},
> 			{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(6)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltString("1120755473020101814179135767224264702961552391386192943129361948990833801454")),
> 			},
> 			{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(7)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltString("209880913220488156313834499058596184255382000503803956865172539718143124843")),
> 			},
> 			{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(8)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltString("1376991842625927585339051248646540699700002432559210813714368158154204417573")),
> 			},
> 			{
> 				operanders: []*hintOperander{
> 					{Name: "y", Kind: uninitialized},
> 					{Name: "x", Kind: fpRelative, Value: feltInt64(9)},
> 				},
> 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
> 					return newIsQuadResidueHint(ctx.operanders["x"], ctx.operanders["y"])
> 				},
> 				check: varValueEquals("y", feltInt64(3)),
> 			},
> ```
> 
> everything passes :)
 @TAdev0  passes all tests!

Tomi-3-0:
@TAdev0  passes all tests! 
 Made a few design changes, but the same logic. I think this is ready for review.
 
 THANKS!

TAdev0 (approved):
lgtm!

cicr99 (approved):

cicr99:
Good work! 🎉

Issue #312, opened 2024-03-18
Blake2sAddUint256


cicr99:
Serializes a uint256 number in a blake2s compatible way (little-endian).

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/cairo_blake2s/blake2s.cairo#L480

PR #324, opened 2024-03-25
Implement Blake2sAddUint256 hint

har777:
Implements #312

### DIFF
--- a/pkg/hintrunner/zero/zerohint_blake.go
+++ b/pkg/hintrunner/zero/zerohint_blake.go
@@ -83,3 +83,77 @@ func createBlake2sAddUint256BigendHinter(resolver hintReferenceResolver) (hinter
 
 	return newBlake2sAddUint256BigendHint(low, high, data), nil
 }
+
+func newBlake2sAddUint256Hint(low, high, data hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "Blake2sAddUint256",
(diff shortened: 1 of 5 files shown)


har777:
I think the right way to write tests for this is to implement a new hintOperander Kind which just has a address value which points to a empty-segment/last-free-offset?

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -83,3 +83,77 @@ func createBlake2sAddUint256BigendHinter(resolver hintReferenceResolver) (hinter
 
 	return newBlake2sAddUint256BigendHint(low, high, data), nil
 }
+
+func newBlake2sAddUint256Hint(low, high, data hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "Blake2sAddUint256",
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
```
cicr99, line 90: This method is practically the same as `Blake2sAddUint256Bigend`, I think the main difference is in the shift you have to use, is that correct? I would suggest then to unify them in a single method, you can use a flag to know if you want the result in big endian or little endian and calculate the correct `shift`. Something similar happened with `Uint256Add` and `Uint256AddLow` hints, you can check that code to see how it was handled

har777: Yep different shift and different positions in memory. I've updated the code.

cicr99 (approved):

PR #330, opened 2024-03-28
Implement Ecnegate hint

har777:
Implements #254


pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -0,0 +1,97 @@
...
+			if !ok {
+				return fmt.Errorf("GetEcBaseBig failed")
+			}
+
+			pointValues, err := hinter.GetConsecutiveValues(vm, point, int16(6))
+			if err != nil {
+				return err
+			}
+
+			yD0, err := pointValues[3].FieldElement()
```
MaksymMalicki, line 36: Maybe keep the generation of array in the hint and move the pack function to the separate utility?

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -0,0 +1,97 @@
...
+
+			valueBig := big.NewInt(0)
+			primeBig := fp.Modulus()
+
+			for idx, yD := range []*fp.Element{yD0, yD1, yD2} {
+				var yDBig big.Int
+				yD.BigInt(&yDBig)
+
+				//> as_int(limb, prime)
+				if yDBig.Cmp(new(big.Int).Div(primeBig, big.NewInt(2))) != -1 {
```
MaksymMalicki, line 67: this is implemented in other PRs like SignedDivRem

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -0,0 +1,97 @@
...
+					yDBig.Sub(&yDBig, primeBig)
+				}
+
+				valueToAddBig := new(big.Int).Exp(baseBig, big.NewInt(int64(idx)), nil)
+				valueToAddBig.Mul(valueToAddBig, &yDBig)
+				valueBig.Add(valueBig, valueToAddBig)
+			}
```
MaksymMalicki, line 74: Idk if this is a valid case, but shouldn't the product of pack method be also modulo SECP_P?
`y = pack(ids.point.y, PRIME) % SECP_P`

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -0,0 +1,184 @@
...
+import (
+	"testing"
+
+	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+)
+
+func TestZeroHintEc(t *testing.T) {
+	runHinterTests(t, map[string][]hintTestCase{
+		"EcNegate": {
+			{
```
MaksymMalicki, line 12: I think it would be way more readable for the reviewers, if we start labeling the testcases somehow. Maybe just a simple comment would be enough (maybe enforce via linter) or adding a new field to the test case with the label (but it would be redundant)

pkg/hintrunner/zero/zerohint_ec.go
har777: Nice catch! I missed this.

pkg/hintrunner/zero/zerohint_ec.go
har777: Will merge main and use!

har777: Wasn't in main so I didn't use. I'll check that PR out!

guha-rahul:
Hi there, i am trying to understand the testcases of this pr. Test Case 3 and 4 have the same limbs value but the result is different , also the limbs are supposed to be 84bit number to hold the BASE field. What am i missing? 
Thank you

har777:
> Hi there, i am trying to understand the testcases of this pr. Test Case 3 and 4 have the same limbs value but the result is different , also the limbs are supposed to be 84bit number to hold the BASE field. What am i missing? Thank you

@guha-rahul 

1. test case 3 and 4 have different limbs. 3 uses two d0 values and 4 uses two d1 values.
2. Its the cairo code which uses the hint that enforces any range requirements right. This hint cant be used by any cairo code so technically someone can use this hint with non 84bit numbers too. BigInt3 is just a struct with 3 felt values. I just used a large number in the tests to check if the results match the python hint. @cicr99 incase I'm making a wrong assumption.

pkg/hintrunner/zero/zerohint_ec.go
har777: I've moved some functionality to reusable util methods.

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -0,0 +1,72 @@
...
+					return err
+				}
+				pointValueBig := pointValue.BigInt(new(big.Int))
+				pointValuesBig[i] = pointValueBig
+			}
+
+			primeBig := fp.Modulus()
+
+			//> y = pack(ids.point.y, PRIME) % SECP_P
+			yBig, err := secp_utils.SecPPacked(pointValuesBig[0], pointValuesBig[1], pointValuesBig[2], primeBig)
```
cicr99, line 45: Could we consider passing `pointValuesBig` instead of each value separately? The `pack` method, as far as I've seen, always receives a triple as argument and you have to actually iterate through its members in the implementation

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -0,0 +1,72 @@
...
+			//> value = (-y) % SECP_P
+			yBig.Neg(yBig)
+			yBig.Mod(yBig, secPBig)
+
+			err = ctx.ScopeManager.AssignVariable("value", yBig)
+			if err != nil {
+				return err
+			}
+
+			return nil
```
cicr99, line 60: You can just return the result of `AssignVariable` here

pkg/hintrunner/zero/zerohint_ec_test.go
cicr99: Actually, for running tests we pass as the name of the test `<TestGroup>_<i>` ([here](https://github.com/NethermindEth/cairo-vm-go/blob/39ec9f675e028685085aec862fd459a58e935baa/pkg/hintrunner/zero/zerohint_test.go#L290)) where `TestGroup` would be the name of the hint you're testing and `i` the specific case being tested for that hint. I think this should be good enough for knowing which specific test is failing, but it's true we could get a more readable error output if adding a label instead of having to count which case it is.
I believe adding a comment at the top of each case explaining what you want to achieve with it is very helpful. In fact, it was the approach I was taking.
Adding a label to the tests sounds like a good idea, but having so many test cases to modify, it would be better to address this in a different PR. For now, anyway, could you add some comments?

wdyt?

pkg/hintrunner/zero/zerohint_ec.go
har777: Yeah I actually had the same thought before. I can modify the method to receive a slice of size 3. What do you think of defining a custom type for this though? A type with d0, d1 and d2 fields? Overkill?

pkg/hintrunner/zero/zerohint_ec_test.go
MaksymMalicki: sounds good to me

pkg/hintrunner/zero/zerohint_ec.go
har777: Just used a `[3]*fp.Element` with a comment.

cicr99: Yeah, I thought about a custom type for this as well, but not sure if it will be worth it. Do we have to use this structure very often? Besides the `pack` method, do we need to implement any other functionality for it? If that's not the case, then I would stick to the current solution with the slice

cicr99 (approved):

Issue #265, opened 2024-03-11
NondetBigint3V1


cicr99:
Writes a value to a specified segment.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/secp256r1/bigint.cairo#L19

PR #331, opened 2024-03-28
Implement Nondetbigint3v1 hint

har777:
Implements #265


pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -95,3 +96,82 @@ func createEcNegateHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+			//?     a = []
+			//>     for _ in range(3):
+			//>         num, residue = divmod(num, BASE)
+			//>         a.append(residue)
+			//>     assert num == 0
+			//>     return a
+			//>
+			//> segments.write_arg(ids.res.address_, split(value))
+
+			var residue big.Int
```
MaksymMalicki, line 141: we should move those functionalities to the separate secp utils file. Maybe the generation of the constants as well?

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -180,5 +182,240 @@ func TestZeroHintEc(t *testing.T) {
...
+
+					err := ctx.ScopeManager.AssignVariable("value", value)
+					if err != nil {
+						t.Errorf("Error assigning variable value in scope")
+					}
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newNondetBigint3V1Hint(ctx.operanders["res"])
+				},
+				check: consecutiveVarValueEquals("res", []*fp.Element{feltString("123456"), feltString("0"), feltString("0")}),
```
MaksymMalicki, line 247: maybe use FeltZero from consts here? I see more cases where you have used feltString("0")

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -180,5 +182,240 @@ func TestZeroHintEc(t *testing.T) {
...
+
+					err := ctx.ScopeManager.AssignVariable("value", value)
+					if err != nil {
+						t.Errorf("Error assigning variable value in scope")
+					}
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newNondetBigint3V1Hint(ctx.operanders["res"])
+				},
+				check: consecutiveVarValueEquals("res", []*fp.Element{feltString("1"), feltString("1"), feltString("0")}),
```
MaksymMalicki, line 332: Maybe use const FeltOne here?

pkg/hintrunner/zero/zerohint_ec.go
MaksymMalicki: I mean the secp constants

har777: Yeah lots of functionality is clumped in 1 utils file right now. We need separate files for math utils, secp utils, etc.

har777: I think its cleaner to just write a `split` function and use it here. I'll do that 👍🏽

har777: Done! Please check.

MaksymMalicki (approved):

pkg/hintrunner/zero/zerohint_ec.go
MaksymMalicki: Looks good!

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -181,5 +183,240 @@ func TestZeroHintEc(t *testing.T) {
...
+					value, ok := new(big.Int).SetString("3618502788666127798953978732740734578953660990361066340291730267696802036752", 10)
+					if !ok {
+						t.Errorf("Error creating big.Int")
+					}
+
+					err := ctx.ScopeManager.AssignVariable("value", value)
+					if err != nil {
+						t.Errorf("Error assigning variable value in scope")
+					}
+				},
```
cicr99, line 225: As you've used this logic several times in these tests, I believe you could create a function with it as to not have to repeat it every time

pkg/hintrunner/zero/zerohint_test.go
```diff
@@ -279,10 +279,14 @@ func runHinterTests(t *testing.T, tests map[string][]hintTestCase) {
...
 				}
-				if _, ok := set[string(stringKey)]; ok {
-					t.Fatalf("%s: duplicated test case case (i=%d) found: %s", testGroup, i, stringKey)
+
+				// TODO: temporary hack, figure out a way to make sure tests dont get marked as duplicate even when the scope they start with is different
+				if tc.ctxInit == nil {
+					if _, ok := set[string(stringKey)]; ok {
+						t.Fatalf("%s: duplicated test case case (i=%d) found: %s", testGroup, i, stringKey)
+					}
+					set[string(stringKey)] = struct{}{}
```
cicr99, line 288: You can remove this sanity check altogether, it's not sustainable.

pkg/hintrunner/zero/zerohint_ec_test.go
har777: Changed to the following format:
```
value := bigIntString("154742504910672534362390530", 10)
ctx.ScopeManager.EnterScope(map[string]any{"value": value})
```

cicr99 (approved):

Issue #270, opened 2024-03-11
FastEcAddAssignNewY


cicr99:
Calculates a new y-coordinate for fast elliptic curve addition.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/cairo_secp/ec.cairo#L222

MaksymMalicki:
While checking the code of lambdalabs vm implementation, I see that this hint probably also utilizes execution context:
https://github.com/lambdaclass/cairo-vm_in_go/blob/31c3628bc10ebc1628685b3cdfa72d0e938b533e/pkg/hints/ec_hint.go#L370

PR #334, opened 2024-03-29
Implement FastEcAddAssignNewY hint

har777:
Implements #270


pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -175,3 +175,78 @@ func createNondetBigint3V1Hinter(resolver hintReferenceResolver) (hinter.Hinter,
 
 	return newNondetBigint3V1Hint(res), nil
 }
+
+func newFastEcAddAssignNewYHint() hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "FastEcAddAssignNewY",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			//> value = new_y = (slope * (x0 - new_x) - y0) % SECP_P
```
har777, line 183: Hmmm `SECP_P` should also be in the context right? Given its not imported. I've seen `PRIME` in the code too before but that doesn't need an import?

MaksymMalicki: I think this is the case, where you actually read SECP_P from the context

cicr99: I was going to say I didn't think that we needed to do that here, as you can see the previous hint block that was executed a few lines above, in the same function, and where SECP_P is imported right there. However, the only line between both hint blocks is calling a method that actually executes another hint, and looking at the code, SECP_P should remain the same, but I think the safest choice is to check the scope, specially for cases when is not that simple to determine if the value remains the same. wdyt?

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -418,5 +418,113 @@ func TestZeroHintEc(t *testing.T) {
...
+						t.Errorf("Error assigning variable value in scope")
+					}
+					err = ctx.ScopeManager.AssignVariable("new_x", new_xBig)
+					if err != nil {
+						t.Errorf("Error assigning variable value in scope")
+					}
+					err = ctx.ScopeManager.AssignVariable("y0", y0Big)
+					if err != nil {
+						t.Errorf("Error assigning variable value in scope")
+					}
```
cicr99, line 447: You can pass all these variables at the creation of the new scope, that's the idea of passing a map as an argument in `EnterScope` function. I think this would make the code look a lot smaller and simpler

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -125,3 +125,73 @@ func createNondetBigint3V1Hinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+
+			slope, err := ctx.ScopeManager.GetVariableValue("slope")
+			if err != nil {
+				return err
+			}
+			slopeBig, ok := slope.(*big.Int)
+			if !ok {
+				return fmt.Errorf("value: %s is not a *big.Int", slope)
+			}
```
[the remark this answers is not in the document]
har777: Yeah I think thats a good idea.

pkg/hintrunner/zero/zerohint_ec_test.go
har777: Ah good idea!

pkg/hintrunner/zero/zerohint_ec.go
har777: So I think the right answer is to always check the context. Everything except PRIME should be in the context unless the hint imports it itself. 

This also technically means that every variable that is created in a hint should be added to the scope. We just add the ones we think are needed down the line in the scope rn. I think thats fine. If we face errors because of some hint not adding a variable to scope we can fix it then.

cicr99 (approved):

Issue #267, opened 2024-03-11
FastEcAddAssignNewX


cicr99:
Calculates a new x-coordinate for fast elliptic curve addition.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/cairo_secp/ec.cairo#L210

PR #335, opened 2024-04-01
Implement FastEcAddAssignNewX hint

har777:
Implements #267

### DIFF
--- a/pkg/hintrunner/zero/zerohint_ec.go
+++ b/pkg/hintrunner/zero/zerohint_ec.go
@@ -3,6 +3,7 @@ package zero
 import (
 	"fmt"
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	hintrunnerUtils "github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/utils"
 	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
 	mem "github.com/NethermindEth/cairo-vm-go/pkg/vm/memory"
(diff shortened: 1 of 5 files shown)


pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -200,3 +200,166 @@ func newFastEcAddAssignNewYHint() hinter.Hinter {
...
+						return err
+					}
+					point1XValues[i] = point1XValue
+				} else {
+					point0YValue, err := point0MemoryValues[i].FieldElement()
+					if err != nil {
+						return err
+					}
+					point0YValues[i-3] = point0YValue
+				}
```
cicr99, line 277: For cases like this, it looks cleaner if you just do two separate for loops, one from i = 0 to 3 and the other from i = 3 to 6. You'll also do less operations, currently you always have to perform the if question.

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -200,3 +200,166 @@ func newFastEcAddAssignNewYHint() hinter.Hinter {
...
+				return err
+			}
+
+			err = ctx.ScopeManager.AssignVariable("y0", y0Big)
+			if err != nil {
+				return err
+			}
+
+			err = ctx.ScopeManager.AssignVariable("new_x", new_xBig)
+			if err != nil {
```
cicr99, line 341: Should we add something like `AssignVariables` for the ScopeManager? You would just need to pass a map with all the variables you want to assign and check for error just once after

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -525,5 +525,108 @@ func TestZeroHintEc(t *testing.T) {
...
+					"x0":    bigIntString("3799719333936312867907730225219317480871818784521830610814991", 10),
+					"x1":    bigIntString("1283798249446970358602040710287144628881017552091260500619997", 10),
+					"y0":    bigIntString("5395443952678709065478416501711989224759665054189740766553850", 10),
+					"value": bigIntString("53863685200989733811273896838983614723181733288322685009664997422229669431265", 10),
+					"new_x": bigIntString("53863685200989733811273896838983614723181733288322685009664997422229669431265", 10),
+				}),
+			},
+			{
+				operanders: []*hintOperander{
+					{Name: "slope.d0", Kind: apRelative, Value: feltString("0")},
```
cicr99, line 565: You can use directly `FeltZero` from the constant file in utils for these values. I would recommend using `feltString` only when is a big value, cause for other cases you can use `feltUint64` or `feltInt64`, or direct values form the constant file as in this particular case, which should be more efficient.

pkg/hintrunner/zero/zerohint_ec.go
har777: I think its a good idea. Added.

cicr99 (approved):

PR #337, opened 2024-04-02
Cairo0 import secp256 r1 p hint

MaksymMalicki:


pkg/hintrunner/zero/zerohint_signatures.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func newImportSecp256R1PHinter() hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "Secp256R1",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			//> from starkware.cairo.common.cairo_secp.secp256r1_utils import SECP256R1_P as SECP_P
+			SECP256R1_PBig, ok := utils.GetSecp256R1_P()
+			if !ok {
+				return fmt.Errorf("SECP256R1_P not found")
```
cicr99, line 62: I would change this error message. Maybe that it failed, instead of not found

pkg/utils/secp_utils.go
```diff
@@ -0,0 +1,7 @@
+package utils
+
+import "math/big"
+
+func GetSecp256R1_P() (*big.Int, bool) {
+	return new(big.Int).SetString("115792089210356248762697446949407573530086143415290314195533631308867097853951", 10)
```
har777, line 6: Nit: I would add a comment saying: "2**256 - 2**224 + 2**192 + 2**96 - 1"

pkg/hintrunner/zero/zerohint_signatures_test.go
```diff
@@ -26,5 +26,17 @@ func TestSignatures(t *testing.T) {
 				},
 			},
 		},
+		"ImportSecp256R1P": {
+			{
+				operanders: []*hintOperander{},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{})
```
cicr99, line 33: I would rather use here `InitializeScope` function, as I think up until this function is initialized the ScopeManager is still nil, right?

cicr99 (approved):

PR #339, opened 2024-04-04
Issue/244 usort enter scope

### LINKED ISSUE #244: UsortEnterScope
Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L15

Depends on issue #245

pxyxyrus:
Fix #244


pxyxyrus:
I haven't added any unit tests for this case, and I believe this functionality will need to be checked using an integration test once everything related to usort is implemented.

.gitignore
```diff
@@ -10,3 +10,4 @@ cpu.out
 .DS_Store
 .idea/
 vendor/
+.venv
```
MaksymMalicki, line 13: idk if this is generated by python vm compilation or some other script, but we don't use any python in this repo so try to remove it or compile cairo in some other dir / local repo

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
+)
+
+func newUsortEnterScopeHinter() hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "UsortEnterScope",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			__usort_max_size, err := ctx.ScopeManager.GetVariableValue("__usort_max_size")
```
MaksymMalicki, line 12: i think we should stick to camel case

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -0,0 +1,28 @@
...
+			ctx.ScopeManager.EnterScope(map[string]any{
+				"__usort_max_size": __usort_max_size,
+			})
+
+			return nil
+		},
+	}
+}
+
+func createUsortEnterScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
```
MaksymMalicki, line 26: i think you can remove resolver from args if you dont use it

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -0,0 +1 @@
+package zero
```
MaksymMalicki, line 1: regarding the [comment](https://github.com/NethermindEth/cairo-vm-go/pull/339#issuecomment-2040713274) in the issue, you could enter scope and add the variable from globals in the ctxInit method in the unittest. Later you can check in the check method, if the value in given scope is the same which you have added.

pxyxyrus: Added tests

pkg/hintrunner/zero/zerohint_usort.go
pxyxyrus: removed

.gitignore
pxyxyrus: Sorry, it was my own setting and I didn't mean to commit this, but now it's removed.

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+)
+
+func TestZeroHintUsort(t *testing.T) {
+	runHinterTests(t, map[string][]hintTestCase{
+		"UsortEnterScope": {
+			{
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{
```
cicr99, line 14: I would rather use here `InitializeScopeManager` function instead

.gitignore
```diff
@@ -9,4 +9,4 @@ cpu.out
 .python-version
 .DS_Store
 .idea/
-vendor/
+vendor/
```
cicr99, line 12: No need to commit this change

cicr99 (approved):

PR #341, opened 2024-04-04
Implement Ecdoubleslopev1 hint

### LINKED ISSUE #256: EcDoubleSlopeV1: Computes the slope for doubling a point on an elliptic curve.

har777:
Closes #256


pkg/hintrunner/utils/math_utils.go
```diff
@@ -19,3 +38,67 @@ func AsInt(valueFelt *fp.Element) *big.Int {
 	}
 	return new(big.Int).Sub(&valueBig, fp.Modulus())
 }
+
+func div_mod(n, m, p *big.Int) (*big.Int, error) {
+	// https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/python/math_utils.py#L26
```
har777, line 43: Eh probably a good idea to add unit tests for this 😭

har777: Done!

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -306,3 +306,84 @@ func createFastEcAddAssignNewXHinter(resolver hintReferenceResolver) (hinter.Hin
...
+				}
+				pointXValues[i] = pointValue
+			}
+			for i := 3; i < 6; i++ {
+				pointValue, err := pointMemoryValues[i].FieldElement()
+				if err != nil {
+					return err
+				}
+				pointYValues[i-3] = pointValue
+			}
```
cicr99, line 349: Same [comment](https://github.com/NethermindEth/cairo-vm-go/pull/342#discussion_r1558913689) is valid here

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -505,5 +505,144 @@ func TestZeroHintEc(t *testing.T) {
...
+					// GetSecPBig()
+					{Name: "point.x.d0", Kind: apRelative, Value: feltString("115792089237316195423570985008687907853269984665640564039457584007908834671663")},
+					{Name: "point.x.d1", Kind: apRelative, Value: feltString("115792089237316195423570985008687907853269984665640564039457584007908834671663")},
+					{Name: "point.x.d2", Kind: apRelative, Value: feltString("115792089237316195423570985008687907853269984665640564039457584007908834671663")},
+					{Name: "point.y.d0", Kind: apRelative, Value: feltString("115792089237316195423570985008687907853269984665640564039457584007908834671663")},
+					{Name: "point.y.d1", Kind: apRelative, Value: feltString("115792089237316195423570985008687907853269984665640564039457584007908834671663")},
+					{Name: "point.y.d2", Kind: apRelative, Value: feltString("115792089237316195423570985008687907853269984665640564039457584007908834671663")},
+				},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{})
```
cicr99, line 520: I'd rather use `InitializeScopeManager` functionality here

pkg/hintrunner/zero/zerohint_ec.go
har777: Will be tackled via https://github.com/NethermindEth/cairo-vm-go/issues/346

rodrigo-pino (commented):
Is the use of `big.Int` in this PR justified?

pkg/hintrunner/utils/math_utils.go
```diff
@@ -1,11 +1,33 @@
...
+	n := big.NewInt(3)
+	n.Mul(n, pointX)
+	n.Mul(n, pointX)
+	n.Add(n, alpha)
+
+	m := big.NewInt(2)
+	m.Mul(m, pointY)
+
+	return div_mod(n, m, p)
+}
```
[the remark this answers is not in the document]
har777: If I remember correctly the initial designs for a few util functions like this did have separate parameters. But we made it like this to match the python version.

har777:
> Is the use of `big.Int` in this PR justified?

@rodrigo-pino I'd say so. Each point is a 256 bit number. And the point is being multiplied by itself in the calculations.

pkg/hintrunner/utils/math_utils.go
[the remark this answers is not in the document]
rodrigo-pino: I don't agree. In my opinion doing smth because X project did it that way is not justification enough. I don't see no performance nor readability gains from this approach. Also, I am just passing by, so I won't inssist

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -505,5 +505,144 @@ func TestZeroHintEc(t *testing.T) {
...
+					{Name: "point.y.d1", Kind: apRelative, Value: &utils.FeltZero},
+					{Name: "point.y.d2", Kind: apRelative, Value: &utils.FeltZero},
+				},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					hinter.InitializeScopeManager(ctx, map[string]any{})
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newEcDoubleSlopeV1Hint(ctx.operanders["point.x.d0"])
+				},
+				errCheck: errorTextContains("point[1] % p == 0"),
```
rodrigo-pino, line 547: I think this test is redundant. By testing numbers bigger than felt252 which have modulo automatically applied to them, means you're testing numbers in the felt252 range, which means there is an overlap with these tests. Do you agree?

pkg/hintrunner/utils/math_utils.go
```diff
@@ -1,11 +1,33 @@
 package utils
 
 import (
+	"errors"
 	"math/big"
 
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
+func EcDoubleSlope(point [2]*big.Int, alpha, p *big.Int) (*big.Int, error) {
```
rodrigo-pino, line 10: Return a `big.Int` and not a pointer to it

pkg/hintrunner/utils/math_utils.go
```diff
@@ -23,3 +45,67 @@ func AsIntBig(value *big.Int) *big.Int {
...
+	if c.Cmp(big.NewInt(1)) != 0 {
+		return nil, errors.New("no solution exists (gcd(m, p) != 1)")
+	}
+	res := new(big.Int)
+	res.Mul(n, a)
+	res.Mod(res, p)
+	return res, nil
+}
+
+func igcdex(a, b *big.Int) (*big.Int, *big.Int, *big.Int) {
```
rodrigo-pino, line 62: Return a `big.Int` and not a pointer to it. I've seen this on several occasions on fucntions down the line. Let's start working towards this other pattern PR by PR.

pkg/hintrunner/utils/math_utils.go
```diff
@@ -23,3 +45,67 @@ func AsIntBig(value *big.Int) *big.Int {
...
+	// https://github.com/sympy/sympy/blob/d91b8ad6d36a59a879cc70e5f4b379da5fdd46ce/sympy/core/intfunc.py#L362
+
+	if a.Cmp(big.NewInt(0)) == 0 && b.Cmp(big.NewInt(0)) == 0 {
+		return big.NewInt(0), big.NewInt(1), big.NewInt(0)
+	}
+	g, x, y := gcdext(a, b)
+	return x, y, g
+}
+
+func gcdext(a, b *big.Int) (*big.Int, *big.Int, *big.Int) {
```
rodrigo-pino, line 72: return a value and not a reference

pkg/hintrunner/utils/math_utils.go
```diff
@@ -23,3 +45,67 @@ func AsIntBig(value *big.Int) *big.Int {
...
+		q, c := new(big.Int).DivMod(a, b, new(big.Int))
+		a, b = b, c
+		x, r = r, new(big.Int).Sub(x, new(big.Int).Mul(q, r))
+		y, s = s, new(big.Int).Sub(y, new(big.Int).Mul(q, s))
+	}
+
+	return a, new(big.Int).Mul(x, big.NewInt(int64(xSign))), new(big.Int).Mul(y, big.NewInt(int64(ySign)))
+}
+
+func sign(n *big.Int) (int, *big.Int) {
```
rodrigo-pino, line 104: the second output should be a value and not a reference

pkg/hintrunner/utils/math_utils.go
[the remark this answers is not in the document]
rodrigo-pino: Ok, so after thinking about it, I don't think this approach is correct. I planned on not giving it too much though but in reality this only obfucaste the purpose of the function. It's only two parameters, points X & Y.

You need to take into account that the Cairo VM when created was meant to be very flexible and be able to work with many paremeters. This is not the case currently.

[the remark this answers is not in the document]
har777: Makes sense.

pkg/hintrunner/zero/zerohint_ec_test.go
har777: I don't think its a big deal. It's just another test in the felt252 range. I can remove it if you want.

rodrigo-pino: To be honest, it doesn't affect anything but it does incurrs in extra lines of code making the code more complex. Don't think in the now, but think in a couple of months, someone trying to decipher what does this do. 

So what I think is, it doesn't affect, but it doesn't add anything as well. The best code is no code

pkg/hintrunner/utils/math_utils.go
```diff
@@ -1,25 +1,109 @@
...
+}
+
+func sign(n *big.Int) (int, big.Int) {
+	// https://github.com/sympy/sympy/blob/d91b8ad6d36a59a879cc70e5f4b379da5fdd46ce/sympy/external/ntheory.py#L119
+
+	if n.Sign() < 0 {
+		return -1, *new(big.Int).Abs(n)
 	}
-	return new(big.Int).Sub(value, fp.Modulus())
+	return 1, *new(big.Int).Set(n)
```
rodrigo-pino, line 108: just return *n as the second output param

pkg/hintrunner/utils/secp_utils.go
rodrigo-pino, line 1: This script needs mayor overhaul and will be updated as part of #355. Let's not change it for now

rodrigo-pino (approved):

PR #342, opened 2024-04-05
Implement VerifyZero hint

### LINKED ISSUE #274: VerifyZero
Verifies that a packed value is zero modulo the SECP256R1 prime.

Refrence: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/cairo_secp/field.cairo#L60

har777:
Closes #274


pkg/hintrunner/zero/zerohint_signature.go
```diff
@@ -0,0 +1,87 @@
...
+			// [d0, d1, d2]
+			var valValues [3]*fp.Element
+
+			for i := 0; i < 3; i++ {
+				valValue, err := valMemoryValues[i].FieldElement()
+				if err != nil {
+					return err
+				}
+				valValues[i] = valValue
+			}
```
cicr99, line 48: I've seen this is used in several places in the code for different hints. I think it would be beneficial to encapsulate this logic in a separate method returning the  [3]*fp.Element structure. Maybe something like `ResolveAsBigInt3`

cicr99 (approved):

pkg/hintrunner/zero/zerohint_signature.go
har777: True. I'll add it.

har777: Can I add this in a separate PR? Many hints(both in main and other PRs) will need to be modified.

cicr99: yes, I've created issue #346 for this. let's solve it in a separate PR

Issue #257, opened 2024-03-11
ComputeSlopeV1


cicr99:
Calculates the slope between two points on an elliptic curve.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/cairo_secp/ec.cairo#L91

har777:
I believe this is the right hint: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/cairo_secp/ec.cairo#L91

cicr99:
Sure! Assigned to the issue

Tomi-3-0:
Hi @cicr99 I opened a [PR](https://github.com/NethermindEth/cairo-vm-go/pull/348) regarding this issue. Thanks

TAdev0:
I close this issue as it has been completed with #377

PR #348, opened 2024-04-10
cairo0 ComputeSlopeV1 hint implementation 

Tomi-3-0:
#### ComputeSlopeV1Hint #257 
This hint  calculates the slope between two points on an elliptic curve. Steps involved:
- Difference of x-coordinates:
It calculates the difference between the x-coordinates of the two points (mod_value).

- Zero Check:
It checks if the difference in x-coordinates mod prime is zero. If it's zero an error is raised.

- Differences for Slope Calculation:
It calculates the difference (y2 - y1) between the y-coordinates of the two points (n).
It calculates the difference (x2 - x1) between the x-coordinates of the two points again (m).

- Modular Inverse:
It calculates the modular inverse of m (the difference in x-coordinates) modulo the prime (secPBig).

- Slope Calculation:
It multiplies n (the difference in y-coordinates) with the modular inverse of m (calculated in step 4). The result, multiplied by the prime (secPBig), is the slope of the line connecting the two points.

Test cases for this hint have been compared with the results produced by the  VM implementations in Rust and Go.


pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -505,5 +505,62 @@ func TestZeroHintEc(t *testing.T) {
...
+
+					return newComputeSlopeV1Hint(ctx.operanders["point0.x.d0"], ctx.operanders["point1.x.d0"])
+				},
+				// Printing the value of point0.x.d0
+
+				check: allVarValueInScopeEquals(map[string]any{
+					"slope": bigIntString("41419765295989780131385135514529906223027172305400087935755859001910844026631", 10),
+					"value": bigIntString("41419765295989780131385135514529906223027172305400087935755859001910844026631", 10),
+				}),
+				//fmt.Printf("Name: %s, Age: %d\n", slope, value)
```
rodrigo-pino, line 562: rogue print?

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -505,5 +505,62 @@ func TestZeroHintEc(t *testing.T) {
...
+				},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{})
+				},
+
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+
+					return newComputeSlopeV1Hint(ctx.operanders["point0.x.d0"], ctx.operanders["point1.x.d0"])
+				},
+				// Printing the value of point0.x.d0
```
rodrigo-pino, line 556: Rogue comment?

rodrigo-pino (commented):
It looks really good. I'll give a second revision later today

pkg/hintrunner/zero/zerohint_ec_test.go
Tomi-3-0: Ohhh, yes! 🤦🏾

pkg/hintrunner/zero/zerohint_ec_test.go
Tomi-3-0: Will correct that

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -306,3 +306,139 @@ func createFastEcAddAssignNewXHinter(resolver hintReferenceResolver) (hinter.Hin
...
+			if err != nil {
+				return err
+			}
+
+			secPBig, ok := secp_utils.GetSecPBig()
+			if !ok {
+				return fmt.Errorf("GetSecPBig failed")
+			}
+
+			modValue := new(big.Int).Sub(x0Big, x1Big)
```
cicr99, line 408: This part belongs to the `line_slope` method. Could you say so in a comment and maybe add a link to the definition of this function?

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -306,3 +306,139 @@ func createFastEcAddAssignNewXHinter(resolver hintReferenceResolver) (hinter.Hin
...
+
+			if modValue.Cmp(big.NewInt(0)) == 0 {
+				return fmt.Errorf("%s is multiple of cairo Prime", modValue.Text(10))
+			}
+
+			// Compute the difference of y-coordinates
+			n := new(big.Int).Sub(y1Big, y0Big)
+
+			// Compute the difference of x-coordinates
+			m := new(big.Int).Sub(x1Big, x0Big)
```
cicr99, line 419: I believe you got these two in a different order. The line says:
```
div_mod(point1[1] - point2[1], point1[0] - point2[0], p)
```
so shouldn't it be `y0 -y1, x0 - x1`?

Also, I think you can probably reuse `modValue` from above so you don't have to do the subtraction twice. keep in mind if you do this that modValue doesn't have to be modulated for the second operation so maybe save the modulated value in a different variable

pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -306,3 +306,139 @@ func createFastEcAddAssignNewXHinter(resolver hintReferenceResolver) (hinter.Hin
...
+
+			// directly computes the remainder: z = n * m_inv (mod prime).
+			m_inv := new(big.Int).ModInverse(m, secPBig)
+
+			value := new(big.Int).Mul(n, m_inv)
+			value.Mod(value, secPBig)
```
cicr99, line 425: This is for the div_mod function right? I think we added a utility for this so you probably can use it directly

cicr99:
It looks great!
There are also some conflicts that need to be solved.

Tomi-3-0:
Hello @cicr99 , I linked a new [PR](https://github.com/NethermindEth/cairo-vm-go/pull/377) that closes this issue...had way too many conflicts on this

Tomi-3-0:
Hello @cicr99 , I linked a new [PR](https://github.com/NethermindEth/cairo-vm-go/pull/377) that closes this issue...had way too many conflicts on this

PR #349, opened 2024-04-10
feat: implemented a ScopeManager with a default scope

### LINKED ISSUE #345: Initialize scopes when creating hint context
Currently the unit testing suit creates the context to pass down to the hint execution without initializing any of its members. The idea for this is to only initialize them when necessary. However, most of the hints use the ScopeManager, and there should always be at least the main scope. Hence, let's modify the code so when the context is created, the Scope Manager is also initialized

fishonamos:
I Initialized DefaultContext function in the hinter package. This will initializes the ScopeManager with a default scope when creating a new HintRunnerContext. 

This ensures that the ScopeManager is always initialized with a default scope when necessary and resolve #345

### DIFF
--- a/pkg/hintrunner/zero/hintcode.go
+++ b/pkg/hintrunner/zero/hintcode.go
@@ -56,8 +56,9 @@ const (
 	uint256MulDivModCode      string = "a = (ids.a.high << 128) + ids.a.low/n b = (ids.b.high << 128) + ids.b.low/n div = (ids.div.high << 128) + ids.div.low/n quotient, remainder = divmod(a * b, div)/n ids.quotient_low.low = quotient & ((1 << 128) - 1)/n ids.quotient_low.high = (quotient >> 128) & ((1 << 128) - 1)/n ids.quotient_high.low = (quotient >> 256) & ((1 << 128) - 1)/n ids.quotient_high.high = quotient >> 384/n ids.remainder.low = remainder & ((1 << 128) - 1)/n ids.remainder.high = remainder >> 128"
 
 	// ------ Usort hints related code ------
-	usortEnterScopeCode string = "vm_enter_scope(dict(__usort_max_size = globals().get('__usort_max_size')))"
-	usortVerifyCode     string = "last_pos = 0\npositions = positions_dict[ids.value][::-1]"
+	usortEnterScopeCode               string = "vm_enter_scope(dict(__usort_max_size = globals().get('__usort_max_size')))"
+	usortVerifyMultiplicityAssertCode string = "assert len(positions) == 0"
+	usortVerifyCode                   string = "last_pos = 0\npositions = positions_dict[ids.value][::-1]"
 
 	// ------ Elliptic Curve hints related code ------
 	ecNegateCode            string = "from starkware.cairo.common.cairo_secp.secp_utils import SECP_P, pack\n\ny = pack(ids.point.y, PRIME) % SECP_P\n# The modulo operation in python always returns a nonnegative number.\nvalue = (-y) % SECP_P"
--- a/pkg/hintrunner/zero/zerohint.go
+++ b/pkg/hintrunner/zero/zerohint.go
@@ -144,6 +144,8 @@ func GetHintFromCode(program *zero.ZeroProgram, rawHint zero.Hint, hintPC uint64
 		// Usort hints
 	case usortEnterScopeCode:
 		return createUsortEnterScopeHinter()
+	case usortVerifyMultiplicityAssertCode:
+		return createUsortVerifyMultiplicityAssertHinter()
 	case usortVerifyCode:
 		return createUsortVerifyHinter(resolver)
 		// Other hints
--- a/pkg/hintrunner/zero/zerohint_usort.go
+++ b/pkg/hintrunner/zero/zerohint_usort.go
@@ -9,7 +9,7 @@ import (
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
-func newUsortEnterScopeHint() hinter.Hinter {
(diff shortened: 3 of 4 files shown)


TAdev0 (changes requested):
@fishonamos hi, I think you just need to call `InitializeScopeManager` on `ctx` which is defined in scope.go file . You dont need to create a new `InitializeScopeManager` function.

Moreover, the first line of your function :
```
ctx.ScopeManager = *DefaultNewScopeManager()
```
already does the job of entering a new scope.  Indeed,  `DefaultNewScopeManager` function will execute `NewScopeManager(make(map[string]any)`, thus creating an  uninitialized scope which will be the main scope. The 2nd line will create another scope and it is not necessary.

Simply calling `ctx.InitializeScopeManager(make(map[string]any))` should work.

@cicr99  @rodrigo-pino  @MaksymMalicki  please correct me if I am wrong.

fishonamos:
> @fishonamos hi, I think you just need to call `InitializeScopeManager` on `ctx` which is defined in scope.go file . You dont need to create a new `InitializeScopeManager` function.
> 
> Moreover, the first line of your function :
> 
> ```
> ctx.ScopeManager = *DefaultNewScopeManager()
> ```
> 
> already does the job of entering a new scope. Indeed, `DefaultNewScopeManager` function will execute `NewScopeManager(make(map[string]any)`, thus creating an uninitialized scope which will be the main scope. The 2nd line will create another scope and it is not necessary.
> 
> Simply calling `ctx.InitializeScopeManager(make(map[string]any))` should work.
> 
> @cicr99 @rodrigo-pino @MaksymMalicki please correct me if I am wrong.

That makes sense. Done it

pkg/hintrunner/hinter/hinter.go
```diff
@@ -23,13 +23,19 @@ type HintRunnerContext struct {
 	ConstantSizeSegment mem.MemoryAddress
 }
 
+// InitializeScopeManager initializes the ScopeManager with a default scope.
+func (ctx *HintRunnerContext) InitializeScopeManager(scope map[string]any) {
+	ctx.ScopeManager = *DefaultNewScopeManager(scope)
+}
+
 func InitializeDefaultContext() *HintRunnerContext {
```
TAdev0, line 31: If i'm right with what i said before, no need to add a new function `InitializeScopeManager` here. 

Simply doing : 
```
func InitializeDefaultContext() *HintRunnerContext {
	ctx := &HintRunnerContext{
		DictionaryManager:         DictionaryManager{},
		SquashedDictionaryManager: SquashedDictionaryManager{},
		ConstantSizeSegment:       mem.UnknownAddress,
	}
	InitializeScopeManager(ctx, make(map[string]any)) // Initialize the ScopeManager with a default scope
	return ctx
}
```

should work

pkg/hintrunner/hinter/hinter.go
```diff
@@ -24,12 +24,13 @@ type HintRunnerContext struct {
...
 
 func InitializeDefaultContext() *HintRunnerContext {
-	return &HintRunnerContext{
+	ctx := &HintRunnerContext{
 		DictionaryManager:         DictionaryManager{},
 		SquashedDictionaryManager: SquashedDictionaryManager{},
-		ScopeManager:              *DefaultNewScopeManager(),
 		ConstantSizeSegment:       mem.UnknownAddress,
 	}
+	InitializeScopeManager(ctx, make(map[string]any)) // Initialize the ScopeManager with a default scope
```
TAdev0, line 32: well, now i look at it again, it seems that the current main branch returns `&HintRunnerContext` with `ScopeManager: *DefaultNewScopeManager(),` . This is actually the same as doing  `InitializeScopeManager(ctx, make(map[string]any))` after removing the ScopeManager field. 

@cicr99  what do you think? I'm not sure this is what was asked with issue #345  . This file is the hinter.go and the hinter effectively creates a new context with a main scope uninitialized.

TAdev0: I guess the code that should be modified is in the `runHinterTests` function in zerohint_test.go  file? 

here : 

```
		ctx := &hinter.HintRunnerContext{}
		if tc.ctxInit != nil {
			tc.ctxInit(ctx)
		}
```
when running each test, the context is reset, with all its fields unitialized

cicr99:
Hey @fishonamos 
For this issue what you need to modify is https://github.com/NethermindEth/cairo-vm-go/blob/b2602abd7a2901daba66ead1bc2805d200538ee2/pkg/hintrunner/zero/zerohint_test.go#L119 as @TAdev0 mentioned in his last comment
The idea is to initialize the scope manager for all tests and not having to do it manually.
In addition, you'll notice that there are some unit tests that have something like the following (see [here](https://github.com/NethermindEth/cairo-vm-go/blob/b2602abd7a2901daba66ead1bc2805d200538ee2/pkg/hintrunner/zero/zerohint_ec_test.go#L26C5-L28C7)):
```
	ctx.ScopeManager.EnterScope(map[string]any{})
},
```
or (see [here](https://github.com/NethermindEth/cairo-vm-go/blob/b2602abd7a2901daba66ead1bc2805d200538ee2/pkg/hintrunner/zero/zerohint_signature_test.go#L123C5-L125C7)):
```
	hinter.InitializeScopeManager(ctx, map[string]any{})
},
```
The idea for these, as you can see is just to initialize the ScopeManager, so once you modify the code to initialize it all the time, that won't be necessary, then those lines can be removed. Make sure to check all test cases for situations like these.

TAdev0:
@fishonamos did you close the PR by mistake?

fishonamos:
Just seeing this. I think it is a mistake.  But not an issue. I will raise
another today. I have found a way.

Issue #234, opened 2024-03-06
UsortVerifyMultiplicityAssert


cicr99:
Asserts that all occurrences of a specific value have been accounted for in the verification process.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L84

Depends on issue #245

PR #350, opened 2024-04-11
Issue/234 usort verify multiplicity assert

pxyxyrus:
Issue [#234](https://github.com/NethermindEth/cairo-vm-go/issues/234)

* Implement UsortVerifyMultiplicityAssert hint

* Added tests for UsortVerifyMultiplicityAssert hint

* Fixed inconsistent function name `newUsortEnterScopeHint` to `newUsortEnterScopeHinter`


rodrigo-pino:
fix #234

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -16,10 +17,51 @@ func TestZeroHintUsort(t *testing.T) {
...
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{
+						"positions": []*fp.Element{},
+					})
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newUsortVerifyMultiplicityAssertHinter()
+				},
+				check: func(t *testing.T, ctx *hintTestContext) {
+					positionsInterface, err := ctx.runnerContext.ScopeManager.GetVariableValue("positions")
```
MaksymMalicki, line 49: If I understand correctly, you are completely reimplementing the hint code inside this check function. It is not necessary, as you are invoking the hint itself in makeHinter field. Here you just need to check if error is nil (for non-empty arrays)

MaksymMalicki: check this method  [errorIsNil](https://github.com/NethermindEth/cairo-vm-go/blob/d50bca65c772c36627177d63e303dd65c9e199a0/pkg/hintrunner/zero/zerohint_utils_test.go#L201)

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -16,10 +17,51 @@ func TestZeroHintUsort(t *testing.T) {
...
+					})
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newUsortVerifyMultiplicityAssertHinter()
+				},
+				errCheck: errorTextContains("assertion `len(positions) == 0` failed"),
+			},
+			{
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{
```
MaksymMalicki, line 41: If I remeber correctly, on daily we agreed, that we will try to create the scope with this [method](https://github.com/NethermindEth/cairo-vm-go/blob/d50bca65c772c36627177d63e303dd65c9e199a0/pkg/hintrunner/hinter/scope.go#L17) and assign value later using the AssignVariable

pxyxyrus: updated

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -24,5 +27,34 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
+		Name: "UsortVerifyMultiplicityAssert",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			// assert len(positions) == 0
+			positionsInterface, err := ctx.ScopeManager.GetVariableValue("positions")
+
+			if err != nil {
+				return err
+			}
+
+			positions, ok := positionsInterface.([]*fp.Element)
```
MaksymMalicki, line 44: This is going to be an array of uint64 according to comments in this [PR](https://github.com/NethermindEth/cairo-vm-go/pull/358/files)

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -24,5 +26,34 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
 func createUsortEnterScopeHinter() (hinter.Hinter, error) {
-	return newUsortEnterScopeHint(), nil
+	return newUsortEnterScopeHinter(), nil
+}
+
+func newUsortVerifyMultiplicityAssertHinter() hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "UsortVerifyMultiplicityAssert",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			// assert len(positions) == 0
```
MaksymMalicki, line 36: nitpick: use //> at the beginning of the comment

MaksymMalicki: Other then that this lgtm. Can you take last look @cicr99 @rodrigo-pino

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -1,11 +1,13 @@
...
 
 import (
+	"fmt"
+
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
 )
 
-func newUsortEnterScopeHint() hinter.Hinter {
+func newUsortEnterScopeHinter() hinter.Hinter {
```
cicr99, line 10: No need to change the name here. If you check other files the format is `new<hintName>Hint` for the actual implementation of the hint. The method `create<hintName>Hinter` is for getting the parameters out of the resolver, if there are any and call this constructor

pkg/hintrunner/zero/zerohint_usort.go
cicr99: Yes, this is only for differentiate the hint code comments from regular comments in the go code

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -24,5 +26,34 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
+		Name: "UsortVerifyMultiplicityAssert",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			// assert len(positions) == 0
+			positionsInterface, err := ctx.ScopeManager.GetVariableValue("positions")
+
+			if err != nil {
+				return err
+			}
+
+			positions, ok := positionsInterface.([]uint64)
```
cicr99, line 43: Are we sure this should be casted to `[]uint64`. I was looking for the values that are inserted in the array and I think it comes back from this https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L16
Notice that `positions` comes from https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L66 and the values to that list are appended here https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L30
Now `i` goes until `input_len` which is a felt and that could have a max of 252 bits. `uint64` only gets to 64 bits

Is there something I'm missing here?

pxyxyrus: Hi, I was requested to make a change based on the discussion on the other PR for a related functionality [here](https://github.com/NethermindEth/cairo-vm-go/pull/358#discussion_r1570503119).

I had the original code in `felt` but to match with the whole [usort multiplicity body](https://github.com/NethermindEth/cairo-vm-go/pull/352) decided to change to `uint64`.

cicr99: Oh, I see that there is an assert where it's checked that `input_len <= __usort_max_size`, and `__usort_max_size` is set to 2**20 somewhere else, so it makes sense to assume the value of `input_len` fits in a uint64

However, I wonder, if `__usort_max_size` is None [here](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L21), is it possible to get a value from input_len that can only fit in a felt? @MaksymMalicki

MaksymMalicki: Yea, thats a good catch @cicr99, I think it will be safer to store the input_len as Felt, so that all the possible cases are handled. On the other side, the __usort_max_size is assigned to 2**20 at an entrypoint function, but I'm not sure if it's always accessible. Rust VM has the __usort_max_size bounded by uint64, I wonder if it is an overlook or intential.

MaksymMalicki: Okey, so after conversation with @cicr99 it is not possible for the __usort_max_size to be greater than 2**20, so the uint64 is ok for this type. We can throw some assertion error, if there is a problem with type casting to uint64. Other than that LGTM.

cicr99 (approved):

PR #351, opened 2024-04-12
Implement DictNew hint

har777:
Implements #288


pkg/hintrunner/zero/zerohint_dictionaries_test.go
```diff
@@ -0,0 +1,75 @@
...
+						valueFelt, err := value.FieldElement()
+						if err != nil {
+							t.Fatalf("mv: %s cannot be converted to felt", value)
+						}
+						expectedValueFelt := new(fp.Element).Mul(&key, new(fp.Element).SetUint64(100))
+
+						if !valueFelt.Equal(expectedValueFelt) {
+							t.Fatalf("at key: %v expected: %s actual: %s", key, expectedValueFelt, valueFelt)
+						}
+					}
```
rodrigo-pino, line 70: There is `NewElement` method for `fp.Element` which takes a uint64 and it is by default a value type.

pkg/hintrunner/zero/zerohint.go
```diff
@@ -107,7 +107,7 @@ func GetHintFromCode(program *zero.ZeroProgram, rawHint zero.Hint, hintPC uint64
 		return createUint256SqrtHinter(resolver)
 	case uint256MulDivModCode:
 		return createUint256MulDivModHinter(resolver)
-	// Signature hints
+		// Signature hints
```
TAdev0, line 110: i think they are all shifted and only the ` // Match hints ` (first one) is in the right place

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -10,6 +10,60 @@ import (
...
+			//> del initial_dict
+			return ctx.ScopeManager.DeleteVariable("initial_dict")
+		},
+	}
+}
+
+func createDictNewHinter() (hinter.Hinter, error) {
+	return newDictNewHint(), nil
+}
+
```
TAdev0, line 66: lgtm

pkg/hintrunner/zero/zerohint_dictionaries_test.go
```diff
@@ -10,6 +10,68 @@ import (
...
+						}
+						expectedValueFelt := new(fp.Element).Mul(&key, feltUint64(100))
+
+						if !valueFelt.Equal(expectedValueFelt) {
+							t.Fatalf("at key: %v expected: %s actual: %s", key, expectedValueFelt, valueFelt)
+						}
+					}
+				},
+			},
+		},
```
TAdev0, line 74: great !

TAdev0 (approved):
@har777  LGTM!

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -10,6 +10,60 @@ import (
 	f "github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
+// DictNew hint creates a new dictionary
+//
+// `newDictNewHint` takes no operander as argument
+func newDictNewHint() hinter.Hinter {
```
rodrigo-pino, line 16: Let's explain a little more. What's the difference between DictNew and DefaultDictNew

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -10,6 +10,60 @@ import (
...
+			//> if '__dict_manager' not in globals():
+			//>   from starkware.cairo.common.dict import DictManager
+			//>   __dict_manager = DictManager()
+			//>
+			//> memory[ap] = __dict_manager.new_dict(segments, initial_dict)
+			//> del initial_dict
+
+			//> if '__dict_manager' not in globals():
+			//>   from starkware.cairo.common.dict import DictManager
+			//>   __dict_manager = DictManager()
```
rodrigo-pino, line 29: Could you share with me the source of this code. I want to look at smth

TAdev0: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/dict.cairo#L9

pkg/hintrunner/zero/zerohint_dictionaries_test.go
```diff
@@ -10,6 +10,68 @@ import (
...
+					for _, key := range []fp.Element{*feltUint64(10), *feltUint64(20), *feltUint64(30)} {
+						value, err := dictionary.At(key)
+						if err != nil {
+							t.Fatalf("error fetching value for key: %v", key)
+						}
+						valueFelt, err := value.FieldElement()
+						if err != nil {
+							t.Fatalf("mv: %s cannot be converted to felt", value)
+						}
+						expectedValueFelt := new(fp.Element).Mul(&key, feltUint64(100))
```
cicr99, line 66: Why would you have to do the multiplication? can you just use the expected values from the beginning? What am I missing here?

har777: Ah thats because the expected values are being checked in the loop. `value = key * 10 * 100` in the test. So it was just easier to iterate over the keys, calculate the expected value as key * 100 and then assert.

pkg/hintrunner/zero/zerohint.go
TAdev0: yeah it shouldnt be. At some point, some of the comment for hints were placed like this. We aligned all of them correctly in a previous PR.

har777: My bad this is a mistake. I'll push a fix.

cicr99 (approved):

PR #355, opened 2024-04-17
feat: Replace secp_utils usage of `big.Int` with uint256 library.

### LINKED ISSUE #353: Replace secp_utils usage of `big.Int` with uint256 library.

TAdev0:
Resolves #353


TAdev0:
@rodrigo-pino  i have a few questions : 

should i also modify `SecPPacked` and `SecPSplit` function? 

Also, `GetSecPBig` for example now returns a `uint256.Int`, i guess i should also modify the code where this utils is used (i.e., zerohint_ec.go and zerohint_signature(s).go ) ?

rodrigo-pino:
Hey @TAdev0! Yes please add both functions since their values are never bigger than the 256 bits space. Regarding the code that uses them, it must be updated as well since we need the CI to pass in order to merge.

In case the consecuences of using uint64 library are very hard to integrate we can do smth good enough now and follow it up with another PR focused on that

TAdev0:
@rodrigo-pino  ready for a first review. Not sure everything is correct with my syntax.

rodrigo-pino:
There are some errors. Especially compilations one. Could you please fix them.

Also you can test most of the CI by doing `make unit`.
This will compile & run all unit tests.

TAdev0:
@rodrigo-pino  yeah thank you! i'm installing Go right now, will do that.

TAdev0:
@rodrigo-pino I have residual mismatch issues, for example:

```
   zerohint_utils_test.go:163: value scope value mismatch:
            have: 25711014748331347274616151948719391133326249308066029554892800
            want: 25711014748331348032841661176477624755444844903972403171819520
```

I double checked that i correctly define constants in `GetSecP` , `getBase` and `GetSecp256R1_P`

pkg/hintrunner/utils/secp_utils.go
```diff
@@ -2,68 +2,69 @@ package utils
...
 	}
 
 	return split, nil
 }
 
-func GetSecp256R1_P() (*big.Int, bool) {
+func GetSecp256R1_P() uint256.Int {
 	// 2**256 - 2**224 + 2**192 + 2**96 - 1
-	return new(big.Int).SetString("115792089210356248762697446949407573530086143415290314195533631308867097853951", 10)
+	result, _ := uint256.FromHex("0xFFFFFFFF00000001000000000000000000000000FFFFFFFFFFFFFFFFFFFFFFFF")
```
rodrigo-pino, line 67: Please, don't parse strings, create the uint256 directly

rodrigo-pino:
> I suspect that the issue comes from SecPPacked as it is the common denominator for all failing tests. But i'm struggling debugging efficiently, i checked if it could come from DivMod which behaves differently with BigInt and uint256, but it seems ok.

I see that's weird, maybe something to do with references? My tip for debugging would be testing the failing functions one by one with some prints scattered around. You'll eventually find the wrong piece of code

TAdev0:
@rodrigo-pino  comments addressed. I still need to figure out this issue.

pkg/hintrunner/utils/secp_utils.go
```diff
@@ -2,68 +2,78 @@ package utils
...
 
-	return packedBig, nil
+	return *packedUint256, nil
 }
 
-func GetBetaBig() *big.Int {
-	return big.NewInt(7)
+func GetBetaUint256() uint256.Int {
+	return *uint256.NewInt(7)
 }
```
rodrigo-pino, line 48: Oops, this one as well please

pkg/hintrunner/utils/secp_utils.go
```diff
@@ -2,68 +2,78 @@ package utils
...
+	baseUint256 := getBaseUint256()
 
-	var residue big.Int
 	for i := 0; i < 3; i++ {
-		num.DivMod(num, baseBig, &residue)
-		split[i] = new(big.Int).Set(&residue)
+		var residue uint256.Int
+		num.DivMod(num, &baseUint256, &residue)
+		item := uint256.NewInt(0).Set(&residue)
+		split[i] = *item
```
rodrigo-pino, line 61: Why cannot store residue directly in split[i]?

rodrigo-pino:
Ok left some other small ones, regarding the issue, I'll try to look it later. Can you debug it manually with prints?

pkg/hintrunner/utils/secp_utils.go
TAdev0: a question i asked myself. I thought there was some reason specific to go but it seems useless indeed

pkg/hintrunner/utils/secp_utils.go
TAdev0: done

TAdev0:
@rodrigo-pino  ok so basically, i fail for all `EcNegate` tests (and others, but i think the root cause is the same). These tests call `newEcNegateHint` , which will use `GetSecPUint256` and `SecPPacked` among other stuff i didnt change. 

`GetSecPUint256` returns a const which should be fine.
I tried printing different variables and indeed, the result value of `newEcNegateHint`, which is `yBig` , is always a little bit different, with generally 8 identical bytes and the rest different.

`yBig` is the NEG + MOD result of `SecPPacked` , and `SecPPacked` seems ok as well, using `getBaseUint256` . 

I even tried keeping BigInt for `SecPPacked` computation, without any modification, and i still have these errors.

har777:
I did a super quick check. In `SecPPacked` before returning the `packedBig` variable I tried converting it to a `uint256` and then back to a `big.Int` and tests failed.

Before:
```
return *packedBig, nil
```

After:
```
packedUint256 := new(uint256.Int)
packedUint256.SetFromBig(packedBig)

return *packedUint256.ToBig(), nil
```

The failing tests seem to be using quite large values though. Could it be the test values are just unrealistic? Though I would say the results should match the python vm regardless.

rodrigo-pino:
By looking at the python VM code you immediately realize that input values are always less than 256 bits, and output values are always smaller as well. This means that assuming the current implementation is correct, the one with big.Ints was flawed. It is possible the tests are actually not returning a correct value by using wrong inputs.

The python and the go vm should always return the same value, but if the python vm expect always it's input to be in the 256 bits range, whatever we throw at it bigger than it is not valid for comparison. We are loooking to match the python vm in the 256 bit range not on bigger stuff, assuming of course, that to the python VM you could feed bigger stuff, to which there is not test to show it.

rodrigo-pino:
If there is a test that inputs a number bigger than 256 bits or returns as an output a number bigger than 256 bits, that test is not correct.

har777:
> If there is a test that inputs a number bigger than 256 bits or returns as an output a number bigger than 256 bits, that test is not correct.

The inputs are not larger than 256 bits in the tests. The outputs can be. The output can be a scope value and not a memory value.

har777:
In the case of secp packed the result should be a 256 bit number though.

MaksymMalicki:
> > If there is a test that inputs a number bigger than 256 bits or returns as an output a number bigger than 256 bits, that test is not correct.
> 
> The inputs are not larger than 256 bits in the tests. The outputs can be. The output can be a scope value and not a memory value.

If it works like this, then it is also problematic in terms of replacing scope any interface with the new struct, if the values are bigger then 256 bits, but that is not relevant in this scope.

TAdev0:
The weird thing is that my impl with uint256 is supposed to be the same. I guess there is some issue related to bigint to uint256 conversion  if all inputs are indeed < 256bits

rodrigo-pino:
> The inputs are not larger than 256 bits in the tests. The outputs can be. The output can be a scope value and not a memory value.

How can they be? The only way to get a higher than 256 bits is in packed right? Looking at the Cairo VM description it is stated that the number that needs to be formed needs to be in that range.

rodrigo-pino:
> ... The output can be a scope value and not a memory value.

I am not convinced either that a scope value can be bigger than 256 bits. Obviously the python VM allows it (python allows many things by itself). The question is, is there any Cairo 0 code that could be written that could create a number bigger than 256 bits in the scope.

Also, remember that the Scope was defined by us as well, so using it as an argument of what it can have inside or not is not valid since we can always redefine it.

rodrigo-pino:
> The weird thing is that my impl with uint256 is supposed to be the same. I guess there is some issue related to bigint to uint256 conversion for the output if all inputs are indeed < 256bits ...

I think you've solved it correctly and we are finding some inconsistencies unrelated with them. After this PR the VM will be better

har777:
> > The inputs are not larger than 256 bits in the tests. The outputs can be. The output can be a scope value and not a memory value.
> 
> How can they be? The only way to get a higher than 256 bits is in packed right? Looking at the Cairo VM description it is stated that the number that needs to be formed needs to be in that range.
> 
> That Python doesn't have statically typed values doesn't mean we have to follow through if it doesn't make sense. I haven't studied this part long enough, but it feels to me that both `pack` and `split` are meant to use in tandem. So I don't see why the inputs for`pack` should be bigger than 256 bits

1. No input in the tests are larger than 256 bits. 
2. It's possible that the 3 values passed to pack is resulting in a value greater than 256 bits in the tests. I haven't checked this.
3. I'm guessing the programmer must assert the input ranges of the arguments to `pack` in the cairo code. If they don't then the python vm and our vm will result in differing results written to the scope. I don't know if any programmer will have a valid reason for doing something like this.

TAdev0:
@har777  indeed if i replace `SecPPacked` with the previous version using bigInt, tests for ec pass. I confirm that the value after exponentiation and packing in `SecPPacked` will be greater than `type(uint256).max` . But i still have 2 tests failing  in this configuration: 

```
--- FAIL: TestZeroHintEc (0.00s)
    --- FAIL: TestZeroHintEc/NondetBigint3V1_3 (0.00s)
        zerohint_utils_test.go:195: expected an error containing "num != 0", got nil err
--- FAIL: TestVerifyZeroHint (0.00s)
    --- FAIL: TestVerifyZeroHint/VerifyZero_1 (0.00s)
        zerohint_utils_test.go:197: 
                Error Trace:    /Users/tristan/Desktop/Dev/cairo-vm-go/pkg/hintrunner/zero/zerohint_utils_test.go:197
                                                        /Users/tristan/Desktop/Dev/cairo-vm-go/pkg/hintrunner/zero/zerohint_test.go:255
                                                        /Users/tristan/Desktop/Dev/cairo-vm-go/pkg/hintrunner/zero/zerohint_test.go:267
                Error:          Error "verify_zero: Invalid input (1, 1, 1)" does not contain "verify_zero: Invalid input (1, 1, 1)."
                Test:           TestVerifyZeroHint/VerifyZero_1
                ```

**Comment by har777** [comment]
I dont think so. I tried changing just that part:
```
num2 := new(uint256.Int)
num2.SetFromBig(num)
if num2.Cmp(uint256.NewInt(0)) != 0 {
  return nil, fmt.Errorf("num != 0")
}
```
And it passed.

**Comment by TAdev0** [comment]
I also tried for DivMod it passes as well

**Comment by TAdev0** [comment]
@har777 you want to remove all tests? what is the reason for the 2 others failing?

**Comment by har777** [comment]
> @har777 you want to remove all failing tests? what is the reason for the 2 others failing? I cannot find any reason for now.

Just this one:
```
    --- FAIL: TestZeroHintEc/NondetBigint3V1_3 (0.00s)
        zerohint_utils_test.go:195: expected an error containing "num != 0", got nil err
```

**Comment by TAdev0** [comment]
mmh i dont understand. If i remove this test, i still have many tests failing.

**Comment by TAdev0** [comment]
@rodrigo-pino  @har777  I annotated all failing tests so that tests pass for the CI. 

For now, I use uint256 for everything in ec utils.

It includes :  
- a bunch of `EcNegate` 
-  one `FastEcAddAssignNewX` test
- one `VerifyECDSASignature` test
- one `GetPointFromX` test
- one `VerifyZero` test

File: pkg/hintrunner/utils/secp_utils.go [hunk 4]
```diff
@@ -2,68 +2,83 @@ package utils
...
+	packedUint256 := uint256.NewInt(0)
 	for idx, limb := range limbs {
-		limbBig := AsInt(limb)
-		valueToAddBig := new(big.Int).Exp(baseBig, big.NewInt(int64(idx)), nil)
-		valueToAddBig.Mul(valueToAddBig, limbBig)
-		packedBig.Add(packedBig, valueToAddBig)
+		limbUint256, _ := uint256.FromBig(AsInt(limb))
+		valueToAddUint256 := uint256.NewInt(0).Exp(&baseUint256, uint256.NewInt(uint64(int64(idx))))
+		valueToAddUint256.Mul(valueToAddUint256, limbUint256)
+		packedUint256.Add(packedUint256, valueToAddUint256)
```
rodrigo-pino, line 40: This as a nitpick but I dont see the need to suffix all the variable names with their type, perhaps only limb. I am aware it was like that before, could you please update it.

pkg/hintrunner/utils/secp_utils.go
```diff
@@ -2,68 +2,83 @@ package utils
...
+	baseUint256 := getBaseUint256()
 
-	packedBig := new(big.Int)
+	packedUint256 := uint256.NewInt(0)
 	for idx, limb := range limbs {
-		limbBig := AsInt(limb)
-		valueToAddBig := new(big.Int).Exp(baseBig, big.NewInt(int64(idx)), nil)
-		valueToAddBig.Mul(valueToAddBig, limbBig)
-		packedBig.Add(packedBig, valueToAddBig)
+		limbUint256, _ := uint256.FromBig(AsInt(limb))
```
rodrigo-pino, line 37: This conversion is not ideal. The best conversion is:
1. felt252 to bytes
2. bytes to uint256

smth like this:

```go
limbUint256 = new(uint256.Int).SetBytes(limb.Bytes()[:])
```

pkg/hintrunner/zero/zerohint_signatures.go
```diff
@@ -81,7 +82,7 @@ func newGetPointFromXHinter(xCube, v hinter.ResOperander) hinter.Hinter {
 			}
 
 			//> from starkware.cairo.common.cairo_secp.secp_utils import SECP_P, pack
-			secpBig, _ := secp_utils.GetSecPBig()
+			secpUint256 := secp_utils.GetSecPUint256()
```
rodrigo-pino, line 85: Regarding variable names here as well, I don't think we need to append the type behind it. We can explicitly declare their type if you think it is needed for readability.

rodrigo-pino: This happen overall in this function

TAdev0: i remove it . For `vBig := v.BigInt(new(big.Int))`. i need to keep it as using `v` will not compile

TAdev0:
@rodrigo-pino  comments addressed. I left the type where it makes sense to leave it, and remove in all other places

TAdev0:
@rodrigo-pino  had a few conflicts, i have other tests failing now. It is getting a bit difficult to understand whats going on, maybe should we close this and do it once all hints are implemented? Or maybe merging this soon, because there will be additionnal conflicts after PR are merged

rodrigo-pino:
Hey @TAdev0 the solution is good, let's put it on hold until we can asses with clarity the `uint256` vs `BigInt` situation

TAdev0:
sure

cicr99:
I'm adding the `blocked` label here, so we know we won't be working on this rn

cicr99:
We want to wait until the hints are all merged, this means there are going to be a lot of conflicts to fix and a lot more of code to modify. I'm closing this PR for the moment, we can open a new one with fresh changes

Issue #233, opened 2024-03-06
UsortVerify


cicr99:
Prepares for verifying the multiplicity of a specific value in the sorted output by reversing its positions list.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L64

Probably depends on how issue #245 gets implemented, but once that's decided this one should be straight forward

PR #357, opened 2024-04-17
Issue/233 usort verify

pxyxyrus:
Issue https://github.com/NethermindEth/cairo-vm-go/issues/233

Implement UsortVerify hint

Added test for UsortVerify hint

### DIFF
--- a/pkg/hintrunner/zero/zerohint_usort.go
+++ b/pkg/hintrunner/zero/zerohint_usort.go
@@ -1,8 +1,12 @@
 package zero
 
 import (
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -26,3 +30,48 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
+				return err
+			}
+
+			positions := positionsDict[value]
+			utils.Reverse(positions)
+
+			ctx.ScopeManager.AssignVariables(map[string]any{
+				"last_pos":  0,
+				"positions": positions,
+			})
```
pxyxyrus, line 62: Should I return a feltUint64 here?

PR #352, opened 2024-04-12
Implement UsortBody hint

MaksymMalicki:

PR #358, opened 2024-04-17
Issue/233 usort verify

pxyxyrus:
Issue https://github.com/NethermindEth/cairo-vm-go/issues/233

Implement UsortVerify hint

Added test for UsortVerify hint


pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -26,3 +30,52 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
+	return &GenericZeroHinter{
+		Name: "UsortVerify",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			positionsDictInterface, err := ctx.ScopeManager.GetVariableValue("positions_dict")
+
+			if err != nil {
+				return err
+			}
+
+			positionsDict, ok := positionsDictInterface.(map[uint64][]*fp.Element)
```
MaksymMalicki, line 44: I belive if you check [this](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L8) function, you will see, that verify usort is being called at the end and is being executed in the same scope as this [PR](https://github.com/NethermindEth/cairo-vm-go/pull/352/files) (implements usort method). Notice that positionsDict is of type map[fp.Element][]uint64. Make sure that all the types are compatible with this PR. 
I noticed, that in my [PR](https://github.com/NethermindEth/cairo-vm-go/pull/352/files) I need to pass positionsDict into the scope for you to read it in this hint, but I don't know if we can pass maps into scopes or is it the functionality of a DictManager. @rodrigo-pino do you have any thoughts?

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -26,3 +30,52 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
+
+			if !ok {
+				return fmt.Errorf("casting positions_dict into an dictionary failed")
+			}
+
+			value, err := hinter.ResolveAsUint64(vm, value)
```
MaksymMalicki, line 50: same here, to make those hints compatible, the value must be a Felt

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -21,5 +23,41 @@ func TestZeroHintUsort(t *testing.T) {
 				check: varValueInScopeEquals("__usort_max_size", feltUint64(1)),
 			},
 		},
+		"UsortVerify": {
+			{
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					hinter.InitializeScopeManager(ctx, map[string]any{
+						"positions_dict": map[uint64][]*fp.Element{
+							0: {feltUint64(1), feltUint64(2), feltUint64(3)},
```
MaksymMalicki, line 31: change type of positions_dict

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -21,5 +23,41 @@ func TestZeroHintUsort(t *testing.T) {
...
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newUsortVerifyHinter(ctx.operanders["value"])
+				},
+				check: func(t *testing.T, ctx *hintTestContext) {
+					positions, err := ctx.runnerContext.ScopeManager.GetVariableValue("positions")
+					require.NoError(t, err)
+
+					require.Equal(t, []*fp.Element{feltUint64(3), feltUint64(2), feltUint64(1)}, positions)
+
+					lastPos, err := ctx.runnerContext.ScopeManager.GetVariableValue("last_pos")
```
MaksymMalicki, line 47: I think we have an utility for hits called varValueInScopeEquals()

MaksymMalicki (changes requested):

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -1,14 +1,21 @@
...
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
 func newUsortEnterScopeHint() hinter.Hinter {
 	return &GenericZeroHinter{
 		Name: "UsortEnterScope",
 		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			//> last_pos = 0
+			//> positions = positions_dict[ids.value][::-1]
+
```
cicr99, line 18: This is not the correct code for this hint. The one corresponding to this hint should be this [one](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L15)

cicr99: I just noticed you didn't add the code in the constructor of the `UsortVerify` hint, so it was probably that you wrote it in this method by mistake. Could you please add the corresponding codes in both functions?

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -26,3 +33,52 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
+			err = ctx.ScopeManager.AssignVariables(map[string]any{
+				"last_pos":  0,
+				"positions": positions,
+			})
+
+			if err != nil {
+				return err
+			}
+
+			return nil
```
cicr99, line 71: You can just return `err` at the end, so even better, just directly return the result of `AssignVariables` method
```
return ctx.ScopeManager.AssignVariables(map[string]any{
	"last_pos":  0,
	"positions": positions,
})
```

cicr99 (approved):

MaksymMalicki (approved):

PR #359, opened 2024-04-18
Implement MemcpyContinueCopying hint

### LINKED ISSUE #251: MemcpyContinueCopying
Continues copying memory elements, decrementing the remaining length and setting a flag to continue if more elements are to be copied.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/memcpy.cairo#L24

Depends on #245

danielcdz:
resolves #251


pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,58 @@ package zero
...
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
 )
 
+func newMemcpyContinueCopyingHinter(n, dst hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "MemcpyContinueCopying",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			//n -= 1
+			//ids.continue_copying = 1 if n > 0 else 0
+			var lhs fp.Element
```
MaksymMalicki, line 19: Do we need this variable? In the ```AssignVariable("n")``` we can just create new felt and set it to n-1, like this:
```ctx.ScopeManager.AssignVariable("n", new(fp.Element).Sub(n.(*fp.Element), &utils.FeltOne))

File: pkg/hintrunner/zero/zerohint_memcpy.go [hunk 2]
```diff
@@ -3,9 +3,58 @@ package zero
...
+
+			// n-=1
+			n, err := ctx.ScopeManager.GetVariableValue("n")
+			if err != nil {
+				return err
+			}
+			ctx.ScopeManager.AssignVariable("n", lhs.Sub(n.(*fp.Element), &utils.FeltOne))
+
+			// ids.continue_copying = 1 if n > 0 else 0
+			continue_copying, err := dst.GetAddress(vm)
```
MaksymMalicki, line 29: nitpick: use cammelcase and maybe add some Addr suffix to the variable

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,58 @@ package zero
...
+			} else {
+				v = memory.MemoryValueFromFieldElement(&utils.FeltZero)
+			}
+			return vm.Memory.WriteToAddress(&continue_copying, &v)
+		},
+	}
+}
+
+func createMemcpyContinueCopyingHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
+	 n, err := resolver.GetResOperander("n")
```
MaksymMalicki, line 46: we don't need this, as we are retrieving ```n``` from the scope

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,58 @@ package zero
...
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/core"
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
+	"github.com/NethermindEth/cairo-vm-go/pkg/vm/memory"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
 )
 
+func newMemcpyContinueCopyingHinter(n, dst hinter.ResOperander) hinter.Hinter {
```
MaksymMalicki, line 13: as stated in the comment at the bottom, you can remove n (it is retrieved from the context) and rename dst to output, so that the references names remain consistend

MaksymMalicki:
Also please run ```golangci-lint run``` or ```gofmt -w .``` to fix the lint problems. Thanks!

pkg/hintrunner/zero/zerohint_memcpy.go
cicr99: Hey @MaksymMalicki for me the code block you added in your comment is empty

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,58 @@ package zero
...
+			ctx.ScopeManager.AssignVariable("n", lhs.Sub(n.(*fp.Element), &utils.FeltOne))
+
+			// ids.continue_copying = 1 if n > 0 else 0
+			continue_copying, err := dst.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+
+			var v memory.MemoryValue
+			if !utils.FeltIsPositive(n.(*fp.Element)) {
```
cicr99, line 35: You want to check here the condition for `n>0`, see that this is different from checking `FeltIsPositive`.
`FeltIsPositive` looks like the following:
```
func FeltIsPositive(felt *fp.Element) bool {
	// range_check_builtin.bound is utils.FeltMax128 (1 << 128).
	return FeltLt(felt, &FeltMax128)
}
```
This means it checks the number is smaller than 1<<128, which works on felts to determine whether you'd consider the number positive or negative. However, even if you initialized `n` with a value that fits in a felt, in the python code of the hint, `n` is treated as an integer, not a felt, so you are actually comparing it to 0.

So the correct condition here for `n>0` should be something like `FeltLt(FeltZero, n)`

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,58 @@ package zero
...
+				v = memory.MemoryValueFromFieldElement(&utils.FeltOne)
+			} else {
+				v = memory.MemoryValueFromFieldElement(&utils.FeltZero)
```
cicr99, line 38: Also, I think you got this in the opposite order :sweat_smile:

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,58 @@ package zero
...
+		},
+	}
+}
+
+func createMemcpyContinueCopyingHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
+	 n, err := resolver.GetResOperander("n")
+	if err != nil {
+		return nil, err
+	}
+	output, err := resolver.GetResOperander("output")
```
cicr99, line 50: See that the actual variable involved in the code is called `continue_copying` so it will appear in the resolver with the same name. You'll probably get an error trying to retrieve the "output" variable actually, since it doesn't have to exist in the resolver at this point

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: @MaksymMalicki same for me, there is no code in that block 😢

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: done! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: done! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: done! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,58 @@ package zero
...
+
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
 )
 
+func newMemcpyContinueCopyingHinter(n, dst hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "MemcpyContinueCopying",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			//n -= 1
+			//ids.continue_copying = 1 if n > 0 else 0
```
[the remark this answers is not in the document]
danielcdz: done! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: ty for the explanation! I corrected the if statement! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: oh, that's true, I corrected it now! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: completely true, ty for explaining that! is fixed now! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,58 @@ package zero
...
+			var lhs fp.Element
+
+			// n-=1
+			n, err := ctx.ScopeManager.GetVariableValue("n")
+			if err != nil {
+				return err
+			}
+			ctx.ScopeManager.AssignVariable("n", lhs.Sub(n.(*fp.Element), &utils.FeltOne))
```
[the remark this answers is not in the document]
danielcdz: I'm waiting for @MaksymMalicki response on the comment above about creating a new felt an set it to n-1, looking at the codebase idk exactly how to do that 😞

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,54 @@ package zero
...
+
+			//> ids.continue_copying = 1 if n > 0 else 0
+			continueCopyingAddr, err := output.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+
+			var v memory.MemoryValue
+			if !utils.FeltLt(&utils.FeltZero, n) {
```
TAdev0, line 36:
```
if utils.FeltLt(&utils.FeltZero, n) {
      v = memory.MemoryValueFromFieldElement(&utils.FeltOne)
} else {
      v = memory.MemoryValueFromFieldElement(&utils.FeltZero)
}
```

this removes the `!` operator and respect the order of the python hint (strictly no importance though ^^)

pkg/hintrunner/zero/zerohint_memcpy.go
TAdev0: @danielcdz  i think you can simplify and do something like this : 

```
n, err := ctx.ScopeManager.GetVariableValue("n")
if err != nil {
	return err
}

ctx.ScopeManager.AssignVariable("n", n.(uint64)-1)
``` 

so that you dont need to declare any additionnal variable.

TAdev0: and then : 

```
var v memory.MemoryValue
if n.(uint64) > 0 {
	v = memory.MemoryValueFromFieldElement(&utils.FeltZero)
} else {
	v = memory.MemoryValueFromFieldElement(&utils.FeltOne)
}
```

TAdev0 (changes requested):

pkg/hintrunner/zero/hintcode.go
```diff
@@ -85,6 +85,7 @@ const (
 
 	// ------ Other hints related code ------
 	allocSegmentCode string = "memory[ap] = segments.add()"
+	memcpyContinueCopyingCode string = "n -= 1\nids.continue_copying = 1 if n > 0 else 0"
 	vmEnterScopeCode string = "vm_enter_scope()"
```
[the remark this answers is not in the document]
danielcdz: gotcha, is fixed now! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,54 @@ package zero
...
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/core"
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
+	"github.com/NethermindEth/cairo-vm-go/pkg/vm/memory"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
 )
 
+func newMemcpyContinueCopyingHinter(output hinter.ResOperander) hinter.Hinter {
```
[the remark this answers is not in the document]
danielcdz: is fixed now! 🚀

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: fixed now! ✅

danielcdz: thank you! @TAdev0

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,54 @@ package zero
...
+
+			//> ids.continue_copying = 1 if n > 0 else 0
+			continueCopyingAddr, err := output.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+
+			var v memory.MemoryValue
+			if n.(uint64) > 0 {
+				v = memory.MemoryValueFromFieldElement(&utils.FeltZero)
```
TAdev0, line 36: Inversion of FeltZero and FeltOne
I also inverted in my comment sorry, i modified it now.

hint is : 
```
//> ids.continue_copying = 1 if n > 0 else 0
```

so : 
```
if n.(uint64) > 0 {
	v = memory.MemoryValueFromFieldElement(&utils.Feltone)
} ...
```

TAdev0 (changes requested):
great now you can add tests!

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: Solved! ✅

danielcdz:
> great now you can add tests!

I didn't find any `zerohint_memcpy_test.go` file, should I create a new test file?

TAdev0:
@danielcdz yes!

danielcdz:
Hey, @TAdev0 I added the test cases and added a line in `zerohint_memcpy.go` to store the new `n` value to have the correct value in the if statement.

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,54 @@ package zero
...
+			n = n.(uint64) - 1
+			ctx.ScopeManager.AssignVariable("n", n)
+
+			//> ids.continue_copying = 1 if n > 0 else 0
+			continueCopyingAddr, err := output.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+
+			var v memory.MemoryValue
```
har777, line 34: I would prefer this be called `continueCopyingMv` instead of `v`

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,54 @@ package zero
...
+			if n.(uint64) > 0 {
```
har777, line 35: Isn't `n` already of the required type from the prev line `n = n.(uint64) - 1`

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,54 @@ package zero
 import (
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/core"
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
+	"github.com/NethermindEth/cairo-vm-go/pkg/vm/memory"
+
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
 )
 
+func newMemcpyContinueCopyingHint(output hinter.ResOperander) hinter.Hinter {
```
har777, line 12: Imo this should be `continueCopying hinter.ResOperander` not `output hinter.ResOperander`

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,54 @@ package zero
...
+				v = memory.MemoryValueFromFieldElement(&utils.FeltZero)
+			}
+
+			return vm.Memory.WriteToAddress(&continueCopyingAddr, &v)
+		},
+	}
+}
+
+func createMemcpyContinueCopyingHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
+	output, err := resolver.GetResOperander("continue_copying")
```
har777, line 47: Better to name the variable `continueCopying`

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,54 @@ package zero
...
+			//> n -= 1
+			//> ids.continue_copying = 1 if n > 0 else 0
+
+			//> n-=1
+			n, err := ctx.ScopeManager.GetVariableValue("n")
+			if err != nil {
+				return err
+			}
+
+			n = n.(uint64) - 1
```
har777, line 25: Are we sure it should be `uint64` and not just a `felt`? I quickly went through some code in the cairo codebase and `uint64` most likely is fine. But the code does seem to indicate felt.

eg.
```
func append_felts{data: felt*}(len: felt, arr: felt*) {
    memcpy(dst=data, src=arr, len=len);
    let data = data + len;
    return ();
}
```
Here the `len` variable is of type `felt`. So maybe safer to just use that? cc @cicr99

pkg/hintrunner/zero/zerohint_memcpy_test.go
```diff
@@ -0,0 +1,51 @@
...
+	runHinterTests(t, map[string][]hintTestCase{
+		"MemcPyContinueCopying": {
+			{
+				operanders: []*hintOperander{
+					{Name: "continue_copying", Kind: uninitialized},
+				},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{})
+					err := ctx.ScopeManager.AssignVariables(map[string]any{
+						"n": uint64(1),
```
har777, line 20: Can we add one more test with `"n": uint64(0)`

har777 (changes requested):

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,54 @@ package zero
...
+			ctx.ScopeManager.AssignVariable("n", n)
```
har777, line 26:
```
err = ctx.ScopeManager.AssignVariable("n", n)
if err != nil {
    return err
}
```

har777 (changes requested):

pkg/hintrunner/zero/zerohint_memcpy.go
har777: Type casting like `n = n.(uint64) - 1` can also fail.

I would check and return an error in that case.

eg:

```
n, ok = n.(uint64) - 1
if !ok {
    return fmt.Errorf("value: %s is not a uint64", n)
}

File: pkg/hintrunner/zero/zerohint_memcpy.go (hunk 20 above)
Comment (danielcdz) [reply pkg/hintrunner/zero/zerohint_memcpy.go:26]: what about: 
``` go
if err := ctx.ScopeManager.AssignVariable("n", n); err != nil {
				return err
			}

```

danielcdz: done! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: sure that sounds better! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: Done! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: sure, it is done! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: when I tried to leave just `n > 0` it didn't work since it is treated as an interface and the comparison can't be done

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: so we should work with `n` as uint or change it to a felt? in case to treat `n` as felt, how can I cast the type to felt?

pkg/hintrunner/zero/zerohint_memcpy_test.go
danielcdz: sure!

danielcdz: I found that the uint64 type only supports this range of numbers: 
```go
// uint64 is the set of all unsigned 64-bit integers.
// Range: 0 through 18446744073709551615.
type uint64 uint64

```

can this be another reason to use felt instead of uint ?

pkg/hintrunner/zero/zerohint_memcpy.go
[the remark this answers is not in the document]
MaksymMalicki: Looks good to me, sorry for long answer

TAdev0:
@danielcdz  you have a few conflicts to resolve

pkg/hintrunner/zero/zerohint_memcpy.go
TAdev0: if `n` has already been typecasted to a `uint64` it should work

pkg/hintrunner/zero/zerohint_memcpy.go
TAdev0: Probably better with felt, @cicr99  ?

cicr99: So the thing here is the variable was assigned in the scope with a felt value. I think at the moment of definition you have `n = ids.len`, where `len` is a felt, so when you access the variable, typecasting it to uint64 should not work as it would be an `Element` (according to the library we are using for felts). Then I would go with felt here.

This being said, one note on this. In practice, the actual value of n cannot be a value of 252 bits since that wouldn't be feasible to compute, as this is representing the length of an array. n should fit in uint64. So if from the moment this variable is assigned to the scope we do the conversion, then your code would be good as it is. This will imply modifying a different part of the code. Plus, I think this is a situation that has happened other times in the code, so I would like to handle that in a different PR. I'm creating an issue for this as an optimization task

cicr99: Issue #383

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: @har777 Done! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: @cicr99 gotcha, so taking that into consideration I think this PR is ready, right?

pkg/hintrunner/zero/zerohint_memcpy_test.go
danielcdz: @TAdev0 @har777 Look this [comment](https://github.com/NethermindEth/cairo-vm-go/pull/359/#discussion_r1588858507) from @cicr99

danielcdz: taking that into consideration adding a test where `"n": uint64(0)` is not going to retrieve the desired value in this moment.

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: @TAdev0 it says this: `invalid operation: n > 0 (operator > not defined on interface)`

pkg/hintrunner/zero/zerohint_memcpy.go
cicr99: @danielcdz atm it will be best to work with felt here, as the variable you are using is saved as a felt in a previous hint. You'll need to modify your code for that. For felts we use the `Element` type from https://github.com/Consensys/gnark-crypto/tree/master/ecc/stark-curve/fp

danielcdz: @cicr99 using felt now!

TAdev0:
@danielcdz  you have conflicts to resolve, can you address this ? thank you!

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -3,9 +3,59 @@ package zero
...
+			//> n -= 1
+			//> ids.continue_copying = 1 if n > 0 else 0
+
+			//> n-=1
+			n, err := ctx.ScopeManager.GetVariableValue("n")
+			if err != nil {
+				return err
+			}
+
+			felt := new(f.Element)
```
MaksymMalicki, line 26: nitpick: this can be a oneliner ```n.Sub(n, &utils.FeltOne)```. You don't need felt variable then and can safely use ```n```

pkg/hintrunner/zero/zerohint_memcpy_test.go
```diff
@@ -23,5 +23,43 @@ func TestZeroHintMemcpy(t *testing.T) {
...
+						"n": &utils.FeltOne,
+					})
+					if err != nil {
+						t.Fatal(err)
+					}
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newMemcpyContinueCopyingHint(ctx.operanders["continue_copying"])
+				},
+				check: varValueEquals("continue_copying", &utils.FeltZero),
```
MaksymMalicki, line 43: can you also check the value of n in the scope?

danielcdz: done! ✅

pkg/hintrunner/zero/zerohint_memcpy.go
danielcdz: Hey, @MaksymMalicki when trying this I found that is not possible 😢 when we try to do `n. Sub` `n` is of type `any`, and doesn't know the `.Sub` method, also I tried doing type casting but the operations didn't work as expected

TAdev0 (changes requested):
@danielcdz  can you move your code in the `zerohint_others.go` and `zerohint_others_test.go` files?

danielcdz:
> gofmt -w .

done!

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -3,9 +3,59 @@ package zero
...
+			//> n -= 1
+			//> ids.continue_copying = 1 if n > 0 else 0
+
+			//> n-=1
+			n, err := ctx.ScopeManager.GetVariableValue("n")
+			if err != nil {
+				return err
+			}
+
+			felt := new(f.Element)
```
MaksymMalicki, line 26: could you rename it to something like ```newN``` or ```updatedN```?

pkg/hintrunner/zero/zerohint_others_test.go
```diff
@@ -4,11 +4,56 @@ import (
...
 
 	runHinterTests(t, map[string][]hintTestCase{
+		"MemcPyContinueCopying": {
+			{
+				operanders: []*hintOperander{
+					{Name: "continue_copying", Kind: uninitialized},
+				},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{})
+					err := ctx.ScopeManager.AssignVariables(map[string]any{
```
TAdev0, line 20: you can do this in one line, entering the scope with a variable inside it, this will allow to remove `AssignVariables` function call

TAdev0 (commented):
@danielcdz  left a minor comment for the test, agree with @MaksymMalicki  for the name of the variable, otherwise LGTM

pkg/hintrunner/zero/zerohint_others.go
danielcdz: sure!

pkg/hintrunner/zero/zerohint_others_test.go
danielcdz: for sure!

danielcdz: done! ✅

pkg/hintrunner/zero/zerohint_others.go
danielcdz: done! ✅

danielcdz:
@MaksymMalicki all good here? 💯

har777 (approved):

cicr99 (approved):

PR #361, opened 2024-04-21
Cairo0 unsafe keccak hint

MaksymMalicki:


pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -81,3 +84,104 @@ func createCairoKeccakFinalizeHinter(resolver hintReferenceResolver) (hinter.Hin
...
+					return fmt.Errorf("word %v is out range 0 <= word < 2 ** %d", &word, 8*nBytes)
+				}
+				wordBytes := word.Bytes20()
+				keccakInput = append(keccakInput, wordBytes[20-int(nBytes):]...)
+				dataPtrCopy, err = dataPtrCopy.AddOffset(1)
+				if err != nil {
+					return err
+				}
+			}
+			hash := sha3.NewLegacyKeccak256()
```
MaksymMalicki, line 144: this is temporary, do you guys recomand any any lib for keccak?

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -81,3 +84,104 @@ func createCairoKeccakFinalizeHinter(resolver hintReferenceResolver) (hinter.Hin
...
+			}
+			dataPtrCopy := *dataPtr
+			keccakInput := make([]byte, 0)
+			for i := uint64(0); i < lengthVal; i += 16 {
+				wordFelt, err := vm.Memory.ReadAsElement(dataPtrCopy.SegmentIndex, dataPtrCopy.Offset)
+				if err != nil {
+					return err
+				}
+				word := uint256.Int(wordFelt.Bits())
+				nBytes := lengthVal - i
```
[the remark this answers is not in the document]
MaksymMalicki: I can't think of any case in which it does, as the loop condition prevents i from being greater that lengthValue. I have modified the tests to prove this.

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -81,3 +84,104 @@ func createCairoKeccakFinalizeHinter(resolver hintReferenceResolver) (hinter.Hin
...
+			//>		keccak_input += word.to_bytes(n_bytes, 'big')
+			//>	hashed = keccak(keccak_input)
+			//>	ids.high = int.from_bytes(hashed[:16], 'big')
+			//>	ids.low = int.from_bytes(hashed[16:32], 'big')
+
+			lengthVal, err := hinter.ResolveAsUint64(vm, length)
+			if err != nil {
+				return err
+			}
+			keccakMaxSize, err := ctx.ScopeManager.GetVariableValueAsUint64("__keccak_max_size")
```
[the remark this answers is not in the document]
MaksymMalicki: Well since we are not handling that right now I'm leaving it as it is. I think I will open new issue for handling globals

MaksymMalicki:
Currently the NewLegacyKeccak256() from the https://pkg.go.dev/golang.org/x/crypto/sha3 is used, which is also used in the [go-ethereum crypto package](https://github.com/ethereum/go-ethereum/blob/master/crypto/crypto.go). Alternatively, we could use [this package](https://pkg.go.dev/leb.io/hashland/keccak#section-readme), which is 10x faster on the benchmarks, than the package used in lambdaclass golang-vm.

TAdev0:
@cicr99  @rodrigo-pino  what do you think? Why not using a faster implementation ?

MaksymMalicki:
Actually after some research, the benchmarks aren't that better than the official goland lib for sha3 and it's not maintaned, so I think we should stick to what we are using currently

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -68,6 +69,114 @@ func createCairoKeccakFinalizeHinter(resolver hintReferenceResolver) (hinter.Hin
...
```
TAdev0, line 103: keccak_max_size is hardcoded to `2**20` in the python VM. Should we do this as well, and remove the value in scope? It will never be in the scope we dont add it manually so i guess we should remove that

TAdev0: lambda one just copy pasted the python one. It seems like the current impl will just `return err` when running cairo file that use keccak because `keccakMaxSize` wont be in the scope

MaksymMalicki: Yea, that sounds legit. This is wrongly implemented still, even if we use the const. Good catch!

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -68,6 +69,114 @@ func createCairoKeccakFinalizeHinter(resolver hintReferenceResolver) (hinter.Hin
...
+			if err != nil {
+				return err
+			}
+			if lengthVal > keccakMaxSize {
+				return fmt.Errorf("unsafe_keccak() can only be used with length<=%d.\n Got: length=%d.", keccakMaxSize, lengthVal)
+			}
```
har777, line 109: Put this check in `if err == nil` instead? Given its a conditional check.

MaksymMalicki: Technically you are right, but I assigned an arbitrary value of 2**20 to the keccakMaxSize (may be moved to globals in the future)

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -68,6 +69,111 @@ func createCairoKeccakFinalizeHinter(resolver hintReferenceResolver) (hinter.Hin
...
+			for i := uint64(0); i < lengthVal; i += 16 {
+				wordFelt, err := vm.Memory.ReadAsElement(dataPtr.SegmentIndex, dataPtr.Offset)
+				if err != nil {
+					return err
+				}
+				word := uint256.Int(wordFelt.Bits())
+				nBytes := lengthVal - i
+				if nBytes > 16 {
+					nBytes = 16
+				}
```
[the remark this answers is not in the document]
MaksymMalicki: Good idea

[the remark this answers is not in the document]
MaksymMalicki: Although the Min function from math is meant to work with floats, so it requires a lot of typecasts. I added generic Min function to the math utils, which should do the trick and I used a similar function in other PRs.

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -68,6 +69,111 @@ func createCairoKeccakFinalizeHinter(resolver hintReferenceResolver) (hinter.Hin
 	return newCairoKeccakFinalizeHint(keccakPtrEnd), nil
 }
 
+// UnsafeKeccak computes keccak hash of the data in memory without validity enforcement and writes the result in the `low` and `high` memory cells
+//
+// `newUnsafeKeccakHint` takes 4 operanders as arguments
+//   - `data` is the address in memory to the base of the data array to hash is stored. Each word in the array is 16 bytes long, except the last one
```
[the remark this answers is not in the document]
MaksymMalicki: Done 🫡

cicr99 (approved):

TAdev0 (approved):
lgtm!

PR #366, opened 2024-04-23
SquashDictInnerAssertLenKeys

### LINKED ISSUE #302: SquashDictInnerAssertLenKeys
Asserts the length of the keys array during the squashing process.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/squash_dict.cairo#L216

depends on #245

jbmkahdeksan:
resolves #302 

Added Squash Dict Inner Assert Len Keys

### DIFF
--- a/pkg/hintrunner/zero/zerohint.go
+++ b/pkg/hintrunner/zero/zerohint.go
@@ -146,6 +146,9 @@ func GetHintFromCode(program *zero.ZeroProgram, rawHint zero.Hint, hintPC uint64
 		return createUsortEnterScopeHinter()
 	case usortVerifyCode:
 		return createUsortVerifyHinter(resolver)
+		// Dictionaries hints
+	case squashDictInnerAssertLenKeys:
+		return createSquashDictInnerAssertLenKeysHinter()
 		// Other hints
 	case allocSegmentCode:
 		return createAllocSegmentHinter(resolver)
--- a/pkg/hintrunner/zero/zerohint_dictionaries.go
+++ b/pkg/hintrunner/zero/zerohint_dictionaries.go
@@ -1,0 +1,31 @@
+package zero
+
+import (
+	"fmt"
+
+	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
+	f "github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+)
+
+func newSquashDictInnerAssertLenKeysHint() hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "SquashDictInnerAssertKeys",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			//> assert len(keys) == 0
+			keys_, err := ctx.ScopeManager.GetVariableValue("keys")
+			if err != nil {
+				return err
+			}
+			keys := keys_.([]f.Element)
+			if len(keys) != 0 {
+				return fmt.Errorf("assertion `len(keys) == 0` failed")
+			}
+			return nil
+		},
+	}
+}
+
+func createSquashDictInnerAssertLenKeysHinter() (hinter.Hinter, error) {
+	return newSquashDictInnerAssertLenKeysHint(), nil
+}
--- a/pkg/hintrunner/zero/zerohint_dictionaries_test.go
+++ b/pkg/hintrunner/zero/zerohint_dictionaries_test.go
@@ -1,0 +1,35 @@
+package zero
+
+import (
+	"testing"
+
+	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+)
+
+func TestZeroHintDictionaries(t *testing.T) {
+	runHinterTests(t, map[string][]hintTestCase{
+		"SquashDictInnerAssertKeys": {
+			{
+				operanders: []*hintOperander{},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{"keys": []fp.Element{}})
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newSquashDictInnerAssertLenKeysHint()
+				},
+				check: func(t *testing.T, ctx *hintTestContext) {},
+			},
+			{
+				operanders: []*hintOperander{},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{"keys": []fp.Element{*feltUint64(1), *feltUint64(2)}})
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newSquashDictInnerAssertLenKeysHint()
+				},
+				errCheck: errorTextContains("assertion `len(keys) == 0` failed"),
+			},
+		},
+	})
+}
--- a/pkg/hintrunner/zero/hintcode.go
+++ b/pkg/hintrunner/zero/hintcode.go
@@ -84,6 +84,7 @@ const (
 	keccakWriteArgs string = "segments.write_arg(ids.inputs, [ids.low % 2 ** 64, ids.low // 2 ** 64])\nsegments.write_arg(ids.inputs + 2, [ids.high % 2 ** 64, ids.high // 2 ** 64])"
 
 	// ------ Dictionaries hints related code ------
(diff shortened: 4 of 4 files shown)


rodrigo-pino (commented):
Great! Could you add a unit test for this hint?

TAdev0 (changes requested):
@jbmkahdeksan you have a linting issue to resolve.

otherwise lgtm!

pkg/hintrunner/zero/zerohint.go
```diff
@@ -137,6 +137,9 @@ func GetHintFromCode(program *zero.ZeroProgram, rawHint zero.Hint, hintPC uint64
 		// Usort hints
 	case usortEnterScopeCode:
 		return createUsortEnterScopeHinter()
+		// Dicitonaries hints
```
TAdev0, line 140: Dicitonaries -> Dictionaries

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -0,0 +1,24 @@
+package zero
+
+import (
+	"fmt"
+	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
+)
+
```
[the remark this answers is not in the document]
jbmkahdeksan: didn't notice this, thank you! 👍

pkg/hintrunner/zero/zerohint.go
jbmkahdeksan: Oops! 😅

pkg/hintrunner/zero/zerohint_dictionaries_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+
+	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+)
+
+func TestZeroHintDictionaries(t *testing.T) {
+	runHinterTests(t, map[string][]hintTestCase{
+		"SquashDictInnerAssertKeys": {
+			{
```
TAdev0, line 13: I think you can have only one test : 

```
			{
				operanders: []*hintOperander{},
					err := hinter.InitializeSquashedDictionaryManager(ctx)
					if err != nil {
						t.Fatal(err.Error())
					}
				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newSquashDictInnerAssertLenKeysHint()
				},
				check: func(t *testing.T, ctx *hintTestContext) {
				},
			},
```

actually, you dont need anything in `check`, because the check is the hint itself.

The `ctxInit` step will call `InitializeSquashedDictionaryManager`  which creates a `SquashedDictionaryManager` without any key in the list.

Then the hint will do the check that `len == 0`

If you try this : 

```
		"SquashDictInnerAssertKeys": {
			{
				operanders: []*hintOperander{},
					err := hinter.InitializeSquashedDictionaryManager(ctx)
					if err != nil {
						t.Fatal(err.Error())
					}

					ctx.SquashedDictionaryManager.Keys = append(ctx.SquashedDictionaryManager.Keys, fp.Element{})

				},
				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
					return newSquashDictInnerAssertLenKeysHint()
				},
				check: func(t *testing.T, ctx *hintTestContext) {
				},
			},
```

The test will fail because of `ctx.SquashedDictionaryManager.Keys = append(ctx.SquashedDictionaryManager.Keys, fp.Element{})` during the `ctxInit` , and its actually the code of the hint that will fail.

TAdev0: The test consists more of making the initial setup and running the hint than making a check after the hint

TAdev0 (changes requested):

pkg/hintrunner/zero/zerohint_dictionaries_test.go
TAdev0: if there is a way in go to indicate that test should actually have failing code to pass, then you can add a second test where you add some key in the list before calling the hint !

jbmkahdeksan: Got it, thanks a lot!! ✌️
I was actually a little confused with this part and learned a lot from this review!

jbmkahdeksan: new changes were pushed!
thanks a lot @TAdev0 🫡

TAdev0 (approved):

TAdev0:
LGTM ! @rodrigo-pino  @chee-chyuan  @MaksymMalicki  ready for final review before merging

har777:
I'm not really sure about this.

`assert len(keys) == 0` seems to assert that the length of value assigned to scope value `keys` is zero. Don't think directly checking the length of the squashed dict is the same 🤔 IMO you should just fetch the scope value of `keys` as a list and then assert its length is 0. @cicr99 

reference -> `keys` is being set [here](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/squash_dict.cairo#L55)

Also another test where the assertion fails would be ideal.

TAdev0:
@har777  i just checked , indeed it seems that `keys` should be retrieve as a scope variable. but i think that checking the descending list `Keys` which is a field of the `SquashedDictionaryManager` (manager of all squashed dicts) should work as well?  But yeah i think you're right and we should fetch keys in the same way as the python hint

pkg/hintrunner/core/hint.go
```diff
@@ -978,6 +978,11 @@ func (hint *InitSquashData) Execute(vm *VM.VirtualMachine, ctx *hinter.HintRunne
 		return fmt.Errorf("get first key: %w", err)
 	}
 
+	err = ctx.ScopeManager.AssignVariable("keys", ctx.SquashedDictionaryManager.Keys)
+	if err != nil {
+		return err
+	}
```
har777, line 984: Why is this required? I was thinking some other hint will be populating the value of `keys`. We don't need to do anything extra. cc @cicr99

jbmkahdeksan: I don't understand what you mean. Could you please elaborate? Maybe via Telegram...

TAdev0: why would you need to modify the _hint.go_ file?

jbmkahdeksan: I had no idea before, I thought 'keys' had to be obtain it's somewhere in order for this hint to work...

jbmkahdeksan: new changes pushed @TAdev0 @har777 @rodrigo-pino @cicr99

har777:
lgtm!

pkg/hintrunner/zero/zerohint_dictionaries_test.go
```diff
@@ -0,0 +1,39 @@
...
+				operanders: []*hintOperander{},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{"keys": []fp.Element{}})
+					err := hinter.InitializeSquashedDictionaryManager(ctx)
```
TAdev0, line 17: Why is it needed to call `InitializeSquashedDictionaryManager` here? 
If we purely test the hint, then we just need to enter the scope with keys.

If you remove it , test pass. `InitializeSquashedDictionaryManager`   deals with the `SquashedDictionaryManager` which shouldn't be managed manually for this hint anyway as previously discussed.

You dont use it in the second test, i guess it can be removed ? @har777  @cicr99

TAdev0 (changes requested):

pkg/hintrunner/zero/zerohint_dictionaries_test.go
har777: Yep just `hinter.InitializeScopeManager(ctx, map[string]any{"keys": []fp.Element{}})` is enough

jbmkahdeksan: thank you, will fix it!

jbmkahdeksan: sure, you are right!
`InitializeSquashedDictionaryManager` is beyond the hint implementation
I don't need it to test the hint because the hint itself is a granular and independent part from the DictionaryManager and I only need a scope to test the hint.
Is that correct?

jbmkahdeksan: @TAdev0 @har777 new changes were pushed! 
thank you, I value a lot your feedback!

TAdev0 (approved):

cicr99 (approved):

Issue #235, opened 2024-03-06
UsortVerifyMultiplicityBody


cicr99:
Processes each position of a specific value in the sorted output, updating indices for verification.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L92

Depends on issue #245

wugalde19:
@cicr99 
Hi Carmen
Hope you are doing great.
Looking forward to contributing and learning more about the project

If not yet assigned, can I get the chance to work on this issue?

cicr99:
Hey @wugalde19 sure! Make sure to check out this [guide](https://github.com/NethermindEth/cairo-vm-go/blob/main/pkg/hintrunner/zero/README.md) on how to add new hints first. Let us know if you have any questions

wugalde19:
> Hey @wugalde19 sure! Make sure to check out this [guide](https://github.com/NethermindEth/cairo-vm-go/blob/main/pkg/hintrunner/zero/README.md) on how to add new hints first. Let us know if you have any questions

Hi @cicr99 
Sure I'll take a look and I'll reach out as soon as I've gone through that doc and have set up the project
Thanks for the opportunity

PR #368, opened 2024-04-24
Implement UsortVerifyMultiplicityBody hint

wugalde19:
Resolves https://github.com/NethermindEth/cairo-vm-go/issues/235

### DIFF
--- a/pkg/hintrunner/zero/zerohint_usort.go
+++ b/pkg/hintrunner/zero/zerohint_usort.go
@@ -125,7 +125,11 @@ func newUsortVerifyMultiplicityBodyHint(nextItemIndex hinter.ResOperander) hinte
 				return err
 			}
 
-			current_pos := utils.Pop(&positions)
+			current_pos, err := utils.Pop(&positions)
+			if err != nil {
+				return err
+			}
+
 			last_pos := current_pos + 1
 
 			// Calculate `next_item_index` memory value
--- a/pkg/hintrunner/zero/zerohint_usort.go
+++ b/pkg/hintrunner/zero/zerohint_usort.go
@@ -127,28 +127,28 @@ func newUsortVerifyMultiplicityBodyHint(nextItemIndex hinter.ResOperander) hinte
 				return fmt.Errorf("cannot cast positionsInterface to []int")
 			}
 
-			new_current_pos, err := utils.Pop(&positions)
+			newCurrentPos, err := utils.Pop(&positions)
 			if err != nil {
 				return err
 			}
 
-			current_pos, err := ctx.ScopeManager.GetVariableValue("current_pos")
+			currentPos, err := ctx.ScopeManager.GetVariableValue("current_pos")
 			if err != nil {
 				return err
 			}
 
-			current_pos_int, ok := current_pos.(int)
+			currentPosInt, ok := currentPos.(int)
 			if !ok {
 				return fmt.Errorf("cannot cast current_pos to int")
 			}
 
-			last_pos, err := ctx.ScopeManager.GetVariableValue("last_pos")
+			lastPos, err := ctx.ScopeManager.GetVariableValue("last_pos")
 			if err != nil {
 				return err
 			}
 
 			// Calculate `next_item_index` memory value
-			newNextItemIndexValue := current_pos_int - last_pos.(int)
+			newNextItemIndexValue := currentPosInt - lastPos.(int)
 			newNextItemIndexMemoryValue := memory.MemoryValueFromInt(newNextItemIndexValue)
 
 			// Save `next_item_index` value in address
@@ -164,8 +164,8 @@ func newUsortVerifyMultiplicityBodyHint(nextItemIndex hinter.ResOperander) hinte
 
 			// Save `current_pos` and `last_pos` values in scope variables
 			return ctx.ScopeManager.AssignVariables(map[string]any{
-				"current_pos": new_current_pos,
-				"last_pos":    current_pos_int + 1,
+				"current_pos": newCurrentPos,
+				"last_pos":    currentPosInt + 1,
 			})
 		},
 	}
--- a/pkg/hintrunner/zero/zerohint_usort.go
+++ b/pkg/hintrunner/zero/zerohint_usort.go
@@ -136,7 +136,7 @@ func newUsortVerifyMultiplicityBodyHint(nextItemIndex hinter.ResOperander) hinte
 			}
 
 			// Calculate `next_item_index` memory value
-			newNextItemIndexValue := current_pos - last_pos
+			newNextItemIndexValue := current_pos - last_pos.(uint64)
 			newNextItemIndexMemoryValue := memory.MemoryValueFromUint(newNextItemIndexValue)
 
 			// Save `next_item_index` value in address
@@ -151,12 +151,10 @@ func newUsortVerifyMultiplicityBodyHint(nextItemIndex hinter.ResOperander) hinte
 			}
 
 			// Save `current_pos` and `last_pos` values in scope variables
-			err = ctx.ScopeManager.AssignVariables(map[string]any{
+			return ctx.ScopeManager.AssignVariables(map[string]any{
 				"current_pos": current_pos,
 				"last_pos":    current_pos + 1,
 			})
-
-			return err
 		},
 	}
 }
--- a/pkg/hintrunner/zero/zerohint_usort.go
+++ b/pkg/hintrunner/zero/zerohint_usort.go
@@ -1,8 +1,6 @@
 package zero
 
 import (
-	"fmt"
-
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
 	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
@@ -119,12 +117,12 @@ func newUsortVerifyMultiplicityBodyHint(nextItemIndex hinter.ResOperander) hinte
 
 			positionsInterface, err := ctx.ScopeManager.GetVariableValue("positions")
 			if err != nil {
-				return fmt.Errorf("getting positions from scope failed: %w", err)
+				return err
 			}
 
 			positions, ok := positionsInterface.([]uint64)
 			if !ok {
-				return fmt.Errorf("casting positions into an array of uint64 failed: %w", err)
+				return err
 			}
 
 			current_pos := utils.Pop(&positions)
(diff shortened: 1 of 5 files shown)


pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -21,5 +22,79 @@ func TestZeroHintUsort(t *testing.T) {
 				check: varValueInScopeEquals("__usort_max_size", feltUint64(1)),
 			},
 		},
+		"UsortVerifyMultiplicityBody": {
+			{
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					// Having empty scopes in scope manager will cause an error
```
MaksymMalicki, line 28: There is no need to test here if the scope manager is empty, as the function is always invoked in the scope of usort() function

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -21,5 +22,79 @@ func TestZeroHintUsort(t *testing.T) {
...
+				},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{
+						"positions": []uint64{1, 2, 3},
+					})
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newUsortVerifyMultiplicityBodyHint(ctx.operanders["next_item_index"])
+				},
+				check: varValueInScopeEquals("current_pos", feltUint64(3)),
```
MaksymMalicki, line 66: Maybe also add check if the last_pos value is as expected in scope

MaksymMalicki: You can then remove the next test. There are utilities like ```allVarValueInScopeEquals``` for that

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -26,3 +30,63 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
+				return fmt.Errorf("getting positions from scope failed: %w", err)
+			}
+
+			positions, ok := positionsInterface.([]uint64)
+			if !ok {
+				return fmt.Errorf("casting positions into an array of uint64 failed: %w", err)
+			}
+
+			current_pos := utils.Pop(&positions)
+			last_pos := current_pos + 1
```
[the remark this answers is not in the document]
wugalde19: You're right
I was confused because it didn't make sense to me that we were using a variable that was not declared but now that I understand more the code I noticed that `last_pos` is present in the scope and we just need to get the value from there

Thanks for pointing this out @MaksymMalicki

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -21,5 +23,64 @@ func TestZeroHintUsort(t *testing.T) {
 				check: varValueInScopeEquals("__usort_max_size", feltUint64(1)),
 			},
 		},
+		"UsortVerifyMultiplicityBody": {
+			{
```
MaksymMalicki, line 27: you can add some very short comments on what this test tests

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -21,5 +23,64 @@ func TestZeroHintUsort(t *testing.T) {
...
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{
+						"positions": []uint64{6, 4, 2},
+						"last_pos":  0,
+					})
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newUsortVerifyMultiplicityBodyHint(ctx.operanders["next_item_index"])
+				},
+				check: allVarValueInScopeEquals(map[string]any{
```
MaksymMalicki, line 62: I think you can merge this test and the bottom one into one. You can call two check functions in one test, by doing:
```
check: func(t *testing.T, ctx *hintTestContext) {					varAddrResolvedValueEquals("output_length", feltUint64(3))(t, ctx)
	allVarValueInScopeEquals()(t, ctx)
	allVarValueEquals()(t, ctx)
},
```

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -21,5 +23,64 @@ func TestZeroHintUsort(t *testing.T) {
...
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{
+						"positions": []uint64{6, 4, 2},
+						"last_pos":  1,
+					})
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newUsortVerifyMultiplicityBodyHint(ctx.operanders["next_item_index"])
+				},
+				check: allVarValueEquals(map[string]*fp.Element{
```
MaksymMalicki, line 80: when you merge the tests, you can use here ```varValueEquals``` instead of ```allVarValueEquals```

pkg/utils/arrays.go
```diff
@@ -1,7 +1,22 @@
...
+
 func Reverse[T any](a []T) {
 	for i, j := 0, len(a)-1; i < j; i, j = i+1, j-1 {
 		a[i], a[j] = a[j], a[i]
 	}
 }
+
+func Pop[T any](a *[]T) (T, error) {
+	if len(*a) == 0 {
+		var zeroValue T
```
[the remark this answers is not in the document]
wugalde19: @MaksymMalicki I had it like 
image

But it is giving me a warning
If you want I can just ignore it
image

I was also considering make the return type a pointer `*T` but not sure if that's the best approach

What do you think?

pkg/hintrunner/zero/zerohint_usort_test.go
```diff
@@ -21,5 +22,54 @@ func TestZeroHintUsort(t *testing.T) {
...
+				},
+				errCheck: func(t *testing.T, ctx *hintTestContext, err error) {
+					require.NotNil(t, err)
+				},
+			},
+			// Tests when the positions interface cannot be casted to an array of uint64.
+			{
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					ctx.ScopeManager.EnterScope(map[string]any{
+						"positions": []string{"6", "4", "2"},
```
har777, line 39: the positions seem to be written as:
```
positions_dict = defaultdict(list)
for i in range(input_len):
    val = memory[input_ptr + i]
    positions_dict[val].append(i)
```

So no need to test them with strings. uint64 should suffice.

pkg/utils/arrays.go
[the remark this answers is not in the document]
MaksymMalicki: No no, its okey, I was just wondering haha 😛. Returning pointer is not a good idea, as it will cause heap allocations and performance issues. The current solution looks great

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -26,3 +28,66 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
+				return err
+			}
+
+			err = vm.Memory.WriteToAddress(&addrNextItemIndex, &newNextItemIndexMemoryValue)
+			if err != nil {
+				return err
+			}
+
+			// Save `current_pos` and `last_pos` values in scope variables
+			err = ctx.ScopeManager.AssignVariables(map[string]any{
```
MaksymMalicki, line 76: Sorry, in previous comment I pointed that you can just return err. Actually you can just return what this function returns haha 😅

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -26,3 +28,66 @@ func newUsortEnterScopeHint() hinter.Hinter {
...
+			if !ok {
+				return err
+			}
+
+			current_pos, err := utils.Pop(&positions)
+			if err != nil {
+				return err
+			}
+
+			last_pos, err := ctx.ScopeManager.GetVariableValue("last_pos")
```
har777, line 55: type case to `.(uint64)`

pkg/utils/arrays.go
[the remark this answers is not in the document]
wugalde19: @MaksymMalicki ok cool
Thanks for the explanation

MaksymMalicki:
lgtm

MaksymMalicki:
For some reason the tests are not passing, make sure to run ```make unit``` and resolve the problems. Thanks!

wugalde19:
> For some reason the tests are not passing, make sure to run `make unit` and resolve the problems. Thanks!

@MaksymMalicki My apologies
Something was wrong with the environment
I dockerized it to make sure it does not happen again
image

I did fix the test and made a small change reading the `current_pos` from the scope instead of using the one from `positions` to calculate the value of `next_item_index` and `last_pos`

wugalde19:
@MaksymMalicki I just fixed conflicts after rebasing with main branch

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -107,3 +108,74 @@ func createUsortVerifyHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+			if err != nil {
+				return err
+			}
+
+			positions, ok := positionsInterface.([]int)
+			if !ok {
+				return fmt.Errorf("cannot cast positionsInterface to []int")
+			}
+
+			new_current_pos, err := utils.Pop(&positions)
```
cicr99, line 130: Could you use camelCase for variable names on the go side?

pkg/hintrunner/zero/zerohint_usort.go
```diff
@@ -107,3 +108,74 @@ func createUsortVerifyHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+			//> current_pos = positions.pop()
+			//> ids.next_item_index = current_pos - last_pos
+			//> last_pos = current_pos + 1
+
+			positionsInterface, err := ctx.ScopeManager.GetVariableValue("positions")
+			if err != nil {
+				return err
+			}
+
+			positions, ok := positionsInterface.([]int)
```
cicr99, line 125: It's better to specify int64 for this

pkg/hintrunner/zero/zerohint_usort.go
cicr99: the idea is to do the typecasting here for `last_pos` the same way you did for `current_pos`

cicr99 (commented):
Just these minor details left. The PR looks great!

pkg/hintrunner/zero/zerohint_usort.go
wugalde19: @cicr99 I tried but it was throwing an error, so I had to use `uint64` instead

image

then for `currentPos` and `lastPos` values, I was also getting an error when trying to cast to `int64` so I keep them as `int`

image

wugalde19: Understood and done

Issue #347, opened 2024-04-10
Create a structure for Scope value types instead of using interface


cicr99:
Currently the ScopeManager contains a `map[string]any` to represent a scope, where the `key` is the name of the variables present in the scope and the `value` would be the value of such variable. `any` is being used because we cannot know the type of this value beforehand. As this is used in python code you could get a felt or a bigint, or a mapping, etc.
However, the use of `any` is not very efficient, so the proposed solution to improve this is to create a struct `ScopeValue` which will hold the value of the variables. It could look like the following:
```
type ScopeValue struct {
	FeltValue   *fp.Element
        BigIntValue *Big.Int
}
```
We should add all possible types in there; it could be done on demand, as we go implementing new hints.
Another idea could be to create a `ScopeValueType` enum for all these possible types and add a field `Type ScopeValueType` to the previous struct to know the type of the value without having to check which member is not `nil`

fishonamos:
Hi @cicr99 . Can this be assigned to me?

cicr99:
Hey @fishonamos I've just assigned you to #345 as you also requested . I would like you to finish that one before assigning you to another if that's ok

danielcdz:
Hello, @cicr99! I can help with this, can you assign it to me?

fishonamos:
Hi @cicr99 . I have sent a pr for #345

fmmesen:
Hello @cicr99 . I'd like to contribute to this issue, could you please assign it to me?

rodrigo-pino:
Hey @fmmesen it's yours!

@danielcdz sorry for replying so late, if you are still interested you're free to pick any open issue you like!

fmmesen:
Thanks @rodrigo-pino ! Is there a channel where I can get some quick help? I can't run the cairo-compile command

rodrigo-pino:
> Thanks @rodrigo-pino ! Is there a communication channel for the project?
> 

Hey yes, there is a Telegram group, you will be able to find the link on the Only Dust Hack channel

fmmesen:
Hello! @rodrigo-pino @cicr99 PR #369 is ready for review. Thanks!

danielcdz:
@rodrigo-pino @cicr99 it's me again sorry for the delay, but if this is something I can help with, I'll be happy to work, this will be my third contribution to the project!

danielcdz:
[...]

PR #369, opened 2024-04-24
Create a structure for Scope value types instead of using interface #347

fmmesen:
All unit tests have been run successfully after changes.

### DIFF
--- a/pkg/hintrunner/hinter/scope.go
+++ b/pkg/hintrunner/hinter/scope.go
@@ -3,6 +3,8 @@ package hinter
 import (
 	"fmt"
 	"math/big"
+
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
 // ScopeManager handles all operations regarding scopes:
@@ -10,19 +12,37 @@ import (
 // - Exiting current scope
 // - Variable declaration and assignment inside a certain scope
 // - Accessing variable values
-type ScopeManager struct {
-	scopes []map[string]any
+
+type ScopeValue struct {
+	Type        int
+	FeltValue   *fp.Element
+	BigIntValue *big.Int
 }
 
-func InitializeScopeManager(ctx *HintRunnerContext, scope map[string]any) {
+type ScopeMap map[string]ScopeValue
+
+type ScopeValueType int
+
+const (
+	TypeNone = iota
+	TypeFeltValue
+	TypeBigInt
+)
+
+type ScopeManager struct {
+	scopes []ScopeMap
+}
+
+func InitializeScopeManager(ctx *HintRunnerContext, scope ScopeMap) {
 	if ctx.ScopeManager.scopes == nil {
 		ctx.ScopeManager = *NewScopeManager(scope)
 	}
 }
 
-func NewScopeManager(globals map[string]any) *ScopeManager {
+func NewScopeManager(globals ScopeMap) *ScopeManager {
+
 	return &ScopeManager{
-		scopes: []map[string]any{
+		scopes: []ScopeMap{
 			// One scope needed (current execution scope)
 			globals,
 		},
@@ -30,10 +50,11 @@ func NewScopeManager(globals map[string]any) *ScopeManager {
 }
 
 func DefaultNewScopeManager() *ScopeManager {
-	return NewScopeManager(make(map[string]any))
+	return NewScopeManager(make(ScopeMap))
 }
 
-func (sm *ScopeManager) EnterScope(newScope map[string]any) {
+func (sm *ScopeManager) EnterScope(newScope ScopeMap) {
+
 	sm.scopes = append(sm.scopes, newScope)
 }
 
@@ -45,7 +66,7 @@ func (sm *ScopeManager) ExitScope() error {
 	return nil
 }
 
-func (sm *ScopeManager) AssignVariable(name string, value any) error {
+func (sm *ScopeManager) AssignVariable(name string, value ScopeValue) error {
 	scope, err := sm.getCurrentScope()
 	if err != nil {
 		return err
@@ -55,7 +76,7 @@ func (sm *ScopeManager) AssignVariable(name string, value any) error {
 	return nil
 }
 
-func (sm *ScopeManager) AssignVariables(values map[string]any) error {
+func (sm *ScopeManager) AssignVariables(values ScopeMap) error {
 	for name, value := range values {
 		err := sm.AssignVariable(name, value)
 		if err != nil {
@@ -65,14 +86,14 @@ func (sm *ScopeManager) AssignVariables(values map[string]any) error {
 	return nil
 }
 
-func (sm *ScopeManager) GetVariableValue(name string) (any, error) {
+func (sm *ScopeManager) GetVariableValue(name string) (*ScopeValue, error) {
 	scope, err := sm.getCurrentScope()
 	if err != nil {
 		return nil, err
 	}
 
 	if value, ok := (*scope)[name]; ok {
-		return value, nil
+		return &value, nil
 	}
 
 	return nil, fmt.Errorf("variable %s not found in current scope", name)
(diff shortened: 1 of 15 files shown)


pkg/hintrunner/hinter/scope.go
```diff
@@ -3,37 +3,58 @@ package hinter
...
 
 // ScopeManager handles all operations regarding scopes:
 // - Creating a new scope
 // - Exiting current scope
 // - Variable declaration and assignment inside a certain scope
 // - Accessing variable values
+
+type ScopeValue struct {
+	Type        int
+	FeltValue   *fp.Element
```
MaksymMalicki, line 18: Do we need Value suffix in the type name?

pkg/hintrunner/hinter/scope.go
```diff
@@ -84,17 +105,25 @@ func (sm *ScopeManager) GetVariableValueAsBigInt(name string) (*big.Int, error)
 		return nil, err
 	}
 
-	valueBig, ok := value.(*big.Int)
-	if !ok {
-		return nil, fmt.Errorf("value: %s is not a *big.Int", value)
-	}
+	// valueBig, ok := value.(*big.Int)
```
MaksymMalicki, line 108: It would be cool if we removed the comments, or add some comment if this is still needed in the future (draft PR)

pkg/hintrunner/hinter/scope.go
```diff
@@ -65,14 +86,14 @@ func (sm *ScopeManager) AssignVariables(values map[string]any) error {
 	return nil
 }
 
-func (sm *ScopeManager) GetVariableValue(name string) (any, error) {
+func (sm *ScopeManager) GetVariableValue(name string) (*ScopeValue, error) {
```
MaksymMalicki, line 89: The struct you proposed looks like this:
```type ScopeMap map[string]ScopeValue```
If you save the pointer to the ScopeValue in a ```value``` variable and return the pointer to the ```value```, that might cause dangling pointer and unneccesary heap allocation. It would be better to just return the value.

File: pkg/hintrunner/hinter/scope.go [hunk 4]
```diff
@@ -84,17 +105,25 @@ func (sm *ScopeManager) GetVariableValueAsBigInt(name string) (*big.Int, error)
...
-func (sm *ScopeManager) getCurrentScope() (*map[string]any, error) {
+func (sm *ScopeManager) getCurrentScope() (*ScopeMap, error) {
 	if len(sm.scopes) == 0 {
 		return nil, fmt.Errorf("expected at least one existing scope")
 	}
 	return &sm.scopes[len(sm.scopes)-1], nil
 }
+
+func BigIntScopeValue(value *big.Int) *ScopeValue {
+	return &ScopeValue{Type: TypeBigInt, BigIntValue: value}
```
MaksymMalicki, line 124: To minimize allocations on the heap, try not to return pointers to objects allocated locally in function whenever possible.
So in this case it would be better to return a value instead of pointer

pkg/hintrunner/hinter/scope.go
```diff
@@ -84,17 +105,25 @@ func (sm *ScopeManager) GetVariableValueAsBigInt(name string) (*big.Int, error)
...
+}
+
+func FeltValueScopeValue(value *fp.Element) *ScopeValue {
+	return &ScopeValue{Type: TypeFeltValue, FeltValue: value}
```
MaksymMalicki, line 128: same here

pkg/hintrunner/hinter/scope.go
fmmesen: I already create other function to pass the BigInt and Felt value directly

pkg/hintrunner/hinter/scope.go
fmmesen: Already changed it.

pkg/hintrunner/hinter/scope.go
fmmesen: Done

pkg/hintrunner/hinter/scope.go
fmmesen: Done

pkg/hintrunner/hinter/scope.go
fmmesen: Done

fmmesen:
@MaksymMalicki Already made the changes even for new branches. All tests are running.
Screenshot 2024-04-25 at 05 41 07

pkg/hintrunner/hinter/scope.go
MaksymMalicki: This function is ok, but the case is not to return a pointer created inside this function, but to return ScopeValue as value. This should reduce the heap allocations.
Do you mean that this function is no longer used? If so, then it can be removed. Let me know 😁

MaksymMalicki:
One thing is before pushing to this branch, you can run ```golangci-lint run``` and ```gofmt -w .``` to fix the linter issues. Thanks!

pkg/hintrunner/hinter/scope.go
MaksymMalicki: nvm got it haha

pkg/hintrunner/hinter/scope.go
```diff
@@ -65,36 +86,48 @@ func (sm *ScopeManager) AssignVariables(values map[string]any) error {
...
-func (sm *ScopeManager) GetVariableValue(name string) (any, error) {
+func (sm *ScopeManager) GetScopeValue(name string) (*ScopeValue, error) {
 	scope, err := sm.getCurrentScope()
 	if err != nil {
 		return nil, err
 	}
 
 	if value, ok := (*scope)[name]; ok {
-		return value, nil
+		return &value, nil
```
MaksymMalicki, line 96: Returning pointer to ScopeValue here might cause dangling pointer. Would be better to return the value itself.

pkg/hintrunner/hinter/scope.go
```diff
@@ -65,36 +86,48 @@ func (sm *ScopeManager) AssignVariables(values map[string]any) error {
...
 	}
 
 	return nil, fmt.Errorf("variable %s not found in current scope", name)
 }
 
-func (sm *ScopeManager) GetVariableValueAsBigInt(name string) (*big.Int, error) {
-	value, err := sm.GetVariableValue(name)
+func (sm *ScopeManager) GetVariableValueAsFelt(name string) (*fp.Element, error) {
```
MaksymMalicki, line 102: the same here, return value to reduce heap allocations.

pkg/hintrunner/hinter/scope.go
```diff
@@ -65,36 +86,48 @@ func (sm *ScopeManager) AssignVariables(values map[string]any) error {
...
 		return nil, err
 	}
 
-	valueBig, ok := value.(*big.Int)
-	if !ok {
-		return nil, fmt.Errorf("value: %s is not a *big.Int", value)
+	return &value.Felt, nil
+}
+
+func (sm *ScopeManager) GetVariableValueAsBigInt(name string) (*big.Int, error) {
```
MaksymMalicki, line 111: same here

fmmesen: Hey, what could I do here? I cant return nil if big.Int isn't a pointer

fmmesen: Everything is working now

MaksymMalicki:
Lgtm for know 😁 @cicr99 can you take a look?

pkg/hintrunner/hinter/scope.go
```diff
@@ -65,36 +86,48 @@ func (sm *ScopeManager) AssignVariables(values map[string]any) error {
...
 
-func (sm *ScopeManager) getCurrentScope() (*map[string]any, error) {
+func (sm *ScopeManager) getCurrentScope() (*ScopeMap, error) {
 	if len(sm.scopes) == 0 {
 		return nil, fmt.Errorf("expected at least one existing scope")
 	}
 	return &sm.scopes[len(sm.scopes)-1], nil
 }
+
+func SetBigIntScopeValue(value big.Int) ScopeValue {
```
har777, line 127: I feel like the name can be better. `SetBigIntScopeValue` sounds like its adding a value to scope.

fmmesen: Well, it's adding the big.Int/felt value to the ScopeValue

map[string]ScopeValue

image

TAdev0:
@fmmesen  any update on this?

cicr99:
Thanks for your work @fmmesen . Unfortunately, this solution is no longer applicable for the problem we want to solve, as described [here](https://github.com/NethermindEth/cairo-vm-go/issues/347#issuecomment-2134697873), so I'll be closing this PR.

PR #372, opened 2024-04-25
feat(MemcpyEnterScope): enters a new scope for the memory copy operaton with a specified length

### LINKED ISSUE #250: MemcpyEnterScope
Enters a new scope for the memory copy operation with a specified length.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/memcpy.cairo#L12

Depends on #245

Alvarodb:
resolves #250


pkg/hintrunner/zero/zerohint.go
```diff
@@ -146,7 +146,7 @@ func GetHintFromCode(program *zero.ZeroProgram, rawHint zero.Hint, hintPC uint64
 	case allocSegmentCode:
 		return createAllocSegmentHinter(resolver)
 	case vmEnterScopeCode:
-		return createVMEnterScopeHinter(resolver)
+		return createMemcpyEnterScopeHinter(resolver)
```
MaksymMalicki, line 149: this is the correct of doing it, but you need to add the hint code as separate code. vmEnterScopeCode sure enters the scope, but it's not the exact code you want. So just add to the codes:
```memcpyEnterScope string = "vm_enter_scope({'n': ids.len})"```
and replace vmEnterScopeCode with the memcpyEnterScope in this case.

File: pkg/hintrunner/zero/zerohint_memcpy.go [hunk 2]
```diff
@@ -10,14 +10,23 @@ func createAllocSegmentHinter(resolver hintReferenceResolver) (hinter.Hinter, er
...
 
-func createVMEnterScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
+func newMemcpyEnterScopeHinter(len hinter.ResOperander) hinter.Hinter {
 	return &GenericZeroHinter{
-		Name: "VMEnterScope",
+		Name: "MemcpyEnterScope",
 		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
-			ctx.ScopeManager.EnterScope(make(map[string]any))
+			//>  vm_enter_scope({'n': ids.len})
+			ctx.ScopeManager.EnterScope(map[string]any{"n": len})
```
MaksymMalicki, line 18: You are adding ResOperander to the scope. This is only the reference, we need to get the exact value that it holds. As you can see in the memcpy() method, where this hint is used, the len is a Felt, so you have to resolve it as felt and check the error, using ```hinter.ResolveAsFelt(vm, len)```

MaksymMalicki:
Thank you for you contribution! If you resolve the the len as Felt and load it to the scope, you can try and write some unit tests, which check the insides of the scope. Write on Telegram if you have any problems 😁

pkg/hintrunner/zero/zerohint.go
Alvarodb: Oops, that slipped through! I've fixed it now.

Alvarodb:
> Thank you for you contribution! If you resolve the the len as Felt and load it to the scope, you can try and write some unit tests, which check the insides of the scope. Write on Telegram if you have any problems 😁

Thank you for all the help! I will work on the unit tests.

Alvarodb:
unit test added for the hint

pkg/hintrunner/zero/hintcode.go
```diff
@@ -85,6 +85,6 @@ const (
 
 	// ------ Other hints related code ------
 	allocSegmentCode string = "memory[ap] = segments.add()"
-	vmEnterScopeCode string = "vm_enter_scope()"
+	memcpyEnterScope string = "vm_enter_scope({'n': ids.len})"
```
MaksymMalicki, line 88: nitpick: rename to memcpyEnterScopeCode 😉

MaksymMalicki:
One nitpick left, other than that, looks good to me. Good job! 😎
@cicr99 final look?

MaksymMalicki:
Okey, you need to lint is as well using ```gofmt -w .``` or ```golangci-lint run```

pkg/hintrunner/zero/hintcode.go
Alvarodb: fixed!

Alvarodb:
> Okey, you need to lint is as well using `gofmt -w .` or `golangci-lint run`

Oh sure, fixed

Alvarodb:
> One nitpick left, other than that, looks good to me. Good job! 😎 @cicr99 final look?

I think everything is fixed now guys, thank very much! Let me know if any issues.

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -10,14 +10,27 @@ func createAllocSegmentHinter(resolver hintReferenceResolver) (hinter.Hinter, er
 	return &core.AllocSegment{Dst: hinter.ApCellRef(0)}, nil
 }
 
-func createVMEnterScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
+func newMemcpyEnterScopeHinter(len hinter.ResOperander) hinter.Hinter {
```
TAdev0, line 13: the `new ...` function for hints is supposed to end with `Hint`, not `Hinter`.

So `newMemcpyEnterScopeHint` would be the right name

if you look at uint256, usort, ec , keccak hints, you will see this is the convention. But indeed, in signatures hint , there is `newVerifyECDSASignatureHinter` , `newGetPointFromXHinter`, `newImportSecp256R1PHinter` and `newDivModSafeDivHinter` , it should be renamed with Hint at the end.

But functions starting with `create...` ends with `Hinter` , and here it is ok

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -10,14 +10,27 @@ func createAllocSegmentHinter(resolver hintReferenceResolver) (hinter.Hinter, er
...
-		Name: "VMEnterScope",
+		Name: "MemcpyEnterScope",
 		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
-			ctx.ScopeManager.EnterScope(make(map[string]any))
+			//>  vm_enter_scope({'n': ids.len})
+			len, err := hinter.ResolveAsFelt(vm, len)
+			if err != nil {
+				return err
+			}
+			ctx.ScopeManager.EnterScope(map[string]any{"n": len})
```
TAdev0, line 22: lgtm!

TAdev0 (changes requested):

pkg/hintrunner/zero/zerohint.go
Alvarodb: I see, thanks for the explanation. This is fixed now!

pkg/hintrunner/zero/zerohint_memcpy.go
Alvarodb: I see, thanks for the explanation. This should be fixed now.

TAdev0 (approved):

TAdev0:
LGTM ! waiting for final review @rodrigo-pino  @chee-chyuan  @MaksymMalicki  @har777  and we're good to go :)

TAdev0:
@Alvarodb  will compare the result of your hint with the result produced by the python VM  today to be 100% sure and then we're good to go!

Alvarodb:
> @Alvarodb will compare the result of your hint with the result produced by the python VM today to be 100% sure and then we're good !

Ok, sounds good! Thank you @TAdev0

TAdev0:
@Alvarodb  i just tested your hint. Well, there is not much to test actually. The `len` variable is simply extracted from the current and passed when entering a new scope, so that the new scope contains `len` as the only key, before copying memory from a location to another. 

I used this cairo file that i ran on the python VM : 

```
from starkware.cairo.common.alloc import alloc
from starkware.cairo.common.math import assert_not_zero

// Copies len field elements from src to dst.
func memcpy(dst: felt*, src: felt*, len) {
    struct LoopFrame {
        dst: felt*,
        src: felt*,
    }

    if (len == 0) {
        return ();
    }

    %{ vm_enter_scope({'n': ids.len}) %}
    %{ print(n) %}

    tempvar frame = LoopFrame(dst=dst, src=src);

    loop:
    let frame = [cast(ap - LoopFrame.SIZE, LoopFrame*)];
    assert [frame.dst] = [frame.src];

    let continue_copying = [ap];
    // Reserve space for continue_copying.
    let next_frame = cast(ap + 1, LoopFrame*);
    next_frame.dst = frame.dst + 1, ap++;
    next_frame.src = frame.src + 1, ap++;
    %{
        n -= 1
        ids.continue_copying = 1 if n > 0 else 0
    %}
    static_assert next_frame + LoopFrame.SIZE == ap + 1;
    jmp loop if continue_copying != 0, ap++;
    // Assert that the loop executed len times.
    len = cast(next_frame.src, felt) - cast(src, felt);

    %{ vm_exit_scope() %}
    return ();
}

func main() {
    alloc_locals;

    let (dest: felt*) = alloc();
    let source: felt* = alloc();

    assert source[0] = 1;
    assert source[1] = 1;
    assert source[2] = 1;

    memcpy(dest, source, 3);

    assert_not_zero(dest[0]);
    assert dest[0] = source[0];
    assert_not_zero(dest[1]);
    assert dest[1] = source[1];
    assert_not_zero(dest[2]);
    assert dest[2] = source[2];

    return ();
}
```

I added  a new hint `%{ print(n) %}` to print the value of `n` in the scope. Then, in `main`, i call `    memcpy(dest, source, 3);` to test here with `len = 3` .

When i run this main function, i get `3` as output 
Then, in the test file you created, i used `3` as an expected value for `n` when defining `len` as `3`.  

I also tested for fun with many other values, small or big, everything works well without surprise.
I guess testing this was not necessary but now we're 100% sure :)

TAdev0:
@rodrigo-pino  @cicr99 @MaksymMalicki  @har777  waiting for a 2nd approval !

pkg/hintrunner/zero/hintcode.go
```diff
@@ -84,7 +84,7 @@ const (
 	// ------ Dictionaries hints related code ------
 
 	// ------ Other hints related code ------
-	allocSegmentCode string = "memory[ap] = segments.add()"
-	vmEnterScopeCode string = "vm_enter_scope()"
-	vmExitScopeCode  string = "vm_exit_scope()"
+	allocSegmentCode     string = "memory[ap] = segments.add()"
+	memcpyEnterScopeCode string = "vm_enter_scope({'n': ids.len})"
```
cicr99, line 88: Oh I see you wrote on top of `vmEnterScopeCode`. These two are different hints so you need to handle them separately

cicr99 (changes requested):
It look s great in general but there is the problem I pointed out in the previous comment. I see you wrote on top of the VMEnterScope hint, they are very similar but they are different ones, so you'll need to add memcpyEnterScope as a separate one.
See here an example where VMEnterScope is being used https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/squash_dict.cairo#L28

pkg/hintrunner/zero/hintcode.go
Alvarodb: Got it, should be fixed now!

Alvarodb:
> It looks great in general but there is the problem I pointed out in the comment. I see you wrote on top of the `VMEnterScope` hint, they are very similar but they are different ones, so you'll need to add `memcpyEnterScope` as a separate one. See here an example where `VMEnterScope` is being used https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/squash_dict.cairo#L28

I've added memcpyEnterScope as a separate function, keeping the original implementation of VMEnterScope intact. Thanks for the feedback!

TAdev0 (changes requested):

pkg/hintrunner/zero/hintcode.go
```diff
@@ -84,7 +84,8 @@ const (
 	// ------ Dictionaries hints related code ------
 
 	// ------ Other hints related code ------
-	allocSegmentCode string = "memory[ap] = segments.add()"
-	vmEnterScopeCode string = "vm_enter_scope()"
-	vmExitScopeCode  string = "vm_exit_scope()"
+	allocSegmentCode     string = "memory[ap] = segments.add()"
+	vmEnterScopeCode     string = "vm_enter_scope()"
```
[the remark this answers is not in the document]
Alvarodb: makes sense, should be good to go now! ty

TAdev0 (approved):

TAdev0:
LGTM @cicr99

har777:
lgtm!

PR #373, opened 2024-04-26
Implement blockPermutation hint

TAdev0:
@MaksymMalicki  @cicr99  @rodrigo-pino  this is a first draft.

I changed all the code to use `KeccakF1600` as specified by Rodrigo. This function doesn't return anything but alters the value of the variable the pointer passed as argument points to. This result is then written to memory starting at `keccakPtr` address. 

There is a check in the python hint to revert if  ` keccakStateSize >= 100` , but `keccakStateSize` is defined as a constant value (25). I'm wondering if this check could be removed?

Also, i can see that the `Bits` method actually returns a little endian `[4]uint64` , is this appropriate?


pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -102,3 +105,93 @@ func createKeccakWriteArgsHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+			if err != nil {
+				return err
+			}
+
+			keccakWritePtr, err := hinter.ResolveAsAddress(vm, keccakPtr)
+			if err != nil {
+				return err
+			}
+
+			// why does the need contains this check? keccakStateSize is a constant: 25
```
MaksymMalicki, line 132: Maybe it may be overwritten sometime? Or someone can change that, but that doesn't seem likely. I thing although it seems to be intuitive not to check that, maybe there is some edgecase that we don't see.

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -102,3 +105,93 @@ func createKeccakWriteArgsHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+			if keccakStateSize >= 100 {
+				return fmt.Errorf("keccakStateSize %v is out range 0 <= keccakStateSize < 100", &keccakStateSize)
+			}
+
+			var readAddr = *keccakWritePtr
+			var offset int16 = int16(keccakStateSize)
+			var negOffset int16 = -offset
+
+			readAddr, err = readAddr.AddOffset(negOffset)
```
MaksymMalicki, line 141: Did you check what happens, when the address resulting in this subtraction is negative?

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -102,3 +105,93 @@ func createKeccakWriteArgsHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+				valueUint64 := valueUint256.Uint64()
+				keccakInput = append(keccakInput, valueUint64)
+			}
+
+			var input [25]uint64
+
+			for i := 0; i < 25; i++ {
+				input[i] = keccakInput[i]
+			}
+			builtins.KeccakF1600(&input)
```
MaksymMalicki, line 168: When looking at this something crossed my mind. I was using this function with an array, which contained _keccak_state_size_felts * [0]. So the _keccak_state_size_felts could be 1, could be 99, the result was always the same as for fixed size keccak_input[25]. Assuming, that __keccak_state_size_felts can be different than 25, did you check that the output is the same? For instance, if the __keccak_state_size_felts is 30, in the loop above you are making a cutoff at 25. I'm just curious. 

We have to investigate, if the __keccak_state_size_felts can be different than 25, because if so, implementing the keccak_func by hand might be neccessary.

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -102,3 +105,93 @@ func createKeccakWriteArgsHinter(resolver hintReferenceResolver) (hinter.Hinter,
 
 	return newKeccakWriteArgsHint(inputs, low, high), nil
 }
+
+func newBlockPermutationHint(keccakStateSizeFelts, keccakPtr hinter.ResOperander) hinter.Hinter {
+	name := "BlockPermutation"
+	return &GenericZeroHinter{
+		Name: name,
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+			// from starkware.cairo.common.keccak_utils.keccak_utils import keccak_func
```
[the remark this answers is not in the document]
TAdev0: oh yes nice, thank you!! i didn't do it for my previous hint that was merged, will add it in this pr

pkg/hintrunner/zero/zerohint_keccak.go
TAdev0: yeah i'm trying to think of a reason. But the python VM code is clear, and declares a `KECCAK_STATE_SIZE_FELTS`  const. I dont see any reason for this valid to be something else than 25

pkg/hintrunner/zero/zerohint_keccak.go
TAdev0: keccakF1600 takes `*[25]uint64` as input, i cannot be > 25

are you talking about CairoKeccakFinalize ? I just look at it your code is : 

```
var input [25]uint64
builtins.KeccakF1600(&input)
```

you pass an array of 25 uint64. I think it must always be 25 elements

TAdev0: by the way, for CairoKeccakFinalize, the python code is `keccak_func(inp)` with `inp = [0] * _keccak_state_size_felts`

is it correct to apply KeccakF1600 to an array of 25 elements set to 0 ? what is `[0]` ?

TAdev0: I think that because `KeccakF1600` necessarily takes a ` [25]uint64` , indeed you can have only 1 or 2 values set to a non zero value and you just pass an array of 25 elements with all other values set to 0. But the keccak state cannot be > 25 I suppose as you just couldnt pass it to `KeccakF1600`

pkg/hintrunner/zero/zerohint_keccak.go
TAdev0: i didnt actually. Will do it. I I still need to run the hint on the python vm and do tests

pkg/hintrunner/zero/zerohint_keccak.go
MaksymMalicki: Yea makes sense, but still leave the check here, lets implement what is directly in the hint code

pkg/hintrunner/zero/zerohint_keccak.go
MaksymMalicki: Regarding your previous comment, I have checked the keccak_func(inp) for different length of arrays filled with zero, and the output was always the same as for the array of length 25, so that is not a prolem.

MaksymMalicki: > by the way, for CairoKeccakFinalize, the python code is `keccak_func(inp)` with `inp = [0] * _keccak_state_size_felts`
> 
> is it correct to apply KeccakF1600 to an array of 25 elements set to 0 ? what is `[0]` ?

It is just array of __keccak_state_size_felts filled with zeros

MaksymMalicki: > I think that because `KeccakF1600` necessarily takes a ` [25]uint64` , indeed you can have only 1 or 2 values set to a non zero value and you just pass an array of 25 elements with all other values set to 0. But the keccak state cannot be > 25 I suppose as you just couldnt pass it to `KeccakF1600`

this seems legit, I dont see any overwriting of the 25 const, so you are right

TAdev0: oh yes ok great i'm not used to go syntax sometimes i get confused
perfect!

TAdev0:
@MaksymMalicki  we loop over `inputValuesInRange` to fill the `keccakInput []uint64` array. 

And `inputValuesInRange` is calculated as follows:
```
			inputValuesInRange, err := hinter.GetConsecutiveValues(vm, readAddr, int16(keccakStateSize))
```

or in the python hint : 
```
memory.get_range(ids.keccak_ptr - _keccak_state_size_felts, _keccak_state_size_felts)
```

It seems obvious that `keccakInput` will never contain more than `keccakStateSize` elements, do you confirm?

TAdev0:
@cicr99  @MaksymMalicki  @har777  ok so i'm having real difficulty to test in an isolated way this hint. I always get errors like
```
Error:          Expected nil, but got: &fmt.wrapError{msg:"segment 1, offset 7: no builtin: reading unknown value"` 
```
when trying to run the test without any check. I have no idea how to create a setup to run this hint separately at the moment.

TAdev0:
@MaksymMalicki  thank you so much for the help !

TAdev0:
@MaksymMalicki  i think it doesn't work. The hint will read from memory at location `keccakPtr` , up to 25 elements. It will then write at cell keccakPtr + 25, for 25 more slots.

So i guess initialized ap, ap+ 1.... ap +24 doesn't solve the issue?

MaksymMalicki:
lgtm! 🚀

MaksymMalicki (approved):

TAdev0:
@cicr99  @rodrigo-pino  waiting for your review to merge

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -111,3 +114,101 @@ func createKeccakWriteArgsHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+				valueUint256 := uint256.Int(valueFelt.Bits())
+				valueUint64 := valueUint256.Uint64()
+				keccakInput = append(keccakInput, valueUint64)
+			}
+
+			var input [25]uint64
+
+			for i := 0; i < 25; i++ {
+				input[i] = keccakInput[i]
+			}
```
cicr99, line 184: Would it be better to just directly write on the `input`? You could assert the size of `inputValuesInRange` is the correct one as well. wdyt 🤔 ?

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -111,3 +114,101 @@ func createKeccakWriteArgsHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+
+			var keccakInput []uint64
+
+			for _, valueMemoryValue := range inputValuesInRange {
+				valueFelt, err := valueMemoryValue.FieldElement()
+				if err != nil {
+					return err
+				}
+				valueUint256 := uint256.Int(valueFelt.Bits())
+				valueUint64 := valueUint256.Uint64()
```
cicr99, line 176: Why not use `Uint64()` method for Memory Values here instead? Is it possible the elements you get in `inputValuesInRange` don't fit?

TAdev0: i think i was just not aware at the time i could cast it directly from MemoryValue
thanks!

pkg/hintrunner/zero/zerohint_keccak.go
```diff
@@ -111,3 +114,101 @@ func createKeccakWriteArgsHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+			//> from starkware.cairo.common.keccak_utils.keccak_utils import keccak_func
+			//> _keccak_state_size_felts = int(ids.KECCAK_STATE_SIZE_FELTS)
+			//> assert 0 <= _keccak_state_size_felts < 100
+			//
+			//> output_values = keccak_func(memory.get_range(
+			//> 	ids.keccak_ptr - _keccak_state_size_felts, _keccak_state_size_felts))
+			//> segments.write_arg(ids.keccak_ptr, output_values)
+
+			keccakStateSize, err := hinter.ResolveAsUint64(vm, keccakStateSizeFelts)
```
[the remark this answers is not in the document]
TAdev0: yes ! will do that!

pkg/hintrunner/zero/zerohint_keccak.go
TAdev0: will check that! thanks!

pkg/hintrunner/zero/zerohint_keccak.go
TAdev0: indeed its way better with your solution:)

pkg/hintrunner/zero/zerohint_keccak.go
TAdev0: well we dont even need to assert the size of `inputValuesInRange` as its necessarly 25 after hardcoding keccakStateSize

so basically we can remove a loop ! great :)

cicr99 (approved):

PR #375, opened 2024-04-27
feat: blake2s finalize

### LINKED ISSUE #280: Blake2sFinalize
Finalizes the BLAKE2s hashing process. It involves asserting the packed instances and input chunk size, preparing a dummy message, computing the hash with modified initialization vectors, and adding padding to align with the number of packed instances.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/cairo_blake2s/blake2s.cairo#L184

This hint needs to implement `blake2s_compress` method which is also used in #279, so whichever is solved first will make the other one easier

TropicalDog17:
closes #280


pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +105,103 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
 
 	return newBlake2sAddUint256Hint(low, high, data, bigend), nil
 }
+
+// \\# Add dummy pairs of input and output.
```
MaksymMalicki, line 109: what does this comment do? if it's not neccessary you can remove it. Thanks!

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +105,103 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			//>     h=modified_iv,
+			//>     t0=0,
+			//>     t1=0,
+			//>     f0=0xffffffff,
+			//>     f1=0,
+			//> )
+			//> padding = (modified_iv + message + [0, 0xffffffff] + output) * (_n_packed_instances - 1)
+			//> segments.write_arg(ids.blake2s_ptr_end, padding)
+
+			blake2s_ptr_end, err := hinter.ResolveAsAddress(vm, blake2s_ptr_end)
```
MaksymMalicki, line 156: please use cammel case

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +105,103 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			if err != nil {
+				return err
+			}
+			N_PACKED_INSTANCES, err := hinter.ResolveAsFelt(vm, N_PACKED_INSTANCES)
+			if err != nil {
+				return err
+			}
+			if err != nil {
```
MaksymMalicki, line 164: double error check

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +105,103 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			//>     f1=0,
+			//> )
+			//> padding = (modified_iv + message + [0, 0xffffffff] + output) * (_n_packed_instances - 1)
+			//> segments.write_arg(ids.blake2s_ptr_end, padding)
+
+			blake2s_ptr_end, err := hinter.ResolveAsAddress(vm, blake2s_ptr_end)
+			if err != nil {
+				return err
+			}
+			N_PACKED_INSTANCES, err := hinter.ResolveAsFelt(vm, N_PACKED_INSTANCES)
```
MaksymMalicki, line 160: same here, use cammel case please

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +105,103 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			if err != nil {
+				return err
+			}
+			if err != nil {
+				return err
+			}
```
MaksymMalicki, line 166: missing ```assert 0 <= _n_packed_instances < 20```

pkg/hintrunner/zero/zerohint_blake.go
MaksymMalicki: Also, you can resolve it as uint64 right here, instead of inside of the for loop later

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +105,103 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			var message [16]uint32
+			modifiedIv := utils.IV()
+			modifiedIv[0] = modifiedIv[0] ^ 0x01010020
+			output := utils.Blake2sCompress(modifiedIv, message, 0, 0, 0xffffffff, 0)
+			padding := modifiedIv[:]
+			padding = append(padding, message[:]...)
+			padding = append(padding, 0, 0xffffffff)
+			padding = append(padding, output[:]...)
+			fullPadding := padding
+			for i := uint32(1); i < uint32(N_PACKED_INSTANCES.Uint64()); i++ {
```
MaksymMalicki, line 176: this can even be packed into uint8, since the nPackedInstances in in range [0;20]

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +105,103 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+				fullPadding = append(fullPadding, padding...)
+			}
+			for i, val := range fullPadding {
+				mv := mem.MemoryValueFromInt(val)
+				err = vm.Memory.Write(blake2s_ptr_end.SegmentIndex, blake2s_ptr_end.Offset+uint64(i), &mv)
```
MaksymMalicki, line 181: There is a functionality for incrementing the offset, called .AddOffset(offset), which is made safe by checking overflows and underflows

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +105,103 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+				}
+			}
+			return err
+		},
+	}
+
+}
+
+func createBlake2sFinalizeHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
+	blake2s_ptr_end, err := resolver.GetResOperander("blake2s_ptr_end")
```
MaksymMalicki, line 193: camel cases here as well. Thanks!

pkg/hintrunner/zero/zerohint_blake_test.go
```diff
@@ -427,5 +427,72 @@ func TestZeroHintBlake(t *testing.T) {
 					}),
 			},
 		},
+		"Blake2sFinalize": {
+			{
+				operanders: []*hintOperander{
+					{Name: "blake2s_ptr", Kind: apRelative, Value: addrWithSegment(1, 7)},
+					{Name: "N_PACKED_INSTANCES", Kind: fpRelative, Value: feltString("7")},
+					{Name: "data", Kind: apRelative, Value: addrWithSegment(1, 7)},
```
MaksymMalicki, line 435: There is no res operander called data. You have defined it in the same way as blake2s_ptr, so there is no point in passing it as well. You have however forgotten about the ```INPUT_BLOCK_FELTS```

pkg/hintrunner/zero/zerohint_blake_test.go
```diff
@@ -427,5 +427,72 @@ func TestZeroHintBlake(t *testing.T) {
 					}),
 			},
 		},
+		"Blake2sFinalize": {
```
MaksymMalicki, line 430: if you add the assert, you can add some tests which prove, that the assert works

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +105,103 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+	if err != nil {
+		return nil, err
+	}
+	N_PACKED_INSTANCES, err := resolver.GetResOperander("N_PACKED_INSTANCES")
+	if err != nil {
+		return nil, err
+	}
+	message, err := resolver.GetResOperander("message")
```
MaksymMalicki, line 201: There is no listed ResOperander called ```message```, the only ones are:
```N_PACKED_INSTANCES```, ```blake2s_ptr_end``` and ```INPUT_BLOCK_FELTS```

File: pkg/hintrunner/zero/zerohint_blake.go (hunk 5 above)
Comment (MaksymMalicki) [inline pkg/hintrunner/zero/zerohint_blake.go:166]: missing ```assert 0 <= _blake2s_input_chunk_size_felts < 100```

File: pkg/hintrunner/zero/zerohint_blake_test.go [hunk 12]
```diff
@@ -427,5 +427,72 @@ func TestZeroHintBlake(t *testing.T) {
...
+					{Name: "N_PACKED_INSTANCES", Kind: fpRelative, Value: feltString("7")},
+					{Name: "data", Kind: apRelative, Value: addrWithSegment(1, 7)},
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newBlake2sFinalizeHint(ctx.operanders["blake2s_ptr"], ctx.operanders["N_PACKED_INSTANCES"], ctx.operanders["data"])
+				},
+				check: consecutiveVarAddrResolvedValueEquals(
+					"data",
+					[]*fp.Element{
+						feltString("1795745351"), feltString("3144134277"), feltString("1013904242"), feltString("2773480762"),
```
MaksymMalicki, line 443: for such small numbers please use feltUint64, thanks!

pkg/utils/blake.go
```diff
@@ -0,0 +1,133 @@
...
+		0xBB67AE85,
+		0x3C6EF372,
+		0xA54FF53A,
+		0x510E527F,
+		0x9B05688C,
+		0x1F83D9AB,
+		0x5BE0CD19}
+}
+
+func SIGMA() [10][16]uint32 {
```
MaksymMalicki, line 15: all of those can be uint8 I suppose

TropicalDog17:
@MaksymMalicki Hi, I have refactored based on your reviews, please kindly check it out. Thanks!

pkg/hintrunner/zero/zerohint_blake.go
MaksymMalicki: I mean this entire comment, it is the copy of the comment below, the same code. You can remove this entire comment block and keep the one in the method constructor

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +106,120 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			for i := uint8(1); i < uint8(nPackedInstances); i++ {
+				fullPadding = append(fullPadding, padding...)
+			}
+			for i, val := range fullPadding {
+				mv := mem.MemoryValueFromInt(val)
+				memWithOffset, err := blake2sPtrEnd.AddOffset(int16(i))
+				if err != nil {
+					return err
+				}
+				err = vm.Memory.Write(blake2sPtrEnd.SegmentIndex, memWithOffset.Offset, &mv)
```
MaksymMalicki, line 199: you can use ```WriteToAddress``` here, so it's cleaner

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +106,120 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			//>     f1=0,
+			//> )
+			//> padding = (modified_iv + message + [0, 0xffffffff] + output) * (_n_packed_instances - 1)
+			//> segments.write_arg(ids.blake2s_ptr_end, padding)
+
+			blake2sPtrEnd, err := hinter.ResolveAsAddress(vm, blake2sPtrEnd)
+			if err != nil {
+				return err
+			}
+			nPackedInstancesElement, err := hinter.ResolveAsFelt(vm, nPackedInstances)
```
MaksymMalicki, line 159: There is a method called ```hinter.ResolveAsUint64```. You can use it and also for inputBlockFeltElement

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +106,120 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+				memWithOffset, err := blake2sPtrEnd.AddOffset(int16(i))
+				if err != nil {
+					return err
+				}
+				err = vm.Memory.Write(blake2sPtrEnd.SegmentIndex, memWithOffset.Offset, &mv)
+				if err != nil {
+					return err
+				}
+			}
+			return err
```
MaksymMalicki, line 204: I think we should return ```nil``` here. Sure the error value will be nil, but this will be much clearer

pkg/hintrunner/zero/zerohint_blake.go
MaksymMalicki: coud you also rename ```message``` to ```inputBlocksFelt```?

pkg/utils/blake.go
```diff
@@ -0,0 +1,133 @@
+package utils
+
+func IV() [8]uint32 {
```
MaksymMalicki, line 3: why do you use uint32 for all the types in this file? Is it impossible for the values to exceed 2**32?

pkg/utils/blake.go
```diff
@@ -0,0 +1,133 @@
...
+		{9, 0, 5, 7, 2, 4, 10, 15, 14, 1, 11, 12, 6, 8, 3, 13},
+		{2, 12, 6, 10, 0, 11, 8, 3, 4, 13, 7, 5, 15, 14, 1, 9},
+		{12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11},
+		{13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10},
+		{6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5},
+		{10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0},
+	}
+}
+
+func rightRot(value uint32, n uint32) uint32 {
```
MaksymMalicki, line 30: can you provide some unittests for those functionalities? It seems like a lot of utilities which go untested. Thanks!

pkg/utils/blake.go
TropicalDog17: yes the initialization vector is supposed to be a constant, [8]uint32 is correct. SIGMA can be uint8 though, I've fixed.

pkg/utils/blake.go
TropicalDog17: Done, please review @MaksymMalicki

pkg/utils/blake.go
MaksymMalicki: okok, good job 🚀

pkg/utils/blake.go
```diff
@@ -0,0 +1,133 @@
...
+		state[4],
+		state[9],
+		state[14],
+		message[sigma[14]],
+		message[sigma[15]],
+	)
+	return state
+}
+
+func Blake2sCompress(message []uint32, h [8]uint32, t0 uint32, t1 uint32, f0 uint32, f1 uint32) []uint32 {
```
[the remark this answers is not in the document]
TropicalDog17: from page 20 of this [paper](https://www.blake2.net/blake2.pdf)
> A.2 BLAKE2s
BLAKE2s supports data of any byte length 0 ≤ ` < 264. It works similarly to BLAKE2b, but on
32-bit words instead of 64-bit words (the byte length of a chaining value, a message block, a
counter or finalization flag are thus divided by two).

so message in uint32s is valid, but not uint64 or bigger.

TropicalDog17:
Hi @cicr99 @MaksymMalicki, could you kindly take a look at my PR. Thanks!

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +106,95 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			if err != nil {
+				return err
+			}
+
+			// assert 0 <= _n_packed_instances < 20
+			if nPackedInstances >= 20 {
+				return fmt.Errorf("n_packed_instances should be in range [0, 20), got %d", nPackedInstances)
+			}
+
+			inputBlockFelt, err := hinter.ResolveAsUint64(vm, inputBlockFelt)
```
MaksymMalicki, line 148: could you rename this to blake2sInputChunkSizeFelts?

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -103,3 +106,95 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			message := make([]uint32, inputBlockFelt)
+			modifiedIv := utils.IV()
+			modifiedIv[0] = modifiedIv[0] ^ 0x01010020
+			output := utils.Blake2sCompress(message, modifiedIv, 0, 0, 0xffffffff, 0)
+			padding := modifiedIv[:]
+			padding = append(padding, message[:]...)
+			padding = append(padding, 0, 0xffffffff)
+			padding = append(padding, output[:]...)
+			fullPadding := padding
+			for i := uint8(1); i < uint8(nPackedInstances); i++ {
```
MaksymMalicki, line 166: i think it would be much clearer, if you iterate from 0 to nPackedInstances - 1. Current code is kind of confusing

MaksymMalicki: Also, if you assign
```fullPadding := padding```
and then append ```nPackedInstances - 1``` times, wont this cause the fullPadding to be padding multiplied by ```nPackedInstances``` instead of ```nPackedInstances - 1``` as this line states:
```padding = (modified_iv + message + [0, 0xffffffff] + output) * (_n_packed_instances - 1)```

File: pkg/hintrunner/zero/zerohint_blake_test.go [hunk 22]
```diff
@@ -427,5 +427,98 @@ func TestZeroHintBlake(t *testing.T) {
...
+			{
+				operanders: []*hintOperander{
+					{Name: "blake2s_ptr_end", Kind: apRelative, Value: addrWithSegment(1, 7)},
+					{Name: "N_PACKED_INSTANCES", Kind: fpRelative, Value: feltUint64(7)},
+					{Name: "INPUT_BLOCK_FELTS", Kind: apRelative, Value: feltUint64(200)},
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newBlake2sFinalizeHint(ctx.operanders["blake2s_ptr_end"], ctx.operanders["N_PACKED_INSTANCES"], ctx.operanders["INPUT_BLOCK_FELTS"])
+				},
+				errCheck: errorTextContains("in range [0, 100), got 200"),
```
MaksymMalicki, line 520: could you add the entire errorText in the errCheck using formatted string?

pkg/hintrunner/zero/zerohint_blake.go
TropicalDog17: Yeah my bad, I will fix it

pkg/hintrunner/zero/zerohint_blake_test.go
MaksymMalicki: you can leave the 0 and 100 and place 200 as variable in formatted string, no need to pass "[0, 100)" as separate argument to formatted string

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -115,3 +117,95 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			padding = append(padding, message[:]...)
+			padding = append(padding, 0, 0xffffffff)
+			padding = append(padding, output[:]...)
+			fullPadding := []uint32{}
+			for i := uint64(0); i < nPackedInstances-1; i++ {
+				fullPadding = append(fullPadding, padding...)
+			}
+			for i, val := range fullPadding {
+				mv := mem.MemoryValueFromInt(val)
+				memWithOffset, err := blake2sPtrEnd.AddOffset(int16(i))
```
MaksymMalicki, line 182: since nPackedInstances can be up to uint64 (unlikely but possible), I think the i might fall out of range. It would be good if you incremented the blake2sPtrEnd or its copy by 1 in each iteration. That way it doesn't overflow and correctly includes the entire possible range.

pkg/hintrunner/zero/zerohint_blake_test.go
```diff
@@ -427,5 +428,98 @@ func TestZeroHintBlake(t *testing.T) {
...
+			{
+				operanders: []*hintOperander{
+					{Name: "blake2s_ptr_end", Kind: apRelative, Value: addrWithSegment(1, 7)},
+					{Name: "N_PACKED_INSTANCES", Kind: fpRelative, Value: feltUint64(20)},
+					{Name: "INPUT_BLOCK_FELTS", Kind: apRelative, Value: feltUint64(16)},
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newBlake2sFinalizeHint(ctx.operanders["blake2s_ptr_end"], ctx.operanders["N_PACKED_INSTANCES"], ctx.operanders["INPUT_BLOCK_FELTS"])
+				},
+				errCheck: errorTextContains(fmt.Sprintf("in range %s, got %s", "[0, 20)", "20")),
```
MaksymMalicki, line 508: Also in this errorTextContains I meant you can just go for:
```errCheck: errorTextContains(fmt.Sprintf("n_packed_instances should be in range [0, 20), got %d", 20))```
since the error message is hardcoded right?

File: pkg/hintrunner/zero/zerohint_blake_test.go [hunk 25]
```diff
@@ -427,5 +428,98 @@ func TestZeroHintBlake(t *testing.T) {
...
+			{
+				operanders: []*hintOperander{
+					{Name: "blake2s_ptr_end", Kind: apRelative, Value: addrWithSegment(1, 7)},
+					{Name: "N_PACKED_INSTANCES", Kind: fpRelative, Value: feltUint64(7)},
+					{Name: "INPUT_BLOCK_FELTS", Kind: apRelative, Value: feltUint64(200)},
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newBlake2sFinalizeHint(ctx.operanders["blake2s_ptr_end"], ctx.operanders["N_PACKED_INSTANCES"], ctx.operanders["INPUT_BLOCK_FELTS"])
+				},
+				errCheck: errorTextContains(fmt.Sprintf("in range %s, got %s", "[0, 100)", "200")),
```
MaksymMalicki, line 521: same here

pkg/hintrunner/zero/zerohint_blake.go
TropicalDog17: Hey but didn't we have have an assertion that make sure `nPackedInstances` is between 0 and 20?

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -115,3 +117,96 @@ func createBlake2sAddUint256Hinter(resolver hintReferenceResolver, bigend bool)
...
+			padding = append(padding, 0, 0xffffffff)
+			padding = append(padding, output[:]...)
+			fullPadding := []uint32{}
+			for i := uint64(0); i < nPackedInstances-1; i++ {
+				fullPadding = append(fullPadding, padding...)
+			}
+
+			for _, val := range fullPadding {
+				mv := mem.MemoryValueFromInt(val)
+				blake2sPtrEnd, err := blake2sPtrEnd.AddOffset(1)
```
MaksymMalicki, line 183: I think this needs to be incremented at the end of the loop, after you write to the memory

MaksymMalicki:
Also your unit tests are not passing, make sure to check it. Thanks!

har777:
This looks great!

har777 (approved):

MaksymMalicki (approved):

Issue #253, opened 2024-03-11
SearchSortedLower: Searches for the first element in a sorted array that is greater than or equal to a given key and returns its index.


cicr99:

joseSalazar4:
Hey there! Hope you are doing good, Is this issue available? I would like to work on this one.

rodrigo-pino:
Hey @joseSalazar4, it's yours. @danielcdz you're already working on another issue so once you finish that one we will have more for you!

MaksymMalicki:
Hey I'm wondering if there is some syntax problem in the hint code, because the ```else``` block is standalone, so there is an indentation error. Now I'm wondering if the ```for``` or the ```else``` block is misindented, but from the hint code I assume that its the ```else``` block.

joseSalazar4:
Hey! Maksym, actually it's proper syntax, please check this out to see the behavior: https://www.w3schools.com/python/gloss_python_for_else.asp

joseSalazar4:
No worries, it is a very strange syntax tbh shouldn't even be used hahaha but you're welcome! 🥇

PR #376, opened 2024-04-28
Implement searchSortedLower Hint

joseSalazar4:
@cicr99 @rodrigo-pino @MaksymMalicki  Hey there! This is a draft of what I have so far, there's no tests yet due to  me not sure if the go implementation is okay. This tries to implement [this issue](https://github.com/NethermindEth/cairo-vm-go/issues/253)


pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,98 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+			indexAddr, err := index.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+			//> for i in range(n_elms):
+			//> 	if memory[array_ptr + elm_size * i] >= ids.key:
+			//> 		ids.index = i
+			//> 		break
+			for i := uint64(0); i < n_elms.Uint64(); i++ {
+				v := memory.MemoryValueFromFieldElement(utils.Felt127.Add(array_ptr, elm_size))
```
MaksymMalicki, line 114: I belive, you have to resolve array_ptr as an address, add elem_size * i using ```array_ptr.AddOffset()```. Then resolve the read value as felt and compare it with key

joseSalazar4: Thanks for the hand, makes lot of sense that method. Changed!

joseSalazar4:
I see that there's no test file for memcpy so is it necessary for this issue to be solved? 🤔 😄 Anyways, comments addressed

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,98 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+	return newSearchSortedLowerHint(array_ptr, elm_size, n_elms, key, index), nil
+}
+
+func newSearchSortedLowerHint(array_ptr, elm_size, n_elms, key, index hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "SearchSortedLower",
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+
+			//> array_ptr = ids.array_ptr
+			array_ptr, err := hinter.ResolveAsFelt(vm, array_ptr)
```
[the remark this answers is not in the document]
joseSalazar4: Modified!

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,98 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+			if err != nil {
+				return err
+			}
+
+			//> elm_size = ids.elm_size
+			elm_size, err := hinter.ResolveAsFelt(vm, elm_size)
+
+			//> assert isinstance(elm_size, int) and elm_size > 0, \
+			//> 	f'Invalid value for elm_size. Got: {elm_size}.'
+			if err != nil && utils.FeltLt(&utils.FeltZero, elm_size) {
```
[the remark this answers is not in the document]
joseSalazar4: Modified! I used the isZero() func

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,98 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+				return fmt.Errorf("Invalid value for elm_size. Got: %v.", elm_size)
+			}
+
+			//> n_elms = ids.n_elms
+			//> assert isinstance(n_elms, int) and n_elms >= 0, \
+			//> 	f'Invalid value for n_elms. Got: {n_elms}.'
+			n_elms, err := hinter.ResolveAsFelt(vm, n_elms)
+			if err != nil && utils.FeltLt(&utils.FeltZero, n_elms) {
```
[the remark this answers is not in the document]
joseSalazar4: Modified!

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,98 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+				return fmt.Errorf("Invalid value for n_elms. Got: %v.", n_elms)
+			}
+
+			var __find_element_max_size *fp.Element
+			// TODO: How to do this --> if __find_element_max_size is a variable?
```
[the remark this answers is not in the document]
joseSalazar4: I am using the context to get the var now. Modified!

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,119 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+
+			//> n_elms = ids.n_elms
+			//> assert isinstance(n_elms, int) and n_elms >= 0, \
+			//> 	f'Invalid value for n_elms. Got: {n_elms}.'
+			n_elms, err := hinter.ResolveAsFelt(vm, n_elms)
+
+			if err != nil {
+				return err
+			}
+			if n_elms.IsZero() {
```
cicr99, line 91: See here that the condition is `n_elms >= 0`, hence the fact that `n_elms.IsZero()` is true does not contradict the previous fact. I would suggest using `FeltLt` utility instead

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,119 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+			}
+
+			//> elm_size = ids.elm_size
+			elm_size, err := hinter.ResolveAsFelt(vm, elm_size)
+			if err != nil {
+				return err
+			}
+			//> assert isinstance(elm_size, int) and elm_size > 0, \
+			//> 	f'Invalid value for elm_size. Got: {elm_size}.'
+			if elm_size.IsZero() {
```
cicr99, line 79: Again here, just to be safe, I'd use `FeltLe` with the zero value instead (we have the felt for zero in the utils/contants file you can use)

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,119 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+
+			elementMaxSize, err := ctx.ScopeManager.GetVariableValue("__find_element_max_size")
+			if err != nil {
+				return err
+			}
+
+			elementMaxSizeFelt, ok := elementMaxSize.(*fp.Element)
+			if !ok {
+				return err
+			}
```
cicr99, line 103: Just for a better understanding of the code, could you move the comment below this lines up? I feel that this logic belongs to that part as well

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,119 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+
+			elementMaxSize, err := ctx.ScopeManager.GetVariableValue("__find_element_max_size")
+			if err != nil {
+				return err
+			}
+
+			elementMaxSizeFelt, ok := elementMaxSize.(*fp.Element)
+			if !ok {
+				return err
```
cicr99, line 102: You should return a custom error here, something like "expected __find_element_max_size to be a felt".
Notice that at this point in the code `err` is `nil` from line 96, so you wouldn't actually know why your program failed

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,119 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+			}
+
+			indexAddr, err := index.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+
+			//> for i in range(n_elms):
+			for i := uint64(0); i < n_elms.Uint64(); i++ {
```
cicr99, line 125: `n_elems` is a felt so in theory it could be as big as to have 252 bits, but it doesn't make sense for it to not fit in uint64, or we wouldn't be able to compute it. So it's ok if we cast it to uint64, but I would like you to do it separately and return an error if it's not possible to do so.

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,119 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+	return newSearchSortedLowerHint(array_ptr, elm_size, n_elms, key, index), nil
+}
+
+func newSearchSortedLowerHint(array_ptr, elm_size, n_elms, key, index hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "SearchSortedLower",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+
+			//> array_ptr = ids.array_ptr
+			array_ptr, err := hinter.ResolveAsAddress(vm, array_ptr)
```
cicr99, line 67: Can we stick to camelCase for variable names?
You'll have to fix this and the rest of the variables to use the same format

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,119 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+			}
+
+			indexAddr, err := index.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+
+			//> for i in range(n_elms):
+			for i := uint64(0); i < n_elms.Uint64(); i++ {
+				array_ptr.AddOffset(int16(elm_size.Uint64()) * int16(i))
```
cicr99, line 126: Notice that this operation modifies the value of `array_ptr`, which means for the next iteration the result of this operation won't be correct anymore. I see 2 approaches here:
1- create a newArrayPtr each time with the result of this operation without modifying arrayPtr at all.
2- You can modify arrayPtr by adding `elm_size` in each iteration. this should yield the correct result each time. Notice here the first iteration i = 0, so nothing is added to arrayPtr. I would update this variable at the end of the loop.

I personally like the second approach better so you don't have to create extra variables.

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,119 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+				arrayPtrValue := memory.MemoryValueFromMemoryAddress(array_ptr)
+				arrayValue, err := arrayPtrValue.FieldElement()
+
+				if err != nil {
+					return err
+				}
+
+				fe := arrayValue
+				//> 	if memory[array_ptr + elm_size * i] >= ids.key:
+				if utils.FeltLt(key, fe) {
```
cicr99, line 136: You can just use `arrayValue` directly, no need to save it in `fe` variable

cicr99, line 136: the condition is `memory[array_ptr + elm_size * i] >= ids.key` so it's actually `FeltLe` the function you need to use

pkg/hintrunner/zero/zerohint_memcpy.go
```diff
@@ -28,3 +33,119 @@ func createVMExitScopeHinter(resolver hintReferenceResolver) (hinter.Hinter, err
...
+					//> 		ids.index = i
+					//> 		break
+					indexValue := memory.MemoryValueFromFieldElement((new(fp.Element).SetInt64(int64(i))))
```
cicr99, line 139: You could use `SetUint64` directly so you don't have to make the conversion for `i` where you do `SetInt64(int64(i))`

cicr99 (changes requested):

TAdev0:
Hi @joseSalazar4 , you have unit tests failing, can you make them pass before a review?

pkg/hintrunner/zero/zerohint_memcpy_test.go
```diff
@@ -21,6 +22,116 @@ func TestZeroHintMemcpy(t *testing.T) {
...
+				operanders: []*hintOperander{
+					{Name: "array.el0", Kind: apRelative, Value: feltInt64(0)},
+					{Name: "array.el1", Kind: apRelative, Value: feltInt64(1)},
+					{Name: "array.el2", Kind: apRelative, Value: feltInt64(2)},
+					{Name: "array.el3", Kind: apRelative, Value: feltInt64(3)},
+					{Name: "arrayPtr", Kind: apRelative, Value: addr(0)},
+					{Name: "elmSize", Kind: fpRelative, Value: feltInt64(4)},
+					{Name: "nElms", Kind: fpRelative, Value: feltInt64(9)},
+					{Name: "index", Kind: fpRelative, Value: feltInt64(3)},
+					{Name: "key", Kind: fpRelative, Value: feltInt64(0)},
```
cicr99, line 119: I believe these values don't make a lot of sense.
- `nElms` should be the amount of elements in the array
- `index` should have been `uninitialized` kind, because this is something that will be assigned during the execution of the hint. In fact, instead of checking you get a nil error after running the test, you should check the value that was saved in `index` address. This variable will represent the index of the first element of the array that is bigger or equal to the value of `key`
- `key` should have the value you are looking for. You can play with this in the tests, for example if you have array [0, 1, 2, 3] as in this case, apart from trying simple cases, you can try something edge case like `key = 5`. Since all elements of the array are smaller then the else statement after the for should be executed getting `index = 4`.

A suggestion is to try an array with different numbers than the ones you are using so the values of the elements don't coincide with their actual indexes. Notice the array has to be sorted in increasing order

cicr99 (changes requested):

pkg/hintrunner/zero/zerohint_memcpy_test.go
```diff
@@ -21,6 +22,116 @@ func TestZeroHintMemcpy(t *testing.T) {
...
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newSearchSortedLowerHint(
+						ctx.operanders["arrayPtr"],
+						ctx.operanders["elmSize"],
+						ctx.operanders["nElms"],
+						ctx.operanders["key"],
+						ctx.operanders["index"],
+					)
+				},
+				errCheck: errorTextContains("failed obtaining the variable: __find_element_max_size"),
```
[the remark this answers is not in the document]
joseSalazar4: The code checks if the var exists on the scope it does not check the value so do you think this testcase is needed?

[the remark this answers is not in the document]
cicr99: This specific case shouldn't be necessary, as if the value is not in scope then it shouldn't even fail. But just was letting you know that if you need to work with a specific value for that variable then you need to assign it in the scope through ctxInit field

rodrigo-pino:
Hey @joseSalazar4 how is it going. Do you need help with anything?

TAdev0:
hi @joseSalazar4  are you still working on this?

PR #377, opened 2024-04-29
Implement cairo0 compute slope V1 hint

Tomi-3-0:
#### ComputeSlopeV1Hint #257 
This hint calculates the slope between two points on an elliptic curve. A helper function `LineSlope()` is used for calculating the slope.

Test cases for this hint have been compared with the results produced by the  VM implementations in Rust and Go.
LineSlope method link: https://github.com/lambdaclass/cairo-vm_in_go/blob/31c3628bc10ebc1628685b3cdfa72d0e938b533e/pkg/builtins/ec_op.go#L258

### DIFF
--- a/pkg/hintrunner/utils/math_utils.go
+++ b/pkg/hintrunner/utils/math_utils.go
@@ -26,6 +26,23 @@ func EcDoubleSlope(pointX, pointY, alpha, prime *big.Int) (big.Int, error) {
 	return Divmod(n, m, prime)
 }
 
+func LineSlope(point_aX, point_aY, point_bX, point_bY, prime *big.Int) (big.Int, error) {
+	// belongs to the line_slope method
+	modValue := new(big.Int).Mod(new(big.Int).Sub(point_aX, point_aY), prime)
+
+	if modValue.Cmp(big.NewInt(0)) == 0 {
+		return *big.NewInt(0), errors.New("the slope of the line is invalid")
+	}
+
+	// Compute the difference of y-coordinates
+	n := new(big.Int).Sub(point_bX, point_bY)
+
+	// Compute the difference of x-coordinates
+	m := new(big.Int).Sub(point_aX, point_aY)
+
+	return Divmod(n, m, prime)
+}
+
 func AsInt(valueFelt *fp.Element) big.Int {
 	// https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/math_utils.py#L8
 
--- a/pkg/hintrunner/utils/math_utils.go
+++ b/pkg/hintrunner/utils/math_utils.go
@@ -27,7 +27,6 @@ func EcDoubleSlope(pointX, pointY, alpha, prime *big.Int) (big.Int, error) {
 }
 
 func LineSlope(point_aX, point_aY, point_bX, point_bY, prime *big.Int) (big.Int, error) {
-	// belongs to the line_slope method
 	modValue := new(big.Int).Mod(new(big.Int).Sub(point_aX, point_aY), prime)
 
 	if modValue.Cmp(big.NewInt(0)) == 0 {
--- a/pkg/hintrunner/utils/math_utils.go
+++ b/pkg/hintrunner/utils/math_utils.go
@@ -27,19 +27,19 @@ func EcDoubleSlope(pointX, pointY, alpha, prime *big.Int) (big.Int, error) {
 }
 
 func LineSlope(point_aX, point_aY, point_bX, point_bY, prime *big.Int) (big.Int, error) {
-	// https://github.com/lambdaclass/cairo-vm_in_go/blob/31c3628bc10ebc1628685b3cdfa72d0e938b533e/pkg/builtins/ec_op.go#L258
+	// https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/python/math_utils.py#L130
 
-	modValue := new(big.Int).Mod(new(big.Int).Sub(point_aX, point_aY), prime)
+	modValue := new(big.Int).Mod(new(big.Int).Sub(point_aX, point_bX), prime)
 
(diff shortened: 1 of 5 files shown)


pkg/hintrunner/utils/math_utils.go
```diff
@@ -26,6 +26,23 @@ func EcDoubleSlope(pointX, pointY, alpha, prime *big.Int) (big.Int, error) {
 	return Divmod(n, m, prime)
 }
 
+func LineSlope(point_aX, point_aY, point_bX, point_bY, prime *big.Int) (big.Int, error) {
+	// belongs to the line_slope method
```
MaksymMalicki, line 30: could you provide a link for this functionality as in AsInt example?

MaksymMalicki:
At first look everything looks amazing, good job 🚀

pkg/hintrunner/utils/math_utils.go
Tomi-3-0: Hi @MaksymMalicki , here it is https://github.com/lambdaclass/cairo-vm_in_go/blob/31c3628bc10ebc1628685b3cdfa72d0e938b533e/pkg/builtins/ec_op.go#L258

MaksymMalicki: I mean can you paste it in the code, so that the reader knows where does this functionality come from. Thanks!

Tomi-3-0: Okay

pkg/hintrunner/utils/math_utils.go
```diff
@@ -26,6 +26,24 @@ func EcDoubleSlope(pointX, pointY, alpha, prime *big.Int) (big.Int, error) {
 	return Divmod(n, m, prime)
 }
 
+func LineSlope(point_aX, point_aY, point_bX, point_bY, prime *big.Int) (big.Int, error) {
+	// https://github.com/lambdaclass/cairo-vm_in_go/blob/31c3628bc10ebc1628685b3cdfa72d0e938b533e/pkg/builtins/ec_op.go#L258
```
cicr99, line 30: I'd rather use the link of the function in the python VM, this is supposed to be implemented from there.
It would be // https://github.com/lambdaclass/cairo-vm_in_go/blob/31c3628bc10ebc1628685b3cdfa72d0e938b533e/pkg/builtins/ec_op.go#L258

pkg/hintrunner/utils/math_utils.go
```diff
@@ -26,6 +26,24 @@ func EcDoubleSlope(pointX, pointY, alpha, prime *big.Int) (big.Int, error) {
...
+
+	modValue := new(big.Int).Mod(new(big.Int).Sub(point_aX, point_aY), prime)
```
cicr99, line 32: I think the subtraction here is meant to be done between `point_aX` and `point_bX` according to this [line](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/python/math_utils.py#L135C5-L135C44):
`assert (point1[0] - point2[0]) % p != 0`

pkg/hintrunner/utils/math_utils.go
```diff
@@ -26,6 +26,24 @@ func EcDoubleSlope(pointX, pointY, alpha, prime *big.Int) (big.Int, error) {
...
+
+	if modValue.Cmp(big.NewInt(0)) == 0 {
+		return *big.NewInt(0), errors.New("the slope of the line is invalid")
+	}
+
+	// Compute the difference of y-coordinates
+	n := new(big.Int).Sub(point_bX, point_bY)
+
+	// Compute the difference of x-coordinates
+	m := new(big.Int).Sub(point_aX, point_aY)
```
cicr99, line 42: Same over here, I think you mistook the operands in the subtraction :sweat_smile: 
According to this [line](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/python/math_utils.py#L136) I think the correct subtractions should be `point_aY - point_bY` and `point_aX - point_bX`

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -593,5 +593,58 @@ func TestZeroHintEc(t *testing.T) {
...
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newComputeSlopeV1Hint(ctx.operanders["point0.x.d0"], ctx.operanders["point1.x.d0"])
+				},
+				errCheck: errorTextContains("the slope of the line is invalid"),
+			},
+			{
+				operanders: []*hintOperander{
+					// random values
+					{Name: "point0.x.d0", Kind: apRelative, Value: feltString("134")},
```
cicr99, line 623: Instead of using `feltString` I would use `feltUIn64` or `feltInt64`, or the fixed values from the utils/constant file if possible, it should be more efficient this way. I would leave feltString for values that cannot fit in 64 bits

pkg/hintrunner/zero/zerohint_ec_test.go
cicr99, line 1: I'm wondering, how are you generating the test cases? If the comments I previously wrote are correct, then the expected results of these tests shouldn't be the correct ones. For getting the expected result for tests you should either do the calculation yourself using the methods from the python vm, or create a cairo file example that uses this hint and print the result, so you know what is to be expected. Does this make sense?

cicr99 (changes requested):

pkg/hintrunner/utils/math_utils.go
Tomi-3-0: okay

pkg/hintrunner/zero/zerohint_ec_test.go
Tomi-3-0: The results and test values were indeed cross-validated with both the Python VM and LambdaClass's VM implementation. The absence of test failures seem to stem from the manner in which I was passing the variables into the `LineSlope()` function, effectively doing the same maths calculations but in a different manner.

cicr99 (approved):

PR #392, opened 2024-05-07
feat: Run separate cairo integration test files as a separate Go test

### LINKED ISSUE #201: Run separate cairo integration test files as a separate Go test
Doing that would allow us to run a specific file test to skip through the ones that are not relevant to the local changes.

xiaolou86:
Resolves #201 

First checking if the environment variable `INTEGRATION_TESTS_FILTERS` is set in the console.
And if not, then load it from the .env file.

### DIFF
--- a/integration_tests/.env
+++ b/integration_tests/.env
@@ -1,0 +1,2 @@
+# Set to run a specific file test (ex. fib.cairo)
+INTEGRATION_TESTS_FILTER=
--- a/integration_tests/.env
+++ b/integration_tests/.env
@@ -1,2 +1,2 @@
-# Set to run a specific file test (ex. fib.cairo)
-INTEGRATION_TESTS_FILTER=
+# Set to run some specific file tests (ex. fib.cairo,alloc.cairo)
+INTEGRATION_TESTS_FILTERS=
--- a/integration_tests/cairozero_test.go
+++ b/integration_tests/cairozero_test.go
@@ -27,7 +27,9 @@ func TestCairoZeroFiles(t *testing.T) {
 		t.Errorf("Error loading .env file: %v", err)
 	}
 
-	filter := os.Getenv("INTEGRATION_TESTS_FILTER")
+	filtersRaw := os.Getenv("INTEGRATION_TESTS_FILTERS")
+	filtersRaw = strings.TrimSpace(filtersRaw)
+	filters := strings.Split(filtersRaw, ",")
 
 	for _, dirEntry := range testFiles {
 		if dirEntry.IsDir() || isGeneratedFile(dirEntry.Name()) {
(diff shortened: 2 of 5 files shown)


xiaolou86:
@rodrigo-pino @quasilyte @cicr99 Please help me to review. Thank you very much.

MaksymMalicki:
Hey, I left a comment on the issue. Can you please check it?

xiaolou86:
@MaksymMalicki I have made this improvement. Thanks for your suggestion.

integration_tests/.env
```diff
@@ -0,0 +1,2 @@
+# Set to run some specific file tests (ex. fib.cairo,alloc.cairo)
+INTEGRATION_TESTS_FILTERS=
```
rodrigo-pino, line 2: This file is more of a utility, I like the idea that I have this option but I don't think it should be part of the repo

integration_tests/cairozero_test.go
```diff
@@ -19,8 +20,16 @@ func TestCairoZeroFiles(t *testing.T) {
...
 	// filter is for debugging purposes
-	filter := ""
+	err = godotenv.Load("./.env")
+	if err != nil {
+		t.Errorf("Error loading .env file: %v", err)
+	}
+
+	filtersRaw := os.Getenv("INTEGRATION_TESTS_FILTERS")
+	filtersRaw = strings.TrimSpace(filtersRaw)
+	filters := strings.Split(filtersRaw, ",")
```
rodrigo-pino, line 32: I think it is better for it not to err in case the is no .env and instead defaults to an empty string. 

Also, if it could be nice if I could just could set the env from the console, like

```
INTEGRATION_TEST_FILTERS=alloc make integration
```

rodrigo-pino (commented):
Great, I hadn't thought about having multiple filters. I think is a nice addition. Just leaving some minor comments

integration_tests/cairozero_test.go
xiaolou86: So is it better to read from the console, instead of .env file?

rodrigo-pino: Both would be the ideal solutions. With the console taking priority over the .env file

xiaolou86: I have achieved that by first checking if the environment variable is set in the console, and if not, then load it from the .env file. @rodrigo-pino

integration_tests/.env
xiaolou86: I have achieved that by first checking if the environment variable is set in the console, and if not, then load it from the .env file. @rodrigo-pino

rodrigo-pino (commented):
It's great as it is. If you want to improve the PR a little more there are two things we can do:
1. Given the current code it might be nice to refactor into an object `Filter` which will have two functions:
    -  `new` where it will initialize the filter
    - `filter` where it filters the code
2. Add a section to the readme explaining how to use filters

You can do one or two, or none. Just let me know. In case of none, I'll approve immediately.

xiaolou86:
@rodrigo-pino I have refactored into an object Filter. Thank you very much for your suggestion and review.

rodrigo-pino (approved):
Great!

TAdev0:
@xiaolou86  there is a l'intime g error:

Error:   Error return value of `godotenv.Load` is not checked (errcheck)

Can you resolve this?

xiaolou86:
> Error: Error return value of `godotenv.Load` is not checked (errcheck)

@TAdev0 Done. Thank you very much.

TAdev0 (approved):

TAdev0:
@xiaolou86  i think the ideal would be to really check the return value, not just doing `_   = .. ` , something like:

```
    err := godotenv.Load("./.env")

    if err != nil {
        ... // throw an error
    }
    ```
@rodrigo-pino ? isnt it weird the linting issue is resolved while doing `_ = ...` doesnt really resolve the initial issue `Error: Error return value of godotenv.Load is not checked (errcheck)` ?

Issue #395, opened 2024-05-08
Different memory file generated compared to other VMs


har777:
The same compiled cairo file seems to be generating different memory files depending on the VM used.

The comparisons are made between the original python vm, lambdaclass rust vm and our go vm. I'm using shasum to check if the files match. The steps to reproduce this is at the end of the issue. I'm not sure if I'm testing something wrong or if there is indeed a bug.

Examples:

1. rust vm and nethermind vm match but python vm is different:
```
from starkware.cairo.common.alloc import alloc

func main() {
  alloc_locals;

  let (local p) = alloc();

  ret;
}

>> comparing generated memory files:
601af58bee65974f8a742a35684bde3929ccb5329da5cdcf1522bff3ad0aba75  lambdaworks_rust_memory
601af58bee65974f8a742a35684bde3929ccb5329da5cdcf1522bff3ad0aba75  nethermind_memory
6ff36c06517cc549f8a9c728e544425f47871462a7abff7f1060d593a3a74348  starkware_memory
```

```
func get_z() -> (z: felt) {
    return (z=123);
}

func main() {
    alloc_locals;
    let (local z) = get_z();

    return ();
}

4e11c3548b4d158bd12c773315b456be191c8d188a2de6d9b6a6e8fb7af9ee9f  lambdaworks_rust_memory
4e11c3548b4d158bd12c773315b456be191c8d188a2de6d9b6a6e8fb7af9ee9f  nethermind_memory
7704c8a11392ea8c1b73fd7bb5e4b7ab9a3a95fdd5067aad740f90e40dc11010  starkware_memory

```

2. rust vm and python vm match but nethermind vm is different:
```
%builtins output

func main{output_ptr: felt*}() {
    return ();
}


>> comparing generated memory files:
59ec4a9c715e8ad336ec953fb06279c107fde622759910669c198b4670064df9  lambdaworks_rust_memory
44c1ed850fbb5e96974ad639ab45b1d4bd71b9c31a3f697176ab8fdef003cdc2  nethermind_memory
59ec4a9c715e8ad336ec953fb06279c107fde622759910669c198b4670064df9  starkware_memory
```

3. all 3 vm's differ:
```
%builtins range_check

from starkware.cairo.common.alloc import alloc
from starkware.cairo.common.uint256 import Uint256
from starkware.cairo.common.cairo_blake2s.blake2s import blake2s_add_uint256

func main{range_check_ptr: felt}() {

[...]

PR #397, opened 2024-05-09
Fix how lastMemIndex is calculated in DecodeMemory

har777:
Previously the max memory index in a memory file was calculated assuming that the last address in the file would represent the highest index.

```
// calculate the max memory index
lastContentInd := len(content) - (addrSize + feltSize)
lastMemIndex := binary.LittleEndian.Uint64(content[lastContentInd : lastContentInd+addrSize])
```

But this is not always the case. The memory file generated by the python vm can have the addresses in the file written out of order. That means the last address in the file need not be the largest address.

This PR changes that logic. Now the code will iterate over all the addresses in the memory file and find the largest one before initialising the decoded memory array and populating it.

Related to https://github.com/NethermindEth/cairo-vm-go/issues/395

### DIFF
--- a/pkg/vm/vm.go
+++ b/pkg/vm/vm.go
@@ -590,7 +590,7 @@ func DecodeTrace(content []byte) []Trace {
 const addrSize = 8
 const feltSize = 32
 
-// Encody the relocated memory in the (address, value) form
+// Encode the relocated memory in the (address, value) form
 // in a consecutive way
 func EncodeMemory(memory []*f.Element) []byte {
 	// Check non nil elements for optimal array size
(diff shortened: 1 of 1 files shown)


pkg/vm/vm.go
```diff
@@ -624,14 +624,23 @@ func EncodeMemory(memory []*f.Element) []byte {
...
 	// calculate the max memory index
-	lastContentInd := len(content) - (addrSize + feltSize)
-	lasMemIndex := binary.LittleEndian.Uint64(content[lastContentInd : lastContentInd+addrSize])
+	lastMemIndex := uint64(0)
+	for i := 0; i < len(content); i += addrSize + feltSize {
+		memIndex := binary.LittleEndian.Uint64(content[i : i+addrSize])
+		if memIndex > lastMemIndex {
+			lastMemIndex = memIndex
+		}
+	}
```
rodrigo-pino, line 638: Could you add a small comment explaining the reason why we are using the for to scan for the whole memory

pkg/vm/vm.go
```diff
@@ -624,14 +624,23 @@ func EncodeMemory(memory []*f.Element) []byte {
 }
 
 func DecodeMemory(content []byte) []*f.Element {
```
rodrigo-pino, line 626: Could you add a function description, along the same lines of all the function descriptions we have been doing

rodrigo-pino (commented):
Great work @har777! Just left some small nitpicks.

rodrigo-pino (approved):

PR #398, opened 2024-05-09
[bug]: implement correctly `UsortVerifyMultiplicityBody` hint

### LINKED ISSUE #391: `UsortVerifyMultiplicityBody` hint is wrongly implemented
Python hint : 
```
//> current_pos = positions.pop()
//> ids.next_item_index = current_pos - last_pos
//> last_pos = current_pos + 1
```

Current go hint: 

```
newCurrentPos, err := utils.Pop(&positions)
if err != nil {
    return err
}

currentPos, err := ctx.ScopeManager.GetVariableValue("current_pos")
if err != nil {
    return err
}

currentPosInt, ok := currentPos.(int64)
if !ok {
    return fmt.Errorf("cannot cast current_pos to int64")
}

lastPos, err := ctx.ScopeManager.GetVariableValue("last_pos")
if err != nil {
    return err
}

lastPosInt, ok := lastPos.(int64)
if !ok {
    return fmt.Errorf("cannot cast last_pos to int64")
}

// Calculate `next_item_index` memory value
newNextItemIndexValue := currentPosInt - lastPosInt
```

`currentPos, err := ctx.ScopeManager.GetVariableValue("current_pos")` is problematic , and should not exist. `current_pos` is calculated with `utils.Pop(&positions)` and there are only `positions` and `last_pos` taht should be fetched from the scope.

Also: 

```
return ctx.ScopeManager.AssignVariables(map[string]any{
	"current_pos": newCurrentPos,
	"last_pos":    int64(currentPosInt + 1),
}
```
will write in scope current_pos while it shouldn't. `current_pos` is calculated with `positions.pop()`  for each iteration in the recursive call to `verify_multiplicity`

https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/usort.cairo#L99C12-L99C31

TAdev0:
Resolves #391 

This PR rectifies `UsortVerifyMultiplicityBody` hint previously wrongly implemented.
I also added more tests, always with a descending list as input (required). 

note that `current_pos` does not need to be assigned in scope , i also rectified this. `current_scope` is calculated for each iteration with `positions.pop()` and is not required in scope.
Also, i added a reassignement of `positions` in the scope because it was missing, `positions` needs to be updated in scope otherwise the array will ne be emptied in the recursive call, this will cause an infinite loop when calling usort

### DIFF
--- a/pkg/hintrunner/zero/zerohint_usort.go
+++ b/pkg/hintrunner/zero/zerohint_usort.go
@@ -195,7 +195,7 @@ func newUsortVerifyMultiplicityBodyHint(nextItemIndex hinter.ResOperander) hinte
 				return err
 			}
 
-			return ctx.ScopeManager.AssignVariable("last_pos", currentPos.Add(&currentPos, &utils.FeltOne))
+			return ctx.ScopeManager.AssignVariable("last_pos", *currentPos.Add(&currentPos, &utils.FeltOne))
 		},
 	}
 }
--- a/pkg/hintrunner/zero/zerohint_usort.go
+++ b/pkg/hintrunner/zero/zerohint_usort.go
@@ -154,9 +154,9 @@ func newUsortVerifyMultiplicityBodyHint(nextItemIndex hinter.ResOperander) hinte
 				return err
 			}
 
-			positions, ok := positionsInterface.([]int64)
+			positions, ok := positionsInterface.([]fp.Element)
 			if !ok {
-				return fmt.Errorf("cannot cast positionsInterface to []int64")
+				return fmt.Errorf("cannot cast positionsInterface to []fp.Element")
 			}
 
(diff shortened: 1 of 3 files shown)


TAdev0:
not finished:

```
current_pos = positions.pop()
```

still need to reassign in scope `positions` after popping

MaksymMalicki:
This looks legit, the utility that you have mentioned is neccessary. I think you can do it in this PR as it is close to merging.

TAdev0:
@MaksymMalicki  comments addressed! should be good to go
I used again Reflect library in test, this time for `case []int64` ( `[] fp.Element` in SquashDictInnerNextKey PR)

rodrigo-pino (approved):

PR #400, opened 2024-05-09
feat: SquashDictInnerNextKey hint

### LINKED ISSUE #303: SquashDictInnerNextKey: Retrieves the next key for processing during dictionary squashing.

TAdev0:
Resolves #303 

if we want tests to be really good, and check the value in the scope of the array of field elements `keys` after popping, i guess we should add a Case in `varValueInScopeEquals` for arrays of felts.

### DIFF
--- a/pkg/hintrunner/zero/zerohint_dictionaries.go
+++ b/pkg/hintrunner/zero/zerohint_dictionaries.go
@@ -63,7 +63,7 @@ func newSquashDictInnerNextKeyHint(nextKey hinter.ResOperander) hinter.Hinter {
 				return err
 			}
 
-			err = ctx.ScopeManager.AssignVariable("keys", keys)
+			err = ctx.ScopeManager.AssignVariable("key", &newKey)
 			if err != nil {
 				return err
 			}
--- a/pkg/hintrunner/zero/zerohint_dictionaries.go
+++ b/pkg/hintrunner/zero/zerohint_dictionaries.go
@@ -68,7 +68,7 @@ func newSquashDictInnerNextKeyHint(nextKey hinter.ResOperander) hinter.Hinter {
 				return err
 			}
 
-			err = ctx.ScopeManager.AssignVariable("key", &newKey)
+			err = ctx.ScopeManager.AssignVariable("key", newKey)
 			if err != nil {
 				return err
 			}
--- a/pkg/hintrunner/zero/zerohint_dictionaries.go
+++ b/pkg/hintrunner/zero/zerohint_dictionaries.go
@@ -63,6 +63,11 @@ func newSquashDictInnerNextKeyHint(nextKey hinter.ResOperander) hinter.Hinter {
 				return err
 			}
 
+			err = ctx.ScopeManager.AssignVariable("keys", keys)
+			if err != nil {
+				return err
+			}
+
 			err = ctx.ScopeManager.AssignVariable("key", &newKey)
 			if err != nil {
 				return err
--- a/pkg/hintrunner/zero/zerohint_dictionaries.go
+++ b/pkg/hintrunner/zero/zerohint_dictionaries.go
@@ -63,7 +63,10 @@ func newSquashDictInnerNextKeyHint(nextKey hinter.ResOperander) hinter.Hinter {
 				return err
 			}
 
-			ctx.ScopeManager.AssignVariable("keys", keys)
+			err = ctx.ScopeManager.AssignVariable("keys", keys)
+			if err != nil {
+				return err
+			}
 
 			newKeyMemoryValue := memory.MemoryValueFromFieldElement(&newKey)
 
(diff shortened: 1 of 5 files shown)


pkg/hintrunner/zero/zerohint_dictionaries_test.go
```diff
@@ -37,5 +37,37 @@ func TestZeroHintDictionaries(t *testing.T) {
...
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					err := ctx.ScopeManager.AssignVariable("keys", []fp.Element{*feltUint64(1), *feltUint64(2), *feltUint64(3)})
+					if err != nil {
+						t.Fatal(err)
+					}
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newSquashDictInnerNextKeyHint(ctx.operanders["next_key"])
+				},
+				check: varValueEquals("next_key", feltUint64(3)),
```
MaksymMalicki, line 69: we have to check for the value of keys in the scope, so I thinks its a good idea to add new utility

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -33,3 +35,56 @@ func newSquashDictInnerAssertLenKeysHint() hinter.Hinter {
...
+// dictionary squashing after checking that the array of keys is not empty
+//
+// `newSquashDictInnerNextKeyHint` takes 1 operander as argument
+//   - `next_key` variable will be assigned to the next key in `keys`
+func newSquashDictInnerNextKeyHint(nextKey hinter.ResOperander) hinter.Hinter {
+	return &GenericZeroHinter{
+		Name: "SquashDictInnerNextKey",
+		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			//> assert len(keys) > 0, 'No keys left but remaining_accesses > 0.'
+			//> ids.next_key = key = keys.pop()
```
MaksymMalicki, line 49: I'm wondering if adding the updated value of ```key``` to the scope isn't neccessary, as [this hint](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/squash_dict.cairo#L206) uses it as key in the access_indices structure. Since ```squash_dict_inner``` is invoked recursively it might be neccessary.

TAdev0: of `keys` you mean? yeah its necessary because of the recursive call, which will empty the array one element after the other. If we dont, we'll get an infinite loop

pkg/hintrunner/zero/zerohint_dictionaries_test.go
TAdev0: yes thats what i supposed! thanks

pkg/hintrunner/zero/zerohint_dictionaries.go
TAdev0: oh yeah!!!!! sorry, `key` indeed, is used where you mentioned, i need to add it to the scope! thanks a lot!!!!

TAdev0: nice catch!

TAdev0:
@MaksymMalicki  comments addressed.
- now both `key` and `keys` are reassigned in scope, and `ids.next_key` in memory is `key` 
- Created a case for `[] fp.Element` in `varValueInScopeEquals` that uses an external library `reflect` . Code is straightforward,  i think its ok? its for tests anyway

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -33,3 +35,61 @@ func newSquashDictInnerAssertLenKeysHint() hinter.Hinter {
...
+
+			newKey, err := utils.Pop(&keys)
+			if err != nil {
+				return err
+			}
+
+			err = ctx.ScopeManager.AssignVariable("keys", keys)
+			if err != nil {
+				return err
+			}
```
rodrigo-pino, line 69: In case you're unaware, this type of patter can be merged in Go with the following syntax:

```
if err := ctx.ScopeManager.Smth(...); err != nil {
    return err
}
```

Just another way of writing it, you pick whatever you like the most

rodrigo-pino (commented):
All looks good to me. Let's wait to merge Hari's PR and we add an integration test for this as well

rodrigo-pino:
> All looks good to me. Let's wait to merge Hari's PR and we add an integration test for this as well

As @TAdev0 pointed out, it doesn't make sense to have integration test for just one hint, since they are multiple hints involved

pkg/hintrunner/zero/zerohint_dictionaries.go
TAdev0: thanks!

rodrigo-pino (approved):

PR #402, opened 2024-05-13
Implement SetAdd hint

### LINKED ISSUE #246: SetAdd
Adds an element to a set if it's not already present, and determines its index and presence status.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/set.cairo#L34

har777:
Fixes #246 

Will move from draft to open PR once builtin's issue is fixed in a different PR.


pkg/hintrunner/zero/zerohint_others_test.go
```diff
@@ -58,5 +58,128 @@ func TestZeroHintMemcpy(t *testing.T) {
...
+					{Name: "set.4", Kind: apRelative, Value: feltUint64(8)},
+					{Name: "set.5", Kind: apRelative, Value: feltUint64(9)},
+					{Name: "set.6", Kind: apRelative, Value: feltUint64(10)},
+					{Name: "set.7", Kind: apRelative, Value: feltUint64(11)},
+					{Name: "set.8", Kind: apRelative, Value: feltUint64(12)},
+					{Name: "set.9", Kind: apRelative, Value: feltUint64(1)},
+					{Name: "set.10", Kind: apRelative, Value: feltUint64(2)},
+					{Name: "set.11", Kind: apRelative, Value: feltUint64(3)},
+					{Name: "set.12", Kind: apRelative, Value: feltUint64(4)},
+					{Name: "set.9", Kind: apRelative, Value: feltUint64(13)},
```
MaksymMalicki, line 80: nitpick: names out of order

pkg/hintrunner/zero/zerohint_others_test.go
```diff
@@ -58,5 +58,128 @@ func TestZeroHintMemcpy(t *testing.T) {
...
+					)
+				},
+				check: allVarValueEquals(map[string]*fp.Element{
+					"index":         feltUint64(0),
+					"is_elm_in_set": feltUint64(1),
+				}),
+			},
+		},
 	})
 }
```
MaksymMalicki, line 185: maybe add tests for assertions? I mean this:
```        assert ids.elm_size > 0
        assert ids.set_ptr <= ids.set_end_ptr```

File: pkg/hintrunner/zero/zerohint_others.go [hunk 3]
```diff
@@ -150,3 +151,126 @@ func createMemEnterScopeHinter(resolver hintReferenceResolver, memset bool) (hin
...
+			//> elm_list = memory.get_range(ids.elm_ptr, ids.elm_size)
+			//> for i in range(0, ids.set_end_ptr - ids.set_ptr, ids.elm_size):
+			//>     if memory.get_range(ids.set_ptr + i, ids.elm_size) == elm_list:
+			//>         ids.index = i // ids.elm_size
+			//>         ids.is_elm_in_set = 1
+			//>         break
+			//>     else:
+			//>         ids.is_elm_in_set = 0
+
+			elmSize, err := hinter.ResolveAsFelt(vm, elmSize)
```
MaksymMalicki, line 170: I think this could be resolved as uint64 right away. You do it later in the code, an the elmSize practically should never exceed uint64 range

pkg/hintrunner/zero/zerohint_others_test.go
```diff
@@ -58,5 +58,170 @@ func TestZeroHintMemcpy(t *testing.T) {
...
+						ctx.operanders["elm_size"],
+						ctx.operanders["elm_ptr"],
+						ctx.operanders["set_ptr"],
+						ctx.operanders["set_end_ptr"],
+						ctx.operanders["index"],
+						ctx.operanders["is_elm_in_set"],
+					)
+				},
+				check: allVarValueEquals(map[string]*fp.Element{
+					"is_elm_in_set": feltUint64(0),
```
MaksymMalicki, line 188: add check for index uninitialised as well?

pkg/hintrunner/zero/zerohint_others.go
har777: Good idea!

pkg/hintrunner/zero/zerohint_others_test.go
har777: Hmm how do you check that using `varValueEquals`?

cicr99: If you try reading from a place in memory that hasn't been initialized, I think it should treat this as an unknown memory value, but I also believe it should error out. Perhaps you could try accessing that address and see what happens.
Anyway, I don't think adding this check is that important

cicr99 (approved):

MaksymMalicki (approved):

MaksymMalicki:
Btw this has some conflicts

PR #406, opened 2024-05-13
feat: `SquashDictInnerContinueLoop` hint

### LINKED ISSUE #299: SquashDictInnerContinueLoop
Determines if the loop should continue based on remaining access indices.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/squash_dict.cairo#L194

Depends on #245

TAdev0:
Resolves #299 

This PR implements `SquashDictInnerContinueLoop` hint, which checks if the squashing loop should continue or not, depending on whether there are access indices remaining or not.

It writes 1 or 0 in `should_continue` field of `loop_temps` struct variable

Here is a the `LoopTemps` struct :
```

   struct LoopTemps {
        index_delta_minus1: felt,
        index_delta: felt,
        ptr_delta: felt,
        should_continue: felt,
    }
```

I added offset of 4 to write on the 4th item of the struct, as we first store the number of items and then each item contiguously in memory

### DIFF
--- a/pkg/hintrunner/zero/zerohint_dictionaries.go
+++ b/pkg/hintrunner/zero/zerohint_dictionaries.go
@@ -55,6 +55,16 @@ func newSquashDictInnerContinueLoopHint(loopTemps hinter.ResOperander) hinter.Hi
 				return err
 			}
 
+			currentAccessIndices := currentAccessIndices_.([]f.Element)
+			if len(currentAccessIndices) != 0 {
+				return fmt.Errorf("assertion `len(current_access_indices) == 0` failed")
+			}
+
+			loopTemps, err := hinter.ResolveAsUint64(vm, loopTemps)
+			if err != nil {
+				return err
+			}
+
 			keys := keys_.([]f.Element)
 			if len(keys) == 0 {
 				return fmt.Errorf("no keys left but remaining_accesses > 0")
--- a/pkg/hintrunner/zero/zerohint_dictionaries.go
+++ b/pkg/hintrunner/zero/zerohint_dictionaries.go
@@ -7,6 +7,7 @@ import (
 	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
 	VM "github.com/NethermindEth/cairo-vm-go/pkg/vm"
 	"github.com/NethermindEth/cairo-vm-go/pkg/vm/memory"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 	f "github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
@@ -55,44 +56,27 @@ func newSquashDictInnerContinueLoopHint(loopTemps hinter.ResOperander) hinter.Hi
 				return err
 			}
 
-			currentAccessIndices := currentAccessIndices_.([]f.Element)
-			if len(currentAccessIndices) != 0 {
-				return fmt.Errorf("assertion `len(current_access_indices) == 0` failed")
-			}
+			currentAccessIndices := currentAccessIndices_.([]fp.Element)
 
-			loopTemps, err := hinter.ResolveAsUint64(vm, loopTemps)
+			loopTempsAddr, err := loopTemps.GetAddress(vm)
 			if err != nil {
 				return err
 			}
 
-			keys := keys_.([]f.Element)
-			if len(keys) == 0 {
-				return fmt.Errorf("no keys left but remaining_accesses > 0")
-			}
-
-			newKey, err := utils.Pop(&keys)
+			resultAddr, err := loopTempsAddr.AddOffset(4)
 			if err != nil {
 				return err
 			}
 
-			err = ctx.ScopeManager.AssignVariable("keys", keys)
-			if err != nil {
-				return err
+			resultMemZero := memory.MemoryValueFromFieldElement(&utils.FeltZero)
+			resultMemOne := memory.MemoryValueFromFieldElement(&utils.FeltOne)
+
+			if len(currentAccessIndices) == 0 {
+				return vm.Memory.WriteToAddress(&resultAddr, &resultMemZero)
+
+			} else {
+				return vm.Memory.WriteToAddress(&resultAddr, &resultMemOne)
 			}
-
-			err = ctx.ScopeManager.AssignVariable("key", newKey)
-			if err != nil {
-				return err
-			}
-
-			newKeyMemoryValue := memory.MemoryValueFromFieldElement(&newKey)
-
-			addrNextKey, err := nextKey.GetAddress(vm)
-			if err != nil {
-				return err
-			}
-
-			return vm.Memory.WriteToAddress(&addrNextKey, &newKeyMemoryValue)
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -39,6 +40,59 @@ func createSquashDictInnerAssertLenKeysHinter() (hinter.Hinter, error) {
...
+				return err
+			}
+
+			resultAddr, err := loopTempsAddr.AddOffset(4)
+			if err != nil {
+				return err
+			}
+
+			resultMemZero := memory.MemoryValueFromFieldElement(&utils.FeltZero)
+			resultMemOne := memory.MemoryValueFromFieldElement(&utils.FeltOne)
```
har777, line 75: Why not add in the corresponding `if` statement?

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -39,6 +40,59 @@ func createSquashDictInnerAssertLenKeysHinter() (hinter.Hinter, error) {
...
+			}
+
+			currentAccessIndices := currentAccessIndices_.([]fp.Element)
+
+			loopTempsAddr, err := loopTemps.GetAddress(vm)
+			if err != nil {
+				return err
+			}
+
+			resultAddr, err := loopTempsAddr.AddOffset(4)
```
har777, line 69: Are you sure about this?

https://docs.cairo-lang.org/reference/syntax.html#structs seems to indicate the first field is at offset 0?

pkg/hintrunner/zero/zerohint_dictionaries.go
TAdev0: yeah true, my bad

pkg/hintrunner/zero/zerohint_dictionaries.go
TAdev0: i tried with 3 at first spontaneously, but i changed after testing:

```
			operanders: []*hintOperander{
					{Name: "loop_temps", Kind: fpRelative, Value: addr(7)},
					{Name: "loop_temps.index_delta_minus1", Kind: apRelative, Value: feltInt64(0)},
					{Name: "loop_temps.index_delta", Kind: apRelative, Value: feltInt64(0)},
					{Name: "loop_temps.ptr_delta", Kind: apRelative, Value: feltInt64(0)},
					{Name: "loop_temps.should_continue", Kind: uninitialized},
				},
```

this seems a correct setup of all fields of the struct in the correct order. If i change for `3` or `5` in the offset, i get the logic corresponding error, only `4` works. I guess the length is necessarly stored at offset 0, it makes sense also. But i'm not sure about that

har777: You should test with integration tests. Unit tests are just checking how you wrote the code right 🤔 

eg: if you change to `AddOffset(3)` and make make "loop_temps" point to the first value in the struct wont it work?

har777: @MaksymMalicki what do you think?

TAdev0: @har777  yeah sure, my idea was to debate around this, merge, and do integration when all dict hints are implemented. Do you mean doing intermediary integration tests?

yeah sure, but tests might also help to understand the memory /storage setup  or many things. Here, the code of the hint is very basic, so it might make sense to try things with test to understand how things are stored under the hood, no?

har777: Ah I meant that if you are encoding a struct incorrectly as inputs in unit test then the test itself might not help.

TAdev0: btw @har777  there is something i don't understand:
for this hint i can do 
```
{Name: "loop_temps", Kind: apRelative, Value: addr(1)},
```
and it will work.

but with other hints, its often impossible to use an address < 6 , because the 5 first cells are initialized at some point when doing unit test of the hint.
Do you know why?

har777: So you aren't actually using the value of "loop_temps". You are just getting the address of the variable "loop_temps". And then adding an offset to it and getting the actual field you require. Try changing the value to some uint value and it should still work.

TAdev0: oh yeah its clear thanks!

TAdev0: the fact that it works when it is either `apRelative` of `fpRelative` seems normal for you?

TAdev0: so you think it should be
```
		{Name: "loop_temps", Kind: fpRelative, Value: feltInt64(0)},
					{Name: "loop_temps", Kind: apRelative, Value: feltInt64(0)},
					{Name: "loop_temps.ptr_delta", Kind: apRelative, Value: feltInt64(0)},
					{Name: "loop_temps.should_continue", Kind: uninitialized},
```

with `loop_temps` being implicitly the first field of the struct?

har777: Yeah something like:

```
{Name: "loop_temps.index_delta_minus1", Kind: apRelative, Value: feltInt64(0)},
{Name: "loop_temps.index_delta", Kind: apRelative, Value: feltInt64(0)},
{Name: "loop_temps.ptr_delta", Kind: apRelative, Value: feltInt64(0)},
{Name: "loop_temps.should_continue", Kind: uninitialized},
```

And you do:
```
return newSquashDictInnerContinueLoopHint(ctx.operanders["loop_temps.index_delta_minus1"])
```

But let's clarify this with others! 

Btw do the first 3 values even need to be initialised?

har777: Also you can use the `WriteToNthStructField` method too!

TAdev0: no indeed, they dont!

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -39,6 +40,58 @@ func createSquashDictInnerAssertLenKeysHinter() (hinter.Hinter, error) {
...
+			}
+
+			if len(currentAccessIndices) == 0 {
+				resultMemZero := memory.MemoryValueFromFieldElement(&utils.FeltZero)
+				return vm.Memory.WriteToAddress(&resultAddr, &resultMemZero)
+
+			} else {
+				resultMemOne := memory.MemoryValueFromFieldElement(&utils.FeltOne)
+				return vm.Memory.WriteToAddress(&resultAddr, &resultMemOne)
+			}
```
har777, line 81: Try using `WriteToNthStructField` instead.

TAdev0:
@har777  comments addressed. Its nice to use the `WriteToNthStructField` wrapper, it reduces code size. thanks

har777 (approved):

rodrigo-pino (approved):

PR #378, opened 2024-04-29
Implement hints documentation

Iwueseiter:
Resolves #327

PR #409, opened 2024-05-13
Docs: hints docs

Iwueseiter:
Implemented Hints Documentation.

### DIFF
--- a/docs/docs/vm-fundamentals/hints.md
+++ b/docs/docs/vm-fundamentals/hints.md
@@ -2,6 +2,137 @@
 sidebar_position: 3
 ---
 
-# Hints
 
-There are several hints, explain each of them here. How they interact with the VM and affects it.
+# Hint
+  ## HINTS  IN CAIRO VIRTUAL MACHINE
+Hints in Cairo are pieces of Python code only executed by the sequencer.
+
+Hints are used to help the Virtual Machine determine the next path of execution it should take.
+
+Below are some Cairo code with hints:
+
+```cairo
+      [ap] = 25, ap++;
+    %{
+        import math
+         memory[ap] = int(math.sqrt(memory[ap - 1]))
+     %}
+      [ap - 1] = [ap] * [ap], ap++;
+```
+
+
+```cairo
+      func main() {
+          %{ memory[ap] = program_input['secret'] %}
+         [ap] = [ap], ap++;
+         ret;
+      }
+``` 
+
+
+
+### WHY HINTS RUN ON SEQUENCERS/ PROVERS AND NOT VERIFIERS
+
+Sequencers are the leads that produces blocks by executing transaction and updating the blockchain state.
+During the proof generation process, the prover uses hints to generate additional constraints or assumptions about the program's behavior, which makes part of the proof sent to the verifier for verification. In the Cairo Virtual Machine, hints are sorted out during different stages of the compilation and execution process by the sequencer. However, they are not directly involved in the verification stage. Hints are primarily for performance optimization.
+    
+The key is the implementation of relevant and accurate hints that assist the prover or sequencers in their respective tasks.
+
+
+
+Example:
+Let's consider a Cairo function that calculates the factorial of a given input *`n`*:
+
+```Cairo
+    @public
+   func factorial(n : felt) -> (result : felt):
+    if n <= 1:
+        result := 1
+    else:
+     result := n * factorial(n - 1)
+```
+
+In this code, the hint *@public* indicates that the *`n`* factorial should be publicly available in the Cairo program.
+The provers/sequencers will utilize the hint to generate a proof or sequence of instructions for the function to ensure maximum optimization. The verifier will check that the generated proof or sequence of instructions has correctly calculated the factorial of the input *`n`* or any other specification is satisfied.
+
+## IMPLEMENTING HINTS IN CAIRO VIRTUAL MACHINE
+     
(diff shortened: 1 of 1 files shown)


docs/docs/vm-fundamentals/hints.md
```diff
@@ -2,6 +2,137 @@
...
+         ret;
+      }
+``` 
+
+
+
+### WHY HINTS RUN ON SEQUENCERS/ PROVERS AND NOT VERIFIERS
+
+Sequencers are the leads that produces blocks by executing transaction and updating the blockchain state.
+During the proof generation process, the prover uses hints to generate additional constraints or assumptions about the program's behavior, which makes part of the proof sent to the verifier for verification. In the Cairo Virtual Machine, hints are sorted out during different stages of the compilation and execution process by the sequencer. However, they are not directly involved in the verification stage. Hints are primarily for performance optimization.
```
rodrigo-pino, line 37: During proof generating, hints are not used at all. Hints are only used when generating the execution trace. This execution trace is then proven by the Prover but at this stage it doesn't care about hints. Again, please don't use information you're not certain of. 

Let's remove this sentence.

docs/docs/vm-fundamentals/hints.md
```diff
@@ -2,6 +2,137 @@
...
+    @public
+   func factorial(n : felt) -> (result : felt):
+    if n <= 1:
+        result := 1
+    else:
+     result := n * factorial(n - 1)
+```
+
+In this code, the hint *@public* indicates that the *`n`* factorial should be publicly available in the Cairo program.
+The provers/sequencers will utilize the hint to generate a proof or sequence of instructions for the function to ensure maximum optimization. The verifier will check that the generated proof or sequence of instructions has correctly calculated the factorial of the input *`n`* or any other specification is satisfied.
```
rodrigo-pino, line 56: This is very False. Please, I am hesitant to continue reviewing the PR if you are really not invested in it. Knowing this is false just requires reading the Cairo Docs. Also, writing this simple Cairo Program you'll notice this is untrue.

This kind of content seems AI hallucination, the least you could do is double check if what is being generated is true. Wasting the reviewers time is not a good practice.

docs/docs/vm-fundamentals/hints.md
```diff
@@ -2,6 +2,137 @@
...
+           entrypoint,
+           [2, (2,0)],
+           False,
+           vm,
+           hint_processor
+        )[0];
+    %}
+      [ap - 1] = [ap] * [ap], ap++;
+
+```
```
rodrigo-pino, line 92: Is this Cairo? Don't you see it's different from the other examples?

docs/docs/vm-fundamentals/hints.md
```diff
@@ -2,6 +2,137 @@
...
+    else:
+     result := n * factorial(n - 1)
+```
+
+In this code, the hint *@public* indicates that the *`n`* factorial should be publicly available in the Cairo program.
+The provers/sequencers will utilize the hint to generate a proof or sequence of instructions for the function to ensure maximum optimization. The verifier will check that the generated proof or sequence of instructions has correctly calculated the factorial of the input *`n`* or any other specification is satisfied.
+
+## IMPLEMENTING HINTS IN CAIRO VIRTUAL MACHINE
+     
+ In Cairo Virtual Machine, hints are accepted and recognized by its code.
```
rodrigo-pino, line 60: I don't understand what does this sentence means

docs/docs/vm-fundamentals/hints.md
```diff
@@ -2,6 +2,137 @@
...
+```
+
+In this code, the hint *@public* indicates that the *`n`* factorial should be publicly available in the Cairo program.
+The provers/sequencers will utilize the hint to generate a proof or sequence of instructions for the function to ensure maximum optimization. The verifier will check that the generated proof or sequence of instructions has correctly calculated the factorial of the input *`n`* or any other specification is satisfied.
+
+## IMPLEMENTING HINTS IN CAIRO VIRTUAL MACHINE
+     
+ In Cairo Virtual Machine, hints are accepted and recognized by its code. 
+Hints are implemented using the *%{...%}* syntax, which allows you to insert Python code that is only executed by the prover. When implementing hints, the *'variable naming convention'* should be considered. The method to handle specific hints should be outlined like this; *'create hintName Hinter'*. Hints should be grouped by functionality in Cairo programs. The structure returned should implement the interface. Implementing the hint means you have to write code in Go that has the same behavior as the code of the hint in Python. Unit tests are added to check if the behavior of your code is correct.
```
rodrigo-pino, line 61: This sentence is not true. Let's delete it.

rodrigo-pino (commented):
Please, refrain from submitting anymore commits until you've studied the topic. Writing documentation is sometimes even more difficult than writing code because it requires an understanding of the subject.

We don't expect to have the best piece of written content, nobody here is a writer. We DO expect for that content to be correct, at the very least.

If your approach is to write things and get them fact checked by us, so you can fix them and then repeat the until completion, then that is not a very good approach. You are wasting the reviewers time and yours because you're not learning anything. I am seeing it here as well as in the recently closed #378 

I don't know if this is intentional or not, if you keep this quality of work, we won't accept anymore PRs from you. If you're here to learn then please go read the Docs and come here when you feel ready. Otherwise, let's just close this PR now as unsolved.

rodrigo-pino:
Talked a bit more with @Iwueseiter on Telegram. We decided to close the PR for now.

PR #396, opened 2024-05-08
refactor: move GetConsecutiveValues functionality to memory package

MoigeMatino:
## What does this PR do?

- Moves the `GetConsecutiveValues` functionality from hinter package to memory package.
- Makes necessary changes to files that invoked `GetConsecutiveValues` to match new function location.

### Related Issue(s)?
- Resolves #384

PR #421, opened 2024-05-20
doc: rename GetConsecutiveValues to GetConsecutiveMemoryValues in Readme

### LINKED ISSUE #419: README.md needs to be updated after `GetConsecutiveValues` has been moved to memory package and renamed `GetConsecutiveMemoryValues`
Update the README.md to reflect the new definition of GetConsecutiveMemoryValues from GetConsecutiveValues as shown here in the current documentation:

> The operanders allow you to get the address in memory of the variable you need to use. When this variable is a felt just reading the value from that address is enough, but sometimes the variable is actually a struct so you need to get several consecutive values in memory, for that you can use `GetConsecutiveValues(vm *VM.VirtualMachine, ref ResOperander, size int16)` method. For example in `uint256add` hint you need to add `a` and `b` which are both `Uint256` structs, so we used `GetUint256AsFelts` utility which basically calls that function with size 2.

MoigeMatino:
### What does this PR do?

- Updates zero package Readme by renaming `GetConsecutiveValues` to `GetConsecutiveMemoryValues`. 

### Reason for change

- To accurately reference the correct method for retrieving consecutive memory values.

### Related issue(s)?

- Resolves #419 
- As a result of :  #396


pkg/hintrunner/zero/README.md
```diff
@@ -49,7 +49,7 @@ func newAssert250bitsHint(low, high, value hinter.ResOperander) hinter.Hinter {
...
-- The operanders allow you to get the address in memory of the variable you need to use. When this variable is a felt just reading the value from that address is enough, but sometimes the variable is actually a struct so you need to get several consecutive values in memory, for that you can use `GetConsecutiveValues(vm *VM.VirtualMachine, ref ResOperander, size int16)` method. For example in `uint256add` hint you need to add `a` and `b` which are both `Uint256` structs, so we used `GetUint256AsFelts` utility which basically calls that function with size 2.
+- The operanders allow you to get the address in memory of the variable you need to use. When this variable is a felt just reading the value from that address is enough, but sometimes the variable is actually a struct so you need to get several consecutive values in memory, for that you can use `GetConsecutiveMemoryValues(addr MemoryAddress, size int16)` method. For example in `uint256add` hint you need to add `a` and `b` which are both `Uint256` structs, so we used `GetUint256AsFelts` utility which basically calls that function with size 2.
```
TAdev0, line 52: Actually, `GetUint256AsFelts` indeed uses `ReadFromAddress` 2 times to read the low and high parts of the uint256 in memory. `GetConsecutiveMemoryValues` also uses `ReadFromAddress` , inside a loop.

You should rework this section, explaining that they both use `ReadFromAddress`.
`GetUint256AsFelts` doesn't use `GetConsecutiveMemoryValues`  so the current version is wrong

TAdev0 (changes requested):

pkg/hintrunner/zero/README.md
MoigeMatino: Okay, changes addressed.

pkg/hintrunner/zero/README.md
```diff
@@ -49,7 +49,8 @@ func newAssert250bitsHint(low, high, value hinter.ResOperander) hinter.Hinter {
...
+- The operanders allow you to get the address in memory of the variable you need to use. When this variable is a `felt`, simply reading the value from that address is sufficient. However, if the variable is a `struct`, you need to retrieve several consecutive values from memory. For this, you can use the `GetConsecutiveMemoryValues(addr MemoryAddress, size int16)` method. 
+For example, in the `uint256add` hint, you need to add `a` and `b`, which are both `Uint256` structs. To achieve this, we use the `GetUint256AsFelts` utility, which calls `ReadFromAddress` twice to read the low and high parts of the `uint256` in memory. It's important to note that `GetUint256AsFelts` does not use `GetConsecutiveMemoryValues`; instead, it directly uses `ReadFromAddress` two times. Both `GetUint256AsFelts` and `GetConsecutiveMemoryValues` utilize `ReadFromAddress`, but in different ways: the former calls it twice directly, while the latter uses it inside a loop to read multiple consecutive values.
```
TAdev0, line 53: > two times

twice

MoigeMatino:
Hi @TAdev0 , made the changes!

pkg/hintrunner/zero/README.md
```diff
@@ -49,7 +49,8 @@ func newAssert250bitsHint(low, high, value hinter.ResOperander) hinter.Hinter {
...
-- The operanders allow you to get the address in memory of the variable you need to use. When this variable is a felt just reading the value from that address is enough, but sometimes the variable is actually a struct so you need to get several consecutive values in memory, for that you can use `GetConsecutiveMemoryValues(addr MemoryAddress, size int16)` method. For example in `uint256add` hint you need to add `a` and `b` which are both `Uint256` structs, so we used `GetUint256AsFelts` utility which basically calls that function with size 2.
+- The operanders allow you to get the address in memory of the variable you need to use. When this variable is a `felt`, simply reading the value from that address is sufficient. However, if the variable is a `struct`, you need to retrieve several consecutive values from memory. For this, you can use the `GetConsecutiveMemoryValues(addr MemoryAddress, size int16)` method.
```
TAdev0, line 52: > For this, you can use the `GetConsecutiveMemoryValues(addr MemoryAddress, size int16)` method.

For this, you can use the `GetConsecutiveMemoryValues(addr MemoryAddress, size int16)` method, which calls `ReadFromAddress` function inside a loop.

I feel like the flow is not good. We talk about `GetConsecutiveMemoryValues` , and then, we say '' for example '' to explain that `uint256add` doesnt use `GetConsecutiveMemoryValues` .

Can you reformulate in some way ?

TAdev0 (changes requested):

pkg/hintrunner/zero/README.md
MoigeMatino: 🤔Let me see what I can do

TAdev0:
@MoigeMatino  any update ?

TAdev0 (approved):
LGTM!

MaksymMalicki (approved):

PR #424, opened 2024-05-22
Implement `DictSquashCopyDict` hint

### LINKED ISSUE #292: DictSquashCopyDict
Prepares arguments for creating a new dictionary by copying the existing one in the context of squashing.

Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/dict.cairo#L87

Depends on #245

TAdev0:
This PR implements `DictSquashCopyDict` hint, which prepares arguments for creating a new dictionary with dict_new function. It copies the existing squashed dictionary in the  current scope

Closes #292


TAdev0:
@har777  i'm not 100% sure i'm doing write this python code
```
dict(__dict_manager.get_dict(ids.dict_accesses_end))
```

could you check this?
I guess this code only retrieves from operanders the address `dict_accesses_end` , uses `GetDictionary` to retrieve the dictionary (a copy) and assign it in scope?

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -125,6 +127,62 @@ func createDictReadHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+			if err != nil {
+				return err
+			}
+
+			dictionaryCopy, err := dictionaryManager.GetDictionary(*dictAccessesEnd_)
+			if err != nil {
+				return err
+			}
+
+			ctx.ScopeManager.EnterScope(map[string]any{"__dict_manager": dictionaryManager, "initial_dict": dictionaryCopy})
```
har777, line 170: Just to confirm:
1. the same dictionary manager in the current scope must be available in the child scope. 
2. a copy of the `dict_accesses_end` from the same dictionary manager must be added to the child scope.

In this case is `dictionaryCopy` creating a new copied object? 🤔

har777: We should add some tests for this too right?

MaksymMalicki: I agree, that at least one test, which checks if the provided values are in scope is neccessary.

MaksymMalicki:
could you resolve conflits? It will be easier to review ✅

pkg/hintrunner/zero/zerohint_dictionaries.go
TAdev0: yeah i'll add tests i was waiting to do it

TAdev0: @har777  you're right, i need to make a deep copy of `dictionaryCopy` (to be renamed `dictionary` ) 
for that, i need to modify a lot of code because i need to do it manually, which means accessing the fields of a `MemoryValue` which is another package and currently only lowercases and therefore not accessible

TAdev0:
@har777  @MaksymMalicki  just updated, now making a deep copy of the ZeroDict before assign it in the new scope.
I added only one test for now with an empty dictionaryManager, will add other tests tomorrow

TAdev0:
currently `FreeOffset` field of the ZeroDict struct is of type `*uint64` , any reason?

TAdev0:
@har777  ready for review, i think this is the last dict hint

TAdev0:
i guess we should use `FreeOffset uint64` in `ZeroDictionary`, there is no benefit of using a pointer here, we can modify this later

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -191,6 +194,67 @@ func createDictReadHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+
+			dictionary, err := dictionaryManager.GetDictionary(*dictAccessesEnd_)
+			if err != nil {
+				return err
+			}
+
+			dictionaryCopy, err := hinter.CopyZeroDictionary(&dictionary)
+			if err != nil {
+				return err
+			}
```
har777, line 240: You only need to copy the data. Not the entire `ZeroDictionary`. This variable is used in the `DictNew` hint in this same file. So just a copy of the `Data` field. Maybe some method like `dictionaryManager.getDataCopy(dictPtr)`?

har777: Or you can just create a copy of it it inline in this hint. Its not really used anywhere else.

TAdev0: done!

TAdev0:
@har777  now only the dict Data is copied and I inlined the copy

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -191,6 +194,80 @@ func createDictReadHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
...
+
+			dictionaryManager_, err := ctx.ScopeManager.GetVariableValue("__dict_manager")
+			if err != nil {
+				return err
+			}
+
+			dictionaryManager, ok := dictionaryManager_.(hinter.ZeroDictionaryManager)
+			if !ok {
+				return fmt.Errorf("cannot cast dictionaryManager_ to a DictionaryManager struct")
+			}
```
cicr99, line 225: I think there is a utility for this, `GetZeroDictionaryManager` or something like that

cicr99 (approved):

pkg/hintrunner/zero/zerohint_dictionaries.go
TAdev0: what do you mean? i'm just type casting here

TAdev0: @cicr99  just addressed your comment, thanks :)

pkg/hintrunner/zero/zerohint.go
```diff
@@ -187,6 +187,8 @@ func GetHintFromCode(program *zero.ZeroProgram, rawHint zero.Hint, hintPC uint64
 		return createDefaultDictNewHinter(resolver)
 	case dictReadCode:
 		return createDictReadHinter(resolver)
+	case dictSquashCopyDict:
```
har777, line 190: End the variable with "Code"

TAdev0: thanks !

pkg/hintrunner/zero/zerohint_dictionaries_test.go
```diff
@@ -807,5 +807,22 @@ func TestZeroHintDictionaries(t *testing.T) {
 				},
 			},
 		},
+		"DictSquashCopyDict": {
+			{
+				operanders: []*hintOperander{
+					{Name: "dict_accesses_end", Kind: apRelative, Value: addr(5)},
```
har777, line 813: You should add a test for successful operation too right?

PR #425, opened 2024-05-22
chore: improve fmt and use pointer in zerodict

TAdev0:
This PR improves _zero_dict.go_ to use pointers to manipulate structs. 
It also improves fmt , removes using `f` renamed  for `fp` import  for consistency across codebase


TAdev0:
blocked until all dict hints are merged

PR #433, opened 2024-05-27
Implement MemsetEnterScope

har777:
Fixes #430

PR #434, opened 2024-05-27
Implement MemsetContinueLoop

### LINKED ISSUE #431: MemsetContinueLoop
Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/memset.cairo#L22

har777:
Fixes #431

### DIFF
--- a/pkg/hintrunner/zero/zerohint_others.go
+++ b/pkg/hintrunner/zero/zerohint_others.go
@@ -10,10 +10,21 @@ import (
 	f "github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
-func newMemcpyContinueCopyingHint(continueCopying hinter.ResOperander) hinter.Hinter {
+func newMemContinueHint(continueTarget hinter.ResOperander, memset bool) hinter.Hinter {
+	var name string
+	if memset {
+		name = "MemsetContinueLoop"
+	} else {
+		name = "MemcpyContinueCopying"
+	}
 	return &GenericZeroHinter{
-		Name: "MemcpyContinueCopying",
+		Name: name,
 		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			// MemsetContinueLoop
+			//> n -= 1
+			//> ids.continue_loop = 1 if n > 0 else 0
+
+			// MemcpyContinueCopying
 			//> n -= 1
 			//> ids.continue_copying = 1 if n > 0 else 0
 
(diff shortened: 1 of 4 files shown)


pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -30,30 +41,36 @@ func newMemcpyContinueCopyingHint(continueCopying hinter.ResOperander) hinter.Hi
...
-func createMemcpyContinueCopyingHinter(resolver hintReferenceResolver) (hinter.Hinter, error) {
-	continueCopying, err := resolver.GetResOperander("continue_copying")
+func createMemContinueHinter(resolver hintReferenceResolver, memset bool) (hinter.Hinter, error) {
+	var continueTarget hinter.ResOperander
+	var err error
+	if memset {
+		continueTarget, err = resolver.GetResOperander("continue_loop")
+	} else {
+		continueTarget, err = resolver.GetResOperander("continue_copying")
+	}
```
cicr99, line 69: Is it better to pass as a parameter to this function the name of the variable instead of a boolean? You could directly fetch this variable from the resolver and you wouldn't have to do the if statement, wdyt? 🤔

pkg/hintrunner/zero/zerohint_others_test.go
```diff
@@ -38,14 +38,50 @@ func TestZeroHintMemcpy(t *testing.T) {
...
-					return newMemcpyContinueCopyingHint(ctx.operanders["continue_copying"])
+					return newMemContinueHint(ctx.operanders["continue_copying"], false)
 				},
 				check: func(t *testing.T, ctx *hintTestContext) {
 					allVarValueInScopeEquals(map[string]any{"n": feltInt64(4)})(t, ctx)
 					varValueEquals("continue_copying", feltInt64(1))(t, ctx)
 				},
 			},
 		},
+		"MemsetContinueLoop": {
```
cicr99, line 49: I don't think we need to add new tests for this hint. Notice that the only difference in the hint codes is the name of the variable they are using, which is resolved in the createHinter method. Hence, this is not reflected in the unit tests, as you pass directly the resOperander to the newHint method. At the end, you are not testing anything new here.
Only if you came up with a different test, that actually checks a new edge case, then you can leave that

cicr99 (commented):
I think these comments also apply to #433. I'll leave a note there so we don't forget to check

pkg/hintrunner/zero/zerohint_others.go
har777: Also the `newMemContinueHint` function needs to calculate the name of the hint so some sort of `if` condition is still required!

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -10,10 +10,21 @@ import (
...
+		name = "MemcpyContinueCopying"
+	}
 	return &GenericZeroHinter{
-		Name: "MemcpyContinueCopying",
+		Name: name,
 		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
+			// MemsetContinueLoop
+			//> n -= 1
+			//> ids.continue_loop = 1 if n > 0 else 0
+
```
TAdev0, line 26: i think the formatting standard here is to add `//>` also for empty line

TAdev0, line 26: i think the formatting standard here is to add `//>` also for empty line

TAdev0 (approved):
lgtm!
left a minor comment

pkg/hintrunner/zero/zerohint_others.go
cicr99: We do add that if the empty line belongs to the code of the hint, but do not think that's necessary here where you want to create an actual separation.
The reason we use `//>` instead of simple comment lines is because the hint code could be confused in some cases with actual go code from the vm and so it doesn't appear to be just commented code from the codebase

TAdev0: Oh yeah you re right my bad :p

cicr99 (approved):

PR #438, opened 2024-06-03
Implement `searchSortedLower` hint

### LINKED ISSUE #253: SearchSortedLower: Searches for the first element in a sorted array that is greater than or equal to a given key and returns its index.

TAdev0:
Closes #253


TAdev0:
Still need to improve unit test coverage

TAdev0:
@cicr99  @MaksymMalicki @har777  for this one, we need to evaluate the accuracy of executing this check 

```
//> if '__find_element_max_size' in globals():
			//> 	assert n_elms <= __find_element_max_size, \
			//> 		f'find_element() can only be used with n_elms<={__find_element_max_size}. ' \
			//> 		f'Got: n_elms={n_elms}.'
```

this hint is used in one function, thats all. `__find_element_max_size` is set to `2**20` in the execution of the python VM.
I think we should just hardcode it to 2**20 in scope and remove this check

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -144,3 +147,146 @@ func createMemEnterScopeHinter(resolver hintReferenceResolver, memset bool) (hin
...
+			if err != nil {
+				return err
+			}
+
+			nElemsRange := nElms.Uint64()
+
+			//> for i in range(n_elms):
+			for i := uint64(0); i < nElemsRange; i++ {
+				//> 	if memory[array_ptr + elm_size * i] >= ids.key:
+				index := arrayPtr + (elmSize * i)
```
cicr99, line 241: If you take arrayPtr as an address as mentioned above, then you'll probably need to change the Add operation here.
Btw, one idea here. Instead of calculating the multiplication each time, you could instead update the value of index in each iteration by adding elmSize to it. You'd have to start `index = arrPtr` and do the update at the end of the loop so it has the correct value for the next iteration. Wdyt?

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -144,3 +147,146 @@ func createMemEnterScopeHinter(resolver hintReferenceResolver, memset bool) (hin
...
+
+				value, err := vm.Memory.ReadAsElement(uint64(1), index)
```
cicr99, line 243: Why is the segment index set to 1?

pkg/hintrunner/zero/zerohint_others_test.go
```diff
@@ -57,5 +57,246 @@ func TestZeroHintMemcpy(t *testing.T) {
...
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newSearchSortedLowerHint(
+						ctx.operanders["array_ptr"],
+						ctx.operanders["elm_size"],
+						ctx.operanders["n_elms"],
+						ctx.operanders["key"],
+						ctx.operanders["index"],
+					)
+				},
+				errCheck: errorIsNil,
```
cicr99, line 98: Shouldn't you be checking the value of `index`?

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -144,3 +147,146 @@ func createMemEnterScopeHinter(resolver hintReferenceResolver, memset bool) (hin
 	}
 	return newMemEnterScopeHint(value, memset), nil
 }
+
+// SearchSortedLower hint searches for the first element in a sorted array
+// that is greater than or equal to a given key and returns its index
+//
+// `newSearchSortedLowerHint` takes 5 operanders as arguments
+//   - `arrayPtr` represents the offset in the execution segment of memory where starts the sorted array
```
MaksymMalicki, line 155: I'm also not sure about this information. I think it is an address to the base of the input array

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -144,3 +147,146 @@ func createMemEnterScopeHinter(resolver hintReferenceResolver, memset bool) (hin
...
+			//> assert isinstance(elm_size, int) and elm_size > 0, \
+			//> 	f'Invalid value for elm_size. Got: {elm_size}.'
+			if elmSize == 0 {
+				return fmt.Errorf("invalid value for elm_size. Got: %v", elmSize)
+			}
+
+			//> n_elms = ids.n_elms
+			//> assert isinstance(n_elms, int) and n_elms >= 0, \
+			//> 	f'Invalid value for n_elms. Got: {n_elms}.'
+			nElms, err := hinter.ResolveAsFelt(vm, nElms)
```
MaksymMalicki, line 205: I think this can directly be resolved as uint64

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -34,8 +36,8 @@ func newMemContinueHint(continueTarget hinter.ResOperander, memset bool) hinter.
 				return err
 			}
 
-			newN := new(f.Element)
-			newN = newN.Sub(n.(*f.Element), &utils.FeltOne)
+			newN := new(fp.Element)
+			newN = newN.Sub(n.(*fp.Element), &utils.FeltOne)
```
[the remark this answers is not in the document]
TAdev0: done

pkg/hintrunner/zero/zerohint_others.go
TAdev0: true, updated

pkg/hintrunner/zero/zerohint_others.go
TAdev0: updated! better like this for sure, thanks!

pkg/hintrunner/zero/zerohint_others_test.go
TAdev0: yeah, but we cannot have a `check` and `errCheck` at the same time, thats why. I made a bunch of others tests testing the value of `index` , the hint works properly

TAdev0:
@cicr99  @MaksymMalicki  PR updated, addressed all comments, should be good to go!

MaksymMalicki (approved):

cicr99 (approved):

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -150,3 +151,142 @@ func createMemEnterScopeHinter(resolver hintReferenceResolver, memset bool) (hin
...
+				//> 	if memory[array_ptr + elm_size * i] >= ids.key:
+				value, err := vm.Memory.ReadFromAddressAsElement(index)
+				if err != nil {
+					return err
+				}
+
+				if utils.FeltLe(key, &value) {
+					//> 		ids.index = i
+					//> 		break
+					indexValue := memory.MemoryValueFromFieldElement((new(fp.Element).SetUint64(i)))
```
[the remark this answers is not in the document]
TAdev0: oh yeah true thanks !

Issue #436, opened 2024-06-03
`MemcpyContinueCopying` hint is wrongly implemented


TAdev0:
the hint is 

```
	//> n -= 1
        //> ids.continue_copying = 1 if n > 0 else 0
```

`n` is a value stored in the scope.

Currently we do

```
		newN := new(f.Element)
			newN = newN.Sub(n.(*f.Element), &utils.FeltOne)

			if err := ctx.ScopeManager.AssignVariable("n", newN); err != nil {
				return err
			}
```

this is not the expected behavior as we store a pointer to `n`, instead of `n` in scope

danielcdz:
Hello @TAdev0! I worked on adding that hint, I think I have enough context to solve the issue, so I will be happy to help!

TAdev0:
Hey @danielcdz , someone else is already working on it!

PR #439, opened 2024-06-04
Fixed implementation of MemcpyContinueCopying

Sh0g0-1758:
Fixes: #436

### DIFF
--- a/pkg/hintrunner/zero/zerohint_others.go
+++ b/pkg/hintrunner/zero/zerohint_others.go
@@ -41,7 +41,7 @@ func newMemContinueHint(continueTarget hinter.ResOperander, memset bool) hinter.
 				return fmt.Errorf("casting n into a felt failed")
 			}
 
-			newN = *new(fp.Element).Sub(&newN, &utils.FeltOne)
+			newN.Sub(&newN, &utils.FeltOne)
 
 			if err := ctx.ScopeManager.AssignVariable("n", newN); err != nil {
 				return err
--- a/pkg/hintrunner/zero/zerohint_others.go
+++ b/pkg/hintrunner/zero/zerohint_others.go
@@ -40,7 +40,7 @@ func newMemContinueHint(continueTarget hinter.ResOperander, memset bool) hinter.
 			if !ok {
 				return fmt.Errorf("casting n into a felt failed")
 			}
-			
+
 			newN = *new(fp.Element).Sub(&newN, &utils.FeltOne)
 
 			if err := ctx.ScopeManager.AssignVariable("n", newN); err != nil {
--- a/pkg/hintrunner/zero/zerohint_others.go
+++ b/pkg/hintrunner/zero/zerohint_others.go
@@ -1,13 +1,15 @@
 package zero
 
 import (
+	"fmt"
+
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/core"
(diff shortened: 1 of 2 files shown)


pkg/hintrunner/zero/zerohint_others_test.go
```diff
@@ -33,15 +33,16 @@ func TestZeroHintMemcpy(t *testing.T) {
 					{Name: "continue_copying", Kind: uninitialized},
 				},
 				ctxInit: func(ctx *hinter.HintRunnerContext) {
-					ctx.ScopeManager.EnterScope(map[string]any{
-						"n": feltString("5"),
-					})
+					err := ctx.ScopeManager.AssignVariable("n", *feltString("5"))
```
TAdev0, line 36: i guess you can use `feltInt64` or `feltUint64` instead of `feltString` here. Generally we use feltString for big numbers that dont feat in a uint64
it doesnt really matter but its better style and its more consistent with the rest of our code

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -34,8 +36,12 @@ func newMemContinueHint(continueTarget hinter.ResOperander, memset bool) hinter.
 				return err
 			}
 
-			newN := new(f.Element)
-			newN = newN.Sub(n.(*f.Element), &utils.FeltOne)
+			newN, ok := n.(fp.Element)
+			if !ok {
+				return fmt.Errorf("casting n into a felt failed")
+			}
+
```
TAdev0, line 43: i think its more straightforward to do :

```
newN, ok := n.(fp.Element)
	if !ok {
		return fmt.Errorf("casting n into a felt failed")
	}

newN.Sub(&newN, &utils.FeltOne)
```

TAdev0 (changes requested):
Hey @Sh0g0-1758 

left 2 comments, but it looks really good overall :)

pkg/hintrunner/zero/zerohint_others_test.go
Sh0g0-1758: Right, Made the changes.

pkg/hintrunner/zero/zerohint_others.go
Sh0g0-1758: Right, Modified it to that.

TAdev0 (approved):
LGTM!

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -34,8 +36,12 @@ func newMemContinueHint(continueTarget hinter.ResOperander, memset bool) hinter.
...
+			newN.Sub(&newN, &utils.FeltOne)
 
 			if err := ctx.ScopeManager.AssignVariable("n", newN); err != nil {
```
har777, line 46: Can you change this to `ctx.ScopeManager.AssignVariable("n", *newN)` too?

har777 (changes requested):

pkg/hintrunner/zero/zerohint_others.go
TAdev0: its not needed
newN is now a fp.element, not a *fp.element

cicr99 (approved):
lgtm

pkg/hintrunner/zero/zerohint_others.go
har777: ah lol didn't notice the assignment wasn't happening anymore

PR #446, opened 2024-06-07
Implement Blake2sCompute hint

### LINKED ISSUE #279: Blake2sCompute: Executes the BLAKE2s hashing function using the provided segments and output pointer.

har777:
Fixes #279


pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -215,3 +216,143 @@ func createBlake2sFinalizeHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+			if err != nil {
+				return err
+			}
+			var messageUint32 []uint32
+			for i := 0; i < 16; i++ {
+				value, err := hintrunnerUtils.ToSafeUint32(&message[i])
+				if err != nil {
+					return err
+				}
+				messageUint32 = append(messageUint32, value)
```
MaksymMalicki, line 289: since the size of the array is 16, why append instead of setting to an index just like in hUint32

har777: Ah because the method signature of the `Blake2sCompress` function expects that

MaksymMalicki (approved):

pkg/hintrunner/zero/zerohint_blake_test.go
```diff
@@ -521,5 +517,53 @@ func TestZeroHintBlake(t *testing.T) {
 				errCheck: errorTextContains(fmt.Sprintf("in range [0, 100), got %d", 200)),
 			},
 		},
+		"Blake2sCompute": {
+			{
+				operanders: []*hintOperander{
+					{Name: "output", Kind: apRelative, Value: addrWithSegment(1, 31)},
```
TAdev0, line 523: you can just do  `addr(..)` as it resolves to `vm.ExecutionSegment` which is `1` 
but its clearer like this also i agree !

TAdev0 (approved):
great !

PR #454, opened 2024-06-10
Implement EcMulInner

### LINKED ISSUE #453: EcMulInner
Reference: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/cairo_secp/ec.cairo#L292

har777:
Fix #453


pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -694,3 +695,34 @@ func createComputeSlopeV1Hinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+		Op: func(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
+			//> memory[ap] = (ids.scalar % PRIME) % 2
+
+			scalarFelt, err := hinter.ResolveAsFelt(vm, scalar)
+			if err != nil {
+				return err
+			}
+			scalarBytes := scalarFelt.Bytes()
+
+			resultUint256 := new(uint256.Int).SetBytes(scalarBytes[:])
```
MaksymMalicki, line 711: Look good, I am wondering if we should make some new util out of felt modulo.

MaksymMalicki (approved):

pkg/hintrunner/zero/zerohint_ec.go
har777: This is the only hint needing it. So probably ok to skip it for now 🤔

TAdev0 (approved):
lgtm @har777  !

Just to confirm, you dont need to apply `scalar % PRIME` because the scalar operander willl never be > PRIME right (because we resolve as felt)?

har777:
> lgtm @har777 !
> 
> Just to confirm, you dont need to apply `scalar % PRIME` because the scalar operander willl never be > PRIME right (because we resolve as felt)?

Yep thats what I was thinking!

TAdev0:
@har777  mmh ok!

`ids.scalar` is a BigInt3 :

```
struct BigInt3 {
    d0: felt,
    d1: felt,
    d2: felt,
}
```

are you sure resolving this as a felt will work when running a cairo file?

TAdev0:
I think current impl is wrong
A bigint3 takes 3 memory slots

What you do is retrieving only the first field of the struct

TAdev0:
Oh no my bad , it's a bigInt3 in ec mull , not in ec mull inner 

All good !

Issue #465, opened 2024-06-14
IsZeroNondet


Sh0g0-1758:
TODO : Implement the second hint used in the function IsZero. 
Reference : https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/common/cairo_secp/field.cairo#L101

PR #468, opened 2024-06-14
ZeroNonDec Hint Implementation

Sh0g0-1758:
Fixes: #465


pkg/hintrunner/zero/zerohint_signature.go
```diff
@@ -386,3 +387,43 @@ func createDivModNPackedDivmodV1Hinter(resolver hintReferenceResolver) (hinter.H
 
 	return newDivModNPackedDivmodV1Hint(a, b), nil
 }
+
+// IsZeroNondet hint computes whether a value is zero or not
```
TAdev0, line 391: can you look at how comments are formatted for others hints and do the same?

pkg/hintrunner/zero/zerohint_signature_test.go
```diff
@@ -248,5 +248,46 @@ func TestVerifyZeroHint(t *testing.T) {
...
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newIsZeroNondetHint()
+				},
+				check: apValueEquals(feltUint64(0)),
+			},
+			{
+				operanders: []*hintOperander{},
+				ctxInit: func(ctx *hinter.HintRunnerContext) {
+					err := ctx.ScopeManager.AssignVariable("x", bigIntString("-42", 10))
```
TAdev0, line 281: `x` will never be a negative value as it is set to : 

```
x = pack(ids.x, PRIME) % SECP_P
```

TAdev0 (changes requested):

TAdev0 (changes requested):
I think this should not go to signature, as it just check whether a field is equal to 0 or not, module secp256k1 prime

this should go to EC hints (hints related to secp256k1 so it makes sense) i think

pkg/hintrunner/zero/zerohint_signature.go
Sh0g0-1758: done.

pkg/hintrunner/zero/zerohint_signature_test.go
Sh0g0-1758: done.

Sh0g0-1758:
Right Relocated the hints.

pkg/hintrunner/zero/zerohint_ec_test.go
```diff
@@ -846,6 +846,47 @@ func TestZeroHintEc(t *testing.T) {
...
+					if err != nil {
+						t.Fatal(err)
+					}
+				},
+				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
+					return newIsZeroNondetHint()
+				},
+				check: apValueEquals(feltUint64(0)),
+			},
+		},
```
har777, line 889: isn't this test redundant?

har777 (approved):

TAdev0 (approved):

pkg/hintrunner/zero/zerohint_ec_test.go
TAdev0: agree just one test is enough

the function takes as input a field element so it doesnt make sense to test it with negative values. currently the 3 unit tests test the same thing so just one should be enough

Sh0g0-1758: Okay. Removing the test. But 2 tests are required, why do you say one @TAdev0  ?

pkg/hintrunner/zero/zerohint.go
```diff
@@ -122,8 +122,8 @@ func GetHintFromCode(program *zero.ZeroProgram, rawHint zero.Hint, hintPC uint64
 		return createVerifyZeroHinter(resolver)
 	case divModNPackedDivmodV1Code:
 		return createDivModNPackedDivmodV1Hinter(resolver)
-	case isZeroPackCode:
-		return createIsZeroPackHinter(resolver)
+	case isZeroNondetCode:
+		return createIsZeroNondetHinter()
```
TAdev0, line 126: there is an issue here, you dont call the right create function

TAdev0 (changes requested):

Sh0g0-1758:
Hmm... There are some serious branch issues going on here. It seems like two branches have merged for some reason. Wait Let me correct it.

pkg/hintrunner/zero/zerohint.go
Sh0g0-1758: Right, Fixed it.

Sh0g0-1758:
@TAdev0 , we good to merge now ?

TAdev0 (approved):

PR #474, opened 2024-06-20
Fixed a scope bug, updated comments and fixed hint code strings

Sh0g0-1758:
A couple changes in this one : 

1. `DivModNPackedDivmodV1` hint didn't add a , b and N variables in scope, while these are used in the next hint `div_mod_n` int `recover_public_key` cairo function
2. in the hint `newVerifyZeroHint` in the comments we have written `SECP256R1` but its `SECP`
3. in the hint `newGetPointFromXHint` the hint code was not properly formatted so rectified it, both in the comments and in the `hintcode.go` file
4. the string for `reduceV1Code` is also not properly formatted, it does not take into account the extra space after line end

### DIFF
--- a/pkg/hintrunner/zero/zerohint_signature.go
+++ b/pkg/hintrunner/zero/zerohint_signature.go
@@ -12,16 +12,16 @@ import (
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
-// VerifyZero hint verifies that a packed value is zero modulo the SECP256R1 prime
+// VerifyZero hint verifies that a packed value is zero modulo the SECP prime
 // and stores in memory the quotient of the modular divison of the packed value by
-// SECP256R1 prime
+// SECP prime
 //
 // `newVerifyZeroHint` takes 2 operanders as arguments
 //   - `value` is the value that will be verified
 //   - `q` is the variable that will store the quotient of the modular division
 //
 // `newVerifyZeroHint` writes the quotient of the modular division of the packed value
-// by SECP256R1 prime to the memory address corresponding to `q`
+// by SECP prime to the memory address corresponding to `q`
 func newVerifyZeroHint(val, q hinter.ResOperander) hinter.Hinter {
 	return &GenericZeroHinter{
 		Name: "VerifyZero",
@@ -30,8 +30,6 @@ func newVerifyZeroHint(val, q hinter.ResOperander) hinter.Hinter {
 			//> q, r = divmod(pack(ids.val, PRIME), SECP_P)
 			//> assert r == 0, f"verify_zero: Invalid input {ids.val.d0, ids.val.d1, ids.val.d2}."
 			//> ids.q = q % PRIME
-
-			//> from starkware.cairo.common.cairo_secp.secp_utils import SECP_P, pack
 
 			valAddr, err := val.GetAddress(vm)
 			if err != nil {
@@ -165,9 +163,9 @@ func newGetPointFromXHint(xCube, v hinter.ResOperander) hinter.Hinter {
 			//> y_square_int = (x_cube_int + ids.BETA) % SECP_P
 			//> y = pow(y_square_int, (SECP_P + 1) // 4, SECP_P)
 			//> if ids.v % 2 == y % 2:
-			//>	 value = y
+			//>		value = y
 			//> else:
-			//>	 value = (-y) % SECP_P
+			//>		value = (-y) % SECP_P
 
 			xCubeAddr, err := xCube.GetAddress(vm)
 			if err != nil {
@@ -297,7 +295,9 @@ func newDivModSafeDivHint() hinter.Hinter {
 				return err
 			}
 
-			return ctx.ScopeManager.AssignVariable("value", &value)
+			value_Big := new(big.Int).Set(&value)
+
+			return ctx.ScopeManager.AssignVariable("value", value_Big)
 		},
 	}
 }
@@ -319,6 +319,7 @@ func newDivModNPackedDivmodV1Hint(a, b hinter.ResOperander) hinter.Hinter {
 		Op: func(vm *VM.VirtualMachine, ctx *hinter.HintRunnerContext) error {
 			//> from starkware.cairo.common.cairo_secp.secp_utils import N, pack
 			//> from starkware.python.math_utils import div_mod, safe_div
+			//>
 			//> a = pack(ids.a, PRIME)
 			//> b = pack(ids.b, PRIME)
 			//> value = res = div_mod(a, b, N)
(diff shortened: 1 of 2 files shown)


pkg/hintrunner/zero/zerohint_signature.go
```diff
@@ -12,16 +12,16 @@ import (
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
-// VerifyZero hint verifies that a packed value is zero modulo the SECP256R1 prime
+// VerifyZero hint verifies that a packed value is zero modulo the SECP prime
```
TAdev0, line 15: can you replace it with `secp256k1` ? 
just chcked and this hint is related to `secp256k1` curve

pkg/hintrunner/zero/zerohint_signature.go
```diff
@@ -12,16 +12,16 @@ import (
...
 // and stores in memory the quotient of the modular divison of the packed value by
-// SECP256R1 prime
+// SECP prime
```
TAdev0, line 17: same

pkg/hintrunner/zero/zerohint_signature.go
```diff
@@ -12,16 +12,16 @@ import (
...
 //
 // `newVerifyZeroHint` takes 2 operanders as arguments
 //   - `value` is the value that will be verified
 //   - `q` is the variable that will store the quotient of the modular division
 //
 // `newVerifyZeroHint` writes the quotient of the modular division of the packed value
-// by SECP256R1 prime to the memory address corresponding to `q`
+// by SECP prime to the memory address corresponding to `q`
```
TAdev0, line 24: same

pkg/hintrunner/zero/zerohint_signature.go
```diff
@@ -31,8 +31,6 @@ func newVerifyZeroHint(val, q hinter.ResOperander) hinter.Hinter {
 			//> assert r == 0, f"verify_zero: Invalid input {ids.val.d0, ids.val.d1, ids.val.d2}."
 			//> ids.q = q % PRIME
 
-			//> from starkware.cairo.common.cairo_secp.secp_utils import SECP_P, pack
```
TAdev0, line 34: why do you remove this comment? its needed, this hint added the python hint line by line in the impl, you should leave it , we import `secPBig`

pkg/hintrunner/zero/zerohint_signature.go
```diff
@@ -297,7 +295,9 @@ func newDivModSafeDivHint() hinter.Hinter {
 				return err
 			}
 
-			return ctx.ScopeManager.AssignVariable("value", &value)
+			value_Big := new(big.Int).Set(&value)
```
TAdev0, line 298: value is already a bigInt, the previous code was correct, it stores in scope a pointer to a bigInt with `&value`

pkg/hintrunner/zero/zerohint_signature.go
```diff
@@ -366,9 +367,33 @@ func newDivModNPackedDivmodV1Hint(a, b hinter.ResOperander) hinter.Hinter {
...
 			}
 
-			valueBig := new(big.Int).Set(&resBig)
+			value_Big := new(big.Int).Set(&resBig)
+			res_Big := new(big.Int).Set(&resBig)
+			a_Big := new(big.Int).Set(&aPackedBig)
+			b_Big := new(big.Int).Set(&bPackedBig)
+			n_Big := new(big.Int).Set(&nBig)
+
+			if err := ctx.ScopeManager.AssignVariable("res", res_Big); err != nil {
```
TAdev0, line 376: its all good except you dont need to assign `res` , just `value` is needed.

TAdev0 (changes requested):

pkg/hintrunner/zero/zerohint_signature.go
Sh0g0-1758: done.

pkg/hintrunner/zero/zerohint_signature.go
Sh0g0-1758: done.

pkg/hintrunner/zero/zerohint_signature.go
Sh0g0-1758: done.

pkg/hintrunner/zero/zerohint_signature.go
Sh0g0-1758: Screenshot from 2024-06-23 01-46-39

I suppose this is the hint above. I dont see that extra import anywhere.

pkg/hintrunner/zero/zerohint_signature.go
Sh0g0-1758: Right . Agreed.

pkg/hintrunner/zero/zerohint_signature.go
Sh0g0-1758: But in the next hint, res is being used : 

`value = k = safe_div(res * b - a, N)`

pkg/hintrunner/zero/zerohint_signature.go
TAdev0: look line 49 : 

//> q, r = divmod(pack(ids.val, PRIME), SECP_P)

the person who implemented basically wrote 2 times each line of python hint : 
- one time at the beginning of the new function, the whole hint
- another time in the impl line by line 

here, you should leave : 

//> from starkware.cairo.common.cairo_secp.secp_utils import SECP_P, pack


because it describes : 

	secPBig, ok := secp_utils.GetSecPBig()
			if !ok {
				return fmt.Errorf("GetSecPBig failed")
			}

pkg/hintrunner/zero/zerohint_signature.go
TAdev0: oh yeah you are right . my bad!

pkg/hintrunner/zero/zerohint_signature.go
Sh0g0-1758: Well then would it not better to move it just above the following line : 

```
secPBig, ok := secp_utils.GetSecPBig()
		if !ok {
			return fmt.Errorf("GetSecPBig failed")
		}
```

Where it makes sense. Because the current impl seems to suggest that the hint has that extra line of code.

TAdev0: you're right lets do that i guess!

Sh0g0-1758: done.

TAdev0 (approved):

PR #475, opened 2024-06-20
Improve `ZeroDictionary` struct

TAdev0:
This PR modifies the fields of  `ZeroDictionary` struct to have only pointers . Because `GetDictionary` actually creates a copy of the dict : 

```
// Given a memory address, it looks for the right dictionary using the segment index. If no
// segment is associated with the given segment index, it errors
func (dm *ZeroDictionaryManager) GetDictionary(dictAddr mem.MemoryAddress) (ZeroDictionary, error) {
	dict, ok := dm.Dictionaries[dictAddr.SegmentIndex]
	if ok {
		return dict, nil
	}
	return ZeroDictionary{}, fmt.Errorf("no dictionary at address: %s", dictAddr)
}
```

we will gain in performance in storing only an address for the whole mapping in the `ZeroDictionary`

### DIFF
--- a/pkg/hintrunner/hinter/zero_dict.go
+++ b/pkg/hintrunner/hinter/zero_dict.go
@@ -11,27 +11,27 @@ import (
 // Used to keep track of all dictionaries data
 type ZeroDictionary struct {
 	// The Data contained in a dictionary
-	Data map[fp.Element]mem.MemoryValue
+	Data *map[fp.Element]mem.MemoryValue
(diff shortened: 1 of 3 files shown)


TAdev0:
@har777  we discussed the fact that `FreeOffset` field is of type `*uint64` , its actually required with our current impl of `GetDictionary` , as we return a copy

Sh0g0-1758 (approved):

PR #476, opened 2024-06-22
feat: allow exporting memory file and trace file in non proof mode

### LINKED ISSUE #472: Export trace and memory in execution mode
Allow exporting the trace and memory when running in execution mode (non-proof mode) on the CLI. It is currently only doable with proof mode

TAdev0:
Resolves #472


pkg/vm/vm.go
```diff
@@ -537,7 +523,7 @@ func (vm *VirtualMachine) updateFp(instruction *a.Instruction, dstAddr *mem.Memo
 	}
 }
 
-func (vm *VirtualMachine) relocateTrace() []Trace {
+func (vm *VirtualMachine) RelocateTrace() []Trace {
```
rodrigo-pino, line 526: Since we are making it public I'd suggest adding a description to it

pkg/vm/vm.go
```diff
@@ -136,9 +133,7 @@ func (vm *VirtualMachine) RunStep(hintRunner HintRunner) error {
 	}
 
 	// store the trace before state change
-	if vm.config.ProofMode {
-		vm.Trace = append(vm.Trace, vm.Context)
-	}
+	vm.Trace = append(vm.Trace, vm.Context)
```
rodrigo-pino, line 136: I wouldn't add the trace every time. Collecting the trace vs not is a performance hit. I would suggest to have a flag in the VM.config (perhaps `get_trace`, `collect_trace` or any name you think is best) and keep the if

pkg/vm/vm.go
```diff
@@ -91,11 +91,8 @@ type VirtualMachine struct {
 func NewVirtualMachine(
 	initialContext Context, memory *mem.Memory, config VirtualMachineConfig,
 ) (*VirtualMachine, error) {
-	// Initialize the trace if necesary
-	var trace []Context
-	if config.ProofMode {
-		trace = make([]Context, 0)
-	}
+
+	trace := make([]Context, 0)
```
rodrigo-pino, line 95: The same here

cmd/cli/main.go
```diff
@@ -115,24 +115,26 @@ func main() {
 						}
 					}
 
+					runner.EndRun()
```
rodrigo-pino, line 118: I was looking into the code of `EndRun`. Outside of the scope of the PR but could you remove the `if` conditional it has inside. It is usually not a good practice to have a conditional inside a function, it's better to have it outside. For two reasons:
* Is easy for people to understand that `EndRun` only makes sense in Proof Mode
* You remove the performance cost of making a call that does nothing, when there is not proof mode. (Although, in this case, the function is probably getting inlined).

cc: @MaksymMalicki

cmd/cli/main.go
```diff
@@ -115,24 +115,26 @@ func main() {
...
-						if traceLocation != "" {
-							if err := os.WriteFile(traceLocation, trace, 0644); err != nil {
-								return fmt.Errorf("cannot write relocated trace: %w", err)
-							}
+					}
+
+					trace, memory, err := runner.BuildProof()
+					if err != nil {
+						return fmt.Errorf("cannot build proof: %w", err)
+					}
```
rodrigo-pino, line 129: In line with the previous comments, we shouldn't do this unless specified by the user, so it's good to have a flag here. Either proof mode is on or they asked for a trace.

pkg/vm/vm.go
TAdev0: done!

cmd/cli/main.go
TAdev0: done

integration_tests/BenchMarks.txt
rodrigo-pino, line 1: I know this is out of scope for this PR, but you should really compare benchmarks against the Rust VM

rodrigo-pino (approved):

integration_tests/BenchMarks.txt
TAdev0: Well I proposed to do it and the conclusion was that it's not necessary

TAdev0: It was concluded that we can just benchmark locally with the rust one

TAdev0:
I'll merge this and open a new issue for the 1 memory cell / 1 step divergence

PR #477, opened 2024-06-22
FEAT : Benchmark for integration tests

Sh0g0-1758:
This PR aims to add a Benchmark for Integration tests.


integration_tests/cairozero_test.go
```diff
@@ -76,18 +80,27 @@ func TestCairoZeroFiles(t *testing.T) {
...
 
+		start := time.Now()
+
 		pyTraceFile, pyMemoryFile, err := runPythonVm(dirEntry.Name(), compiledOutput)
 		if err != nil {
 			t.Error(err)
 			continue
 		}
 
+		elapsed_py := time.Since(start)
```
TAdev0, line 91: can you benchmark before and after `	res, err := cmd.CombinedOutput()` ?  which is the real run

as runVM and runPythonVM are not strictly the same

TAdev0 (changes requested):

TAdev0:
looks good overall! 
that way we ll have benchmarks for all cairo zero hints, and any cairo program we want

TAdev0:
@Sh0g0-1758  i just worked locally on a branch to run the rust VM as well in integration tests, as it is the one currently in production it might make sense to add it. Just wondering how to deal with workflow github action and running the rust vm

if we add the rust vm then we'll need to modify this as well to benchmark 3 runs

integration_tests/cairozero_test.go
Sh0g0-1758: Hmmm... Can you point out the differences. From what I can see, Rest are constant time operations only and should not add any significant overhead. Even if not so, I think both the functions are sufficiently similar to have the same overheads.

TAdev0: why not benchmarking only the hint? you benchmark additionnal code without any reason right now

Sh0g0-1758: Hmm... Okay will make the changes. My point was that the current impl would not make much difference since the operations other than the main operation take place in constant negligible time.

Sh0g0-1758: done.

integration_tests/cairozero_test.go
```diff
@@ -93,18 +97,20 @@ func TestCairoZeroFiles(t *testing.T) {
 			continue
 		}
 
-		pyTraceFile, pyMemoryFile, err := runPythonVm(dirEntry.Name(), compiledOutput)
+		elapsed_py, pyTraceFile, pyMemoryFile, err := runPythonVm(dirEntry.Name(), compiledOutput)
```
har777, line 100: nit: camelCase

integration_tests/cairozero_test.go
```diff
@@ -127,10 +133,73 @@ func TestCairoZeroFiles(t *testing.T) {
...
+
+// format a row with borders and spaces based on the column widths
+func formatRow(row []string, widths []int) string {
+	var sb strings.Builder
+	sb.WriteString("|")
+	for i, col := range row {
+		sb.WriteString(" " + padRight(col, widths[i]) + " |")
+	}
+	return sb.String()
+}
```
har777, line 196: Check out: https://pkg.go.dev/text/tabwriter
Might be a lot easier and its in the stdlib.

integration_tests/cairozero_test.go
Sh0g0-1758: done.

integration_tests/cairozero_test.go
Sh0g0-1758: I don't think we can achieve different column width for the same row with tabwriter. Can we go ahead with the current impl ?

Sh0g0-1758: Updated with tabwriter. Thanks for the suggestion.

TAdev0 (approved):

PR #484, opened 2024-06-25
feat: convert `felt` to `uint64` in hints where possible

### LINKED ISSUE #383: Felt values in hints representing array length
There are cases such as this [hint](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/memcpy.cairo#L24):
```
%{
        n -= 1
        ids.continue_copying = 1 if n > 0 else 0
%}
```
where the value of n is supposed to be a felt, as you can see in its definition [here](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/common/memcpy.cairo#L12):
```
%{ vm_enter_scope({'n': ids.len}) %}
```
where `len` is of type felt.

However, in practice, the actual value of `n` cannot be a value of 252 bits since that wouldn't be feasible to compute, as this is representing the length of an array. `n` should fit in int64. If we do the conversion from the moment we define the variable then all operations regarding `n` would be a lot easier and more efficient, as the way we can handle felts is using Element from [this library](https://github.com/Consensys/gnark-crypto/tree/master/ecc/stark-curve/fp).

**Task**: Look for all the places in the code of hints where this can be optimized and do so. Notice you'll also need to modify all the code related to operations with these variables

TAdev0:
Resolves #383


TAdev0:
waiting for usort integration tests merging to finish

TAdev0:
ready for review @cicr99  @har777

pkg/hintrunner/zero/zerohint_others.go
```diff
@@ -37,12 +37,11 @@ func newMemContinueHint(continueTarget hinter.ResOperander, memset bool) hinter.
 				return err
 			}
 
-			newN, ok := n.(fp.Element)
+			newN, ok := n.(uint64)
 			if !ok {
 				return fmt.Errorf("casting n into a felt failed")
```
cicr99, line 42: Should we update the error message as well?

cicr99 (approved):
I thought there were other hints that also needed this update such as SearchSortedLower, or FindElement hints, but I see all of them work already with uint64 when it comes to the size of elements or arrays. That's good. It looks good to me

pkg/hintrunner/zero/zerohint_others.go
TAdev0: sure! will do it now

PR #487, opened 2024-06-26
Nit : Clean Integration test 

Sh0g0-1758:
NIT

### DIFF
--- a/integration_tests/cairozero_test.go
+++ b/integration_tests/cairozero_test.go
@@ -60,9 +60,9 @@ func TestCairoZeroFiles(t *testing.T) {
 
 	testFiles := append(testFiles1, testFiles2...)
 
-	testFiles1Map := make(map[string]bool, len(testFiles1))
+	testFiles1Map := make(map[string]struct{}, len(testFiles1))
 	for _, entry := range testFiles1 {
-		testFiles1Map[entry.Name()] = true
+		testFiles1Map[entry.Name()] = struct{}{}
 	}
 
 	// filter is for debugging purposes
--- a/integration_tests/cairozero_test.go
+++ b/integration_tests/cairozero_test.go
@@ -60,9 +60,9 @@ func TestCairoZeroFiles(t *testing.T) {
 
 	testFiles := append(testFiles1, testFiles2...)
 
-	testFiles1Map := make(map[string]struct{}, len(testFiles1))
+	testFiles1Map := make(map[string]bool, len(testFiles1))
 	for _, entry := range testFiles1 {
-		testFiles1Map[entry.Name()] = struct{}{}
+		testFiles1Map[entry.Name()] = true
 	}
 
 	// filter is for debugging purposes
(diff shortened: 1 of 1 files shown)


integration_tests/cairozero_test.go
```diff
@@ -74,11 +74,9 @@ func TestCairoZeroFiles(t *testing.T) {
 			continue
 		}
 
-		var path string
+		path := filepath.Join(root2, dirEntry.Name())
 		if _, found := testFiles1Map[dirEntry.Name()]; found {
 			path = filepath.Join(root1, dirEntry.Name())
-		} else {
```
TAdev0, line 80: why do you remove this?
its part of integration tests, it files without cairo zero hints

integration_tests/cairozero_test.go
```diff
@@ -60,9 +60,9 @@ func TestCairoZeroFiles(t *testing.T) {
 
 	testFiles := append(testFiles1, testFiles2...)
 
-	testFiles1Map := make(map[string]struct{}, len(testFiles1))
```
TAdev0, line 63: does it change anything to do that?

Sh0g0-1758: No, its just cleaner that way. We are using map to detect whether testfile is a part of root1. And mapping to a struct does not really give away that message. But I think mapping to a bool surely does.

integration_tests/cairozero_test.go
Sh0g0-1758: Again, its cleaner like this. 

This is the initializer : `path := filepath.Join(root2, dirEntry.Name())`
And this updates it if it is in root1 : 		
```
if _, found := testFiles1Map[dirEntry.Name()]; found {
    path = filepath.Join(root1, dirEntry.Name())
}
```

integration_tests/cairozero_test.go
TAdev0: we dont care of the type of the value , we just need the key

integration_tests/cairozero_test.go
TAdev0: i dont get it, why would you do with no conditiion

```
path := filepath.Join(root2, dirEntry.Name())
```
?

integration_tests/cairozero_test.go
Sh0g0-1758: yeah um... as I said its just to make it look cleaner.

integration_tests/cairozero_test.go
Sh0g0-1758: Because it gets updated if its root1 and if its not, it remains root2.

integration_tests/cairozero_test.go
Sh0g0-1758: Oh no nothing. Its just have to be assigned something to put it in the map. It can be false also, it does not matter. I chose true just to make intuitive that this map "contains" the root1 files.

TAdev0: ok so i just checked a bit:
in go, using struct{} when you just need to check for keys is better, as it does use any memory for the value, in contrary to `bool`

we should keep `struct{}{}`

TAdev0 (changes requested):

integration_tests/cairozero_test.go
Sh0g0-1758: Oh, interesting. I was not aware of that.

TAdev0 (approved):

PR #512, opened 2024-07-03
typo comment

TAdev0:


Sh0g0-1758 (approved):
The updated comment is correct. In the file, secp_utils, we have `SECP256K1 = Curve(prime=SECP_P, alpha=ALPHA, beta=BETA, order=N)` and in the hint we import SECP_P from that file.

PR #513, opened 2024-07-03
Implement RandomEcPoint hint

har777:


pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -945,3 +940,137 @@ func createRecoverYHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
 
 	return newRecoverYHint(x, p), nil
 }
+
+// RandomEcPoint hint Returns a random non-zero point on the elliptic curve
```
TAdev0, line 944: > Returns

returns

TAdev0 (approved):
LGTM !

pkg/hintrunner/zero/zerohint_ec.go
TAdev0, line 944:
```suggestion
// RandomEcPoint hint returns a random non-zero point on the elliptic curve
```

PR #516, opened 2024-07-08
Fix some integration test bugs

har777:
* fix some incorrect error handling
* fix how `createNondetElementsOverTwoHinter` and `createNondetElementsOverTenHinter` works


TAdev0:
Yep.. noticed also it's finally not "n", also did this locally. 
I can't find out which Cairo file resulted in "n" instead of the 2 variables .. anyway

TAdev0 (approved):
lgtm! made almost the same changes locally on a branch, but yours is better refactoring with `x` being passed in the create function

PR #525, opened 2024-07-13
Code cleanups and comments for the finalisation code

MaksymMalicki:


pkg/vm/vm.go
```diff
@@ -86,8 +86,9 @@ type VirtualMachine struct {
 	config  VirtualMachineConfig
 	// instructions cache
 	instructions map[uint64]*a.Instruction
-	RcLimitsMin  uint64
-	RcLimitsMax  uint64
+	// RcLimitsMin and RcLimitsMax define the range of values of instructions offsets, used for checking the the number of potential range checks holes
```
TAdev0, line 89: > the the

the

pkg/runners/zero/zero.go
```diff
@@ -272,13 +272,14 @@ func (runner *ZeroRunner) RunFor(steps uint64) error {
 // until the checkUsedCells doesn't return any error.
 // Since this vm always finishes the run of the program at the number of steps that is a power of two in the proof mode,
 // there is no need to run additional steps before the loop.
-func (runner *ZeroRunner) EndRun() {
+func (runner *ZeroRunner) EndRun() error {
 	for runner.checkUsedCells() != nil {
 		pow2Steps := utils.NextPowerOfTwo(runner.vm.Step + 1)
 		if err := runner.RunFor(pow2Steps); err != nil {
-			panic(err)
+			return err
```
TAdev0, line 279: is it required not to panic here?

pkg/vm/builtins/range_check.go
```diff
@@ -46,17 +48,18 @@ func (r *RangeCheck) GetAllocatedSize(segmentUsedSize uint64, vmCurrentStep uint
 	return getBuiltinAllocatedSize(segmentUsedSize, vmCurrentStep, r.ratio, inputCellsPerRangeCheck, instancesPerComponentRangeCheck, cellsPerRangeCheck)
 }
 
-func (r *RangeCheck) GetRangeCheckUsage(rangeCheckSegment *memory.Segment) (uint64, uint64) {
-	minVal, maxVal := ^uint64(0), uint64(0)
+// GetRangeCheckUsage returns the min and max values used in the range check segment. Since each range check instance consists of 16-bit parts, the min and max values are calculated by iterating over the segment data and extracting the 16-bit parts from each field element.
```
TAdev0, line 51: // getPermRangeCheckLimits returns the minimum and maximum values used by the range check units in the program.

// GetRangeCheckUsage returns the min and max values used in the range check segment.

doesnt these sound the same?

MaksymMalicki: Nope, bc GetRangeCheckUsage checks the segment, and the getPermRangeCheckLimits checks "all" the range checks done during the execution, which includes range checks done on the instructions

pkg/runners/zero/zero.go
MaksymMalicki: cleaner imo, all of the steps after the execution are handled with an error, so I thought why panic here

pkg/vm/vm.go
MaksymMalicki: got it

pkg/runners/zero/zero.go
TAdev0: Makes sense

pkg/vm/builtins/range_check.go
TAdev0: And range check done on instructions are not in the range check segment ?

cicr99 (commented):
lgtm!
the conversations need to be solved though, I'll let @TAdev0 be the one approving this

pkg/vm/builtins/range_check.go
MaksymMalicki: from what I understand they are not, because for instruction we are checking only if the offsets are in range 2**16, and add that to the trace, so then the virtual column with permutation range checks is constructed by the prover.

MaksymMalicki: There is a difference between regular range check and permutation range check, first one checks if the number is in desired range (for instance, normally we use 8 * 2**16 checks), where permutation range check is a single range check unit

TAdev0 (approved):

Issue #522, opened 2024-07-11
bug: `KeccakWriteArgs` hint in _keccak.starknet_with_keccak.cairo_ resolves `keccak_ptr` with the wrong reference


TAdev0:
keccak.starknet_with_keccak.cairo integration test includes a call to `_block_permutation`  which takes `keccak_ptr` as an implicit argument.
`keccak_ptr`  unexpectedly resolves to the right value + 25 . Integration tests can pass if we slightly modify our hint to read from ` offset - 25` and write to `offset - 25`.

PR #528, opened 2024-07-16
`cairoKeccakFinalizeBlockSize1000Code` missing hint

TAdev0:
variant of `cairoKeccakFinalize`, with block sized max being 1000 instead of 10 (but anyway, blocksize is hardcoded to 3)

whitelist is here : 

https://github.com/starkware-libs/cairo-lang/blob/0e4dab8a6065d80d1c726394f5d9d23cb451706a/src/starkware/starknet/security/whitelists/cairo_keccak.json#L31

cairo file is in in_progress folder because of #522


har777 (approved):
lgtm!

Issue #546, opened 2024-07-17
`FastEcAddAssignNewXV2` hint


TAdev0:
https://github.com/starkware-libs/cairo-lang/blob/0e4dab8a6065d80d1c726394f5d9d23cb451706a/src/starkware/starknet/security/whitelists/vrf.json#L252

diegoTech14:
Hello @TAdev0

My name is Diego Duarte Fernández, a software developer from Costa Rica, I'm new on this repo but I would like to learn more about it and would be happy to help with this issue.

Thanks for your consideration.

Kind regards,
Diego ⚡

Issue #542, opened 2024-07-17
`FastEcAddAssignNewXV3` hint


TAdev0:
https://github.com/lambdaclass/cairo-vm/blob/db7fff18c9c1312024ebf4119fcdfeba23657b12/vm/src/hint_processor/builtin_hint_processor/hint_code.rs#L916

https://github.com/lambdaclass/cairo-vm/blob/db7fff18c9c1312024ebf4119fcdfeba23657b12/cairo_programs/fast_ec_add_v3.cairo#L42

https://github.com/starkware-libs/cairo-lang/blob/0e4dab8a6065d80d1c726394f5d9d23cb451706a/src/starkware/starknet/security/whitelists/cairo_secp.json#L177

PR #547, opened 2024-07-17
`fastEcAddAssignNewXV2` + `fastEcAddAssignNewXV3` hint

TAdev0:
Resolves https://github.com/NethermindEth/cairo-vm-go/issues/542
Resolves https://github.com/NethermindEth/cairo-vm-go/issues/546

While implementing these hints, found a scope bug  leading to this error when executing the next hint:

```
     runtime error: pc 0:185 step 122: execute hint FastEcAddAssignNewY: variable SECP_P not found in current scope
```

corrected this bug also in the PR .
All integration work welll


TAdev0:
Resolves #542 
Resolves #546

TAdev0:
Note that the V2 is whitelisted but i cant find any cairo file that use this hint and works at the same time (import error related to outdated V2 folders...)

cicr99 (approved):
Looks a lot better this way

PR #570, opened 2024-07-18
Fix keccak block permutation integration tests bug

MaksymMalicki:


TAdev0 (approved):
lgtm

pkg/hintrunner/zero/zerohint.go
```diff
@@ -290,11 +290,13 @@ func getParameters(zeroProgram *zero.ZeroProgram, hint zero.Hint, hintPC uint64)
 			if references[i].Pc <= hintPC {
```
TAdev0, line 290: suggestion : 


```
[11](https://demerzelsolutions.slack.com/archives/D070DT5NKJ5/p1721347868981819)
		for i := len(references) - 1; i >= 0; i-- {
			if references[i].Pc <= hintPC {
				if i < len(references)-1 {
					if references[i].Pc != references[i+1].Pc && references[i+1].Pc <= hintPC {
						break
					}
				}
				reference = references[i]
				ok = true

			}
		}
```

MaksymMalicki: Simplified your proposal to this:
```
for i := len(references) - 1; i >= 0; i-- {
	if references[i].Pc <= hintPC {
		if ok && reference.Pc != references[i].Pc {
			break
		}
		reference = references[i]
		ok = true

	}
}
```

Issue #529, opened 2024-07-17
Modify `currentAccessIndex` and `key` scope variables in dict zero hints to pass a pointer instead of a felt


TAdev0:
Modify implementation (assignment in scope and retrievement) and tests accordingly

EmmanuelAR:
Hello @TAdev0

My name is Emmanuel, a software developer from Costa Rica. I am already working on this repository and would be happy to assist with this task. I will review the current implementation and will make the changes that you post in the description 
Please let me know if this is acceptable.

Best regards,
Emmanuel 😁

TAdev0:
hey @EmmanuelAR  sure, please go ahead :)

EmmanuelAR:
thnks @TAdev0, i will let u know if i have questions!

EmmanuelAR:
Hello @TAdev0 i create a the pr, please feel free to review it and let know any suggestion.

Best regards,
Emmanuel 😁

PR #571, opened 2024-07-18
Modify currentAccessIndex and key scope variables in dict zero hints to pass a pointer instead of a felt.

EmmanuelAR:
**Changes:** 
- Create a new pointer 
- Assigned the felt ( currentAccessIndex, key ) to the pointer
- Change the reference in each method that use the felt.

**Unit test running successfully**

Captura de pantalla 2024-07-18 a la(s) 3 06 05 p  m

### DIFF
--- a/pkg/hintrunner/zero/zerohint_dictionaries.go
+++ b/pkg/hintrunner/zero/zerohint_dictionaries.go
@@ -179,7 +179,11 @@ func createDictReadHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
 	if err != nil {
 		return nil, err
 	}
-	key, err := resolver.GetResOperander("key")
+
+	var keyName string = "key"
+	var keyNamePtr *string = &keyName
+
+	key, err := resolver.GetResOperander(*keyNamePtr)
 	if err != nil {
 		return nil, err
 	}
@@ -328,7 +332,11 @@ func createDictWriteHinter(resolver hintReferenceResolver) (hinter.Hinter, error
 	if err != nil {
 		return nil, err
 	}
-	key, err := resolver.GetResOperander("key")
+
+	var keyName string = "key"
+	var keyNamePtr *string = &keyName
+
+	key, err := resolver.GetResOperander(*keyNamePtr)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


EmmanuelAR:
Ticket ref: #529

TAdev0:
ok so what you need to do is this: 

```
	err = ctx.ScopeManager.AssignVariable("current_access_index", currentAccessIndex)
			if err != nil {
				return err
			}
```

here, we assign to the string `current_access_index` the variable `currentAccessIndex`, which is of type `fp.Element`

here, this should be modified for : 


```
	err = ctx.ScopeManager.AssignVariable("current_access_index", &currentAccessIndex)
			if err != nil {
				return err
			}
```

so that we store in the scope at key `current_access_index` a pointer to `currentAccessIndex` fp.Element value

is it more clear?

pkg/hintrunner/zero/zerohint_dictionaries.go
```diff
@@ -179,7 +179,11 @@ func createDictReadHinter(resolver hintReferenceResolver) (hinter.Hinter, error)
 	if err != nil {
 		return nil, err
 	}
-	key, err := resolver.GetResOperander("key")
+
+	var keyName string = "key"
+	var keyNamePtr *string = &keyName
+
+	key, err := resolver.GetResOperander(*keyNamePtr)
```
TAdev0, line 186: this doesnt change anything in what the code does.
`*keyNamePtr` resolves to `keyName` which is "key"

EmmanuelAR:
Hello @TAdev0.

Thanks for the feedback, i make the changes, please review it and let me know :)

TAdev0:
Oh so now you assigned in scope a pointer, you also need to modify the code where you retrieve theses variables . This is why integrations tests are not passing right now

EmmanuelAR:
Okk @TAdev0, I think k fix the unit testing but let me review it again! 🙌🏻

TAdev0:
Not unit test , integrations tests
See that CI is currently not passing

EmmanuelAR:
Ok yes, missed that part, let me fix it!

TAdev0:
Resolves #529

EmmanuelAR:
Unit and integration run
Captura de pantalla 2024-07-23 a la(s) 12 07 39 p  m
Captura de pantalla 2024-07-23 a la(s) 12 07 46 p  m
ning successfully

TAdev0 (approved):
lgtm!

PR #575, opened 2024-07-19
Fix dict implementation

### LINKED ISSUE #574: Deref error running cairo file
Cairo file: https://github.com/lambdaclass/cairo-vm/blob/main/cairo_programs/dict_store_cast_ptr.cairo

har777:
Fix #574 

Fixes 2 bugs:
1. The assumption that the dict key is always a felt. It can be an address too.
2. The assumption in a few dict hints that default_value/values/keys will be always resolved as felt. It can be an address too.


TAdev0 (approved):
lgtm!

I approve but there is just a conflict to resolve

PR #585, opened 2024-07-19
Separate bench cmd

### LINKED ISSUE #526: feat: create a separate command to run benchmarks
Currently, benchmarks are run each time we do `make integration` . This lead to the txt file being overwritten again and again.

fmmesen:
Resolves #526 

Changes made:
Removed the code that created the Benchmarks.txt file when the make integration command is run.
Added a new command `make integrationbench` that creates the Benchmarks.txt file with performance data.

Detailed description:
Command `make integration`: Now only runs the integration tests without generating the Benchmarks.txt file.
Command `make integrationbench`: This new command runs only the benchmark tests and generates the Benchmarks.txt file, providing detailed performance information.

### DIFF
--- a/integration_tests/cairo_zero_hint_tests/ec_double_v4.small.cairo
+++ b/integration_tests/cairo_zero_hint_tests/ec_double_v4.small.cairo
@@ -1,0 +1,84 @@
+%builtins range_check
+
+from starkware.cairo.common.cairo_secp.bigint import BigInt3, UnreducedBigInt3, nondet_bigint3
+from starkware.cairo.common.cairo_secp.ec import compute_doubling_slope, EcPoint
+from starkware.cairo.common.cairo_secp.field import (
+    is_zero,
+    unreduced_mul,
+    unreduced_sqr,
+    verify_zero,
+)
+
+// Computes the addition of a given point to itself.
+//
+// Arguments:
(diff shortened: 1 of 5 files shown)


TAdev0:
hey @fmmesen , thanks for your PR. Could you also modify the readme file to explain how to run benchmarks to compare performances with the python VM  on cairo zero files ?

 there any way to avoid the huge code duplicate that your proposed solution implies?

TAdev0:
hey @fmmesen 

@har777  proposed a more efficient solution to avoid code duplication.
Lets create a `make zerobench` command (because we already have a `make bench` command for something else)

Then you just keep only one benchmarks file, and do something like : 

```
import (
	"flag"
	"testing"
)

var zerobench = flag.Bool("zerobench", false, "run Integration tests and generates benchmarks file")

func TestCairoZeroFilesWithBenchmarks(t *testing.T) {
	// Your test code here
	// ...

	if zerobench {
		wrap the WriteToBenchmark line in it
	}
}
```

fmmesen:
Hello @TAdev0 !
Ok, I'll check it tonight

fmmesen:
Sorry @TAdev0 seems that I made a mistake working with github, but there's a new PR #597

PR #586, opened 2024-07-20
Implement `uint256_sub` hint

MaksymMalicki:


pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -505,3 +505,91 @@ func createUint256MulDivModHinter(resolver hintReferenceResolver) (hinter.Hinter
 
 	return newUint256MulDivModHint(a, b, div, quotientLow, quotientHigh, remainder), nil
 }
+
```
TAdev0, line 508: a small comment like we did for other hints?

TAdev0 (commented):
left just a comment , ohterwise lgtm ! 
i think you can also runk make integration to add it in the benchmark of main (or we do it later nevermind actually)

pkg/hintrunner/zero/zerohint_uint256.go
MaksymMalicki: added annotation

TAdev0 (approved):

PR #587, opened 2024-07-20
Implement `Uint256UnsignedDivRemExpanded` + `Split128` + `Uint128Sqrt` hints

MaksymMalicki:


integration_tests/cairo_zero_hint_tests/uint256_improvements.small.cairo
```diff
@@ -78,4 +295,11 @@ func main{range_check_ptr}() {
     assert sign = 0;
 
     return ();
-}
\ No newline at end of file
+}
+
```
TAdev0, line 299: why is `uint128_mul` function never called?

TAdev0 (approved):
code looks good overall, i give an approve, but maybe we just remove the unused function that doesnt contain hint in the cairo.

Also just some conflicts to resolve

integration_tests/cairo_zero_hint_tests/uint256_improvements.small.cairo
MaksymMalicki: copied entire integration test

TAdev0:
@MaksymMalicki  seems we are missing `uint256Sqrt`

https://github.com/lambdaclass/cairo-vm/blob/24c2349cc19832fd8c1552304fe0439765ed82c6/vm/src/hint_processor/builtin_hint_processor/hint_code.rs#L337

can you add it in the PR? its basically almost the same as `Sqrt` hint, but with a low and a high with high == 0

MaksymMalicki:
> @MaksymMalicki seems we are missing `uint256Sqrt`
> 
> https://github.com/lambdaclass/cairo-vm/blob/24c2349cc19832fd8c1552304fe0439765ed82c6/vm/src/hint_processor/builtin_hint_processor/hint_code.rs#L337
> 
> can you add it in the PR? its basically almost the same as `Sqrt` hint, but with a low and a high with high == 0

done

PR #590, opened 2024-07-22
fix: set right flag value in ShouldSkipSquashLoop hint

### LINKED ISSUE #589: bug: ShouldSkipSquashLoop hint not properly implemented
The [current implementation](https://github.com/NethermindEth/cairo-vm-go/blob/d739744295b0fe9e2cdf9f16ee3185053a30c520/pkg/hintrunner/core/hint.go#L1024-L1029) of `ShouldSkipSquashLoop` skips the squash loop if there is more than one indices, when it should be the other way round:

```go
var shouldSkipLoop f.Element
if lastIndices, err := ctx.SquashedDictionaryManager.LastIndices(); err == nil && len(lastIndices) > 1 {
	shouldSkipLoop.SetOne()
} else if err != nil {
	return fmt.Errorf("get last indices: %w", err)
}
```

For reference, the implementation in rust can be found [here](https://github.com/lambdaclass/cairo-vm/blob/886beca209ec42723e2562ea7ce8ceb0885dd41a/vm/src/hint_processor/cairo_1_hint_processor/hint_processor.rs#L853-L866)

zmalatrax:
Closes #589

### DIFF
--- a/pkg/hintrunner/core/hint.go
+++ b/pkg/hintrunner/core/hint.go
@@ -1021,14 +1021,15 @@ func (hint *ShouldSkipSquashLoop) Execute(vm *VM.VirtualMachine, ctx *hinter.Hin
 		return fmt.Errorf("get should skip loop address: %w", err)
 	}
 
-	var shouldSkipLoop f.Element
+	shouldSkipLoop := &f.Element{}
+	shouldSkipLoop.SetOne()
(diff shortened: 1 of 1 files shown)


TAdev0:
lgtm!

TAdev0 (approved):

PR #595, opened 2024-07-23
chore: update readme for python version

TAdev0:


README.md
```diff
@@ -28,8 +28,11 @@ After completing these steps, you can find the compiled VM in `bin/cairo-vm`.
 
 ### Run The VM
 
-To run the VM you need first to have a compiled Cairo file using the Cairo compiler at [cairo-lang](https://github.com/starkware-libs/cairo-lang).
-Install it with the following command:
+To run the VM you need to have a compiled Cairo file using the Cairo compiler at [cairo-lang](https://github.com/starkware-libs/cairo-lang).
```
rodrigo-pino, line 31: I know this is old, but I would say: _"...file using the Cairo **Zero** compiler..."_

README.md
```diff
@@ -28,8 +28,11 @@ After completing these steps, you can find the compiled VM in `bin/cairo-vm`.
...
+
+First of all, make sure you have Python 3 installed, with version 3.9.11. Indeed, many Python versions are not compatible with the latest version of the cairo-lang repository. You can find the link to download Python 3.9.11 [here](https://www.python.org/downloads/release/python-3911/).
+
+Then, install cairo-lang package latest version with the following command:
```
rodrigo-pino, line 35: I think this might be a bit too long.  I would suggest something shorter along the lines of:

_"First, make sure you have [Python 3.9.11](https://www.python.org/downloads/release/python-3911/) installed in your machine. Since this version is quite old and can cause problem with your system's Python we suggest the use of [pyenv](https://github.com/pyenv/pyenv) to manage different Python versions._

_Then, install the cairo-lang package with the following command:..."_

Notice that I've added some extra info on Python version managing, since of contributors are new to programming in general and might not know about this tool, but I leave it to your discretion to include it or not.

TAdev0: Way better, thanks !

README.md
TAdev0: I like it it's more precise
Thanks !

rodrigo-pino (approved):

Issue #538, opened 2024-07-17
`EcdoubleAssignNewXV2` hint (secp hint)


TAdev0:
https://github.com/lambdaclass/cairo-vm/blob/db7fff18c9c1312024ebf4119fcdfeba23657b12/vm/src/hint_processor/builtin_hint_processor/hint_code.rs#L776

https://github.com/lambdaclass/cairo-vm/blob/db7fff18c9c1312024ebf4119fcdfeba23657b12/cairo_programs/efficient_secp256r1_ec.cairo#L34

https://github.com/starkware-libs/cairo-lang/blob/0e4dab8a6065d80d1c726394f5d9d23cb451706a/src/starkware/starknet/security/whitelists/cairo_secp.json#L133

danielcdz:
Hello @TAdev0 I would like to help with this one! I know that I already have an issue assigned but I already opened a PR for that one

TAdev0:
Let's go !

danielcdz:
@TAdev0 Thank you sir! 🫡

danielcdz:
@TAdev0 this [PR](https://github.com/NethermindEth/cairo-vm-go/pull/601) that solves this issue is ready for review!

PR #601, opened 2024-07-23
Feat `EcdoubleAssignNewXV2` hint (secp hint)

danielcdz:
Resolves issue #538


pkg/hintrunner/zero/zerohint_ec.go
```diff
@@ -566,8 +566,8 @@ func createReduceEd25519Hinter(resolver hintReferenceResolver) (hinter.Hinter, e
 // It also assigns `slope`, `x`, `y` and `new_x` in the current scope
 // so that they are available in the current scope for EcDoubleAssignNewYV1 hint
 //
-// This implementation is valid for both EcDoubleAssignNewX V1 and V4, only the operander differs
-// with `point` used for V1 and `pt` used for V4
+// This implementation is valid for EcDoubleAssignNewX V1,V2 and V4, only the operander differs
+// with `point` used for V1,V2 and `pt` used for V4
```
TAdev0, line 570: Can you add that with V2 , SECP is supposed to be already in scope (contrary to V1 which adds it in scope as well

TAdev0 (changes requested):

pkg/hintrunner/zero/zerohint_ec.go
danielcdz: Done! ✅

PR #606, opened 2024-07-25
Implement `SplitXX` hint

MaksymMalicki:


pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -598,3 +598,91 @@ func createUint256SubHinter(resolver hintReferenceResolver) (hinter.Hinter, erro
 
 	return newUint256SubHint(a, b, res), nil
 }
+
```
TAdev0, line 601: can you add a comment for the hint?

pkg/hintrunner/zero/zerohint_uint256.go
```diff
@@ -598,3 +598,91 @@ func createUint256SubHinter(resolver hintReferenceResolver) (hinter.Hinter, erro
...
+			//> xx = ids.xx.low + (ids.xx.high<<128)
+			//> x = pow(xx, (PRIME + 3) // 8, PRIME)
+			//> if (x * x - xx) % PRIME != 0:
+			//> 	x = (x * II) % PRIME
+			//> if x % 2 != 0:
+			//>   	x = PRIME - x
+			//> ids.x.low = x & ((1<<128)-1)
+			//> ids.x.high = x >> 128
+
+			PRIME, ok := new(big.Int).SetString("57896044618658097711785492504343953926634992332820282019728792003956564819949", 10)
```
TAdev0, line 618: here you can use `GetCurve25519PBig` method from secp_utils to retrieve this value

MaksymMalicki: done

PR #608, opened 2024-07-25
Faster integration tests

har777:
Runs tests concurrently using goroutines if benchmarking is not required.


.gitignore
```diff
@@ -4,6 +4,7 @@ benchmarks
 cov.out
 cpu.out
 *.test
+integration_tests/BenchMarks.txt
```
TAdev0, line 7: why do you add the benchmarks file in gitignore now we have `make zerobench` command to avoid updating the file when running integration tests?

integration_tests/BenchMarks.txt
TAdev0, line 1: file deleted?

.gitignore
har777: yeah i was thinking we shouldn't be committing benchmarks and should just run and check it locally.

TAdev0: isnt it nice to have a benchmark file in the repo that anyone reading the repo can see?

MaksymMalicki: I remember, that during the daily we talked, that once we finish implementing the hints, we will add the benchmarks to the README, so I agree that we put it to the git ignore for now

TAdev0: Oh ok got it !

MaksymMalicki:
So right now only 5 goroutines can be run simultaneously? Is that number arbitrary? Did you test bigger / smaller options and how they affect performance?

MaksymMalicki (approved):

integration_tests/cairozero_test.go
```diff
@@ -67,6 +125,9 @@ func TestCairoZeroFiles(t *testing.T) {
 
 	benchmarkMap := make(map[string][2]int)
 
+	sem := make(chan struct{}, 5) // semaphore to limit concurrency
```
MaksymMalicki, line 128: So right now only 5 goroutines can be run simultaneously? Is that number arbitrary? Did you test bigger / smaller options and how they affect performance?

TAdev0: i think Hari said he wanted to start with only a few, see how CI behaves, and then increment?

har777: It's kinda arbitrary. I didn't see a huge improvement after 5 goroutines and wanted to be conservative for our CI.

TAdev0 (approved):

PR #610, opened 2024-07-25
Feat `Blake2sCompress` hint

### LINKED ISSUE #539: `Blake2sCompress` hint
https://github.com/lambdaclass/cairo-vm/blob/db7fff18c9c1312024ebf4119fcdfeba23657b12/vm/src/hint_processor/builtin_hint_processor/hint_code.rs#L488

https://github.com/lambdaclass/cairo-vm/blob/db7fff18c9c1312024ebf4119fcdfeba23657b12/cairo_programs/example_blake2s.cairo#L60

https://github.com/starkware-libs/cairo-lang/blob/0e4dab8a6065d80d1c726394f5d9d23cb451706a/src/starkware/starknet/security/whitelists/cairo_blake2s.json#L84

danielcdz:
## Summary

Closes #539


danielcdz:
I want to know if I'm storing the result of new_state accordingly or if there is a better way to do it 🤔

pkg/hintrunner/zero/zerohint_blake.go
```diff
@@ -428,3 +428,101 @@ func createBlake2sComputeHinter(resolver hintReferenceResolver) (hinter.Hinter,
...
+			}
+			var messageUint32 []uint32
+			for i := 0; i < 16; i++ {
+				value, err := hintrunnerUtils.ToSafeUint32(&message[i])
+				if err != nil {
+					return err
+				}
+				messageUint32 = append(messageUint32, value)
+			}
+			h := utils.IV()
```
har777, line 491: `h=[IV[0] ^ 0x01010020] + IV[1:]`
I believe you should modify the first value of the returned IV like `IV[0] ^ 0x01010020`
so something like `h[0] = h[0] ^ 0x01010020`

danielcdz: that's true, is fixed now!

TAdev0:
@danielcdz  can you also add unit tests and the corresponding cairo file ?

danielcdz:
@TAdev0 sure thing!

danielcdz:
I'm having this issue when trying to implement the unit test: 
image

MaksymMalicki:
> I'm having this issue when trying to implement the unit test: image

Hey @danielcdz it's an issue somewhere in the unittests, when creating `addr(7)`.  I see two problems:
1. You need to treat `n_bytes` as number (felt) instead of pointer to memory cell 1:7, because `t0` argument is expected to be a number.
2. You pass a 1:7 addressto `blake2s_start`, which is correct, but you don't populate the memory with any values. Thats probably what causes the issue, you are reading from 1:7, but there is no value inside. You might treat [this as reference](https://github.com/NethermindEth/cairo-vm-go/blob/a55dae8d65d234b6e5a6b2d53cfbfb3d096c377d/pkg/hintrunner/zero/zerohint_blake_test.go#L495)

danielcdz:
> > I'm having this issue when trying to implement the unit test: image
> 
> Hey @danielcdz it's an issue somewhere in the unittests, when creating `addr(7)`. I see two problems:
> 
> 1. You need to treat `n_bytes` as number (felt) instead of pointer to memory cell 1:7, because `t0` argument is expected to be a number.
> 2. You pass a 1:7 addressto `blake2s_start`, which is correct, but you don't populate the memory with any values. Thats probably what causes the issue, you are reading from 1:7, but there is no value inside. You might treat [this as reference](https://github.com/NethermindEth/cairo-vm-go/blob/a55dae8d65d234b6e5a6b2d53cfbfb3d096c377d/pkg/hintrunner/zero/zerohint_blake_test.go#L495)

Thanks for the help @MaksymMalicki I'll take a look into it 🫡

TAdev0:
@danielcdz can you also move blake_finalize_v3.starknet_with_keccak.cairo file to the folder not in progress? so that we make sure integration tests pass when your hint is involved

danielcdz:
> @danielcdz can you also move blake_finalize_v3.starknet_with_keccak.cairo file to the folder not in progress? so that we make sure integration tests pass when your hint is involved

@TAdev0 Sure! By not in progress you refer to cairo_zero_hint_tests?

TAdev0:
Yes

danielcdz:
@MaksymMalicki @TAdev0 I updated the unit test following the example of `Blake2sCompute` as suggested, I think I'm missing or doing something wrong when populating the memory values cause I'm receiving this error: 
image

danielcdz:
@TAdev0 this should be good to go! all integration and unit tests are passing

MaksymMalicki (approved):

MaksymMalicki:
Approve! Good job 🚀

danielcdz:
> Approve! Good job 🚀

Thank you for the help! you guys are awesome 💯 @TAdev0 @MaksymMalicki @har777

PR #622, opened 2024-08-03
feat:  `Sha256` utils tests + chore improvements

TAdev0:
Also created folder and files for cairo 1 runner
I named the Package `Cairo` with the cairo zero runner being in the `Zero` package.  Dont know if `Cairo` is the best naming, maybe `Cairo1` (but now its Cairo 2 actually, I think Cairo vs Cairo Zero is a good terminology, @zmalatrax  uses this naming)


pkg/utils/math_test.go
```diff
@@ -64,3 +64,58 @@ func TestFeltDivRem(t *testing.T) {
 		}
 	}
 }
+
```
TAdev0, line 67: moved this here because its used for Blake but also for Sha2

MaksymMalicki (approved):

MaksymMalicki:
I like this idea of naming `cairo0` runner zero and `cairo` for the future cairo versions. Gave approve 😉

TAdev0:
ok for now its `Cairo` and `Zero` right?

PR #623, opened 2024-08-05
Poseidon integration tests fix

MaksymMalicki:


MaksymMalicki:
After comparing our memory with rust I've noticed, that in lambdaclass they are only infering the memory values for the segment addresses, which were found during getting op0 and op1. The memory values under those addresses are initially of type unknown. They decided to store all the produced hash inputs and outputs in a separate cache (address - mv map) and then populate only the memory values under the relevant addresses. What we were doing (and still are in different builtins), is that we are immiadetely writing the produced hash to the memory instead of cache, populating also the addresses which are apparently not relevant to the execution. That was the reason of the difference in the traces.

For instance in the screenshot below, I printed only the addresses found during getting op0 and op1 , which should infer the values for.
image
By writing the entire hash to the memory, we populated the holes between addresses (like 9:3 - 9:9) and thus the differences in the trace.

If you guys agree with the usage of cache, we should implement it for the `ec_op` and `keccak`.

har777 (approved):

pkg/vm/builtins/ecop.go
```diff
@@ -36,6 +37,12 @@ func (e *EcOp) CheckWrite(segment *mem.Segment, offset uint64, value *mem.Memory
 }
 
 func (e *EcOp) InferValue(segment *mem.Segment, offset uint64) error {
+	// check if the value is already in the cache
+	value, ok := e.cache[offset]
+	if ok {
+		mv := mem.MemoryValueFromFieldElement(&value)
+		return segment.Write(offset, &mv)
+	}
```
har777, line 45: Is it possible that `InferValue` is called on the same offset multiple times? Because in that case the value might already be written in the segment.

TAdev0 (approved):
lgtm!

PR #625, opened 2024-08-12
Hint reference parsing

### LINKED ISSUE #548: Missing expression in parser
Expressions of the kind `cast(1, felt)` seem to be missing from the parser.

Sample file which can fail: https://github.com/lambdaclass/cairo-vm/blob/main/cairo_programs/bitand_hint.cairo

cicr99:
Fix #548 

The changes in the grammar:
- allow more complex expression to be parsed. For example, more than only two terms to be operated
- include missing cases such as:
    - a simple number as the expression to be casted
    - double dereference inside the expression
- parse the numbers inside the expressions as a string as to allow numbers bigger than 64 bits


pkg/utils/math.go
```diff
@@ -126,6 +127,32 @@ func FeltDivRem(a, b *fp.Element) (div fp.Element, rem fp.Element) {
...
 }
 
+func Int16FromFelt(n *fp.Element) (int16, bool) {
+	bigN := n.BigInt(new(big.Int))
+	return Int16FromBigInt(bigN)
+}
+
+func Int16FromBigInt(n *big.Int) (int16, bool) {
+	mod := fp.Modulus()
+	negN := new(big.Int).Sub(mod, n)
```
Sh0g0-1758, line 137: Can you please explain this implementation. I don't quite understand why you have to take fp_mod - N. suppose that N is bigger than NegN, does that not imply that N is bigger than maxInt16 ?

cicr99: `n` is representing a felt value, so they don't actually have negative numbers, but we know that if `n > P/2 `then we could use the negative value which is `n - P` (`P` in this case is `mod`). As we want to take the int16 value, I'm just trying to obtain the smaller number possible, so you check which one is smaller, n or negN, and then you compare that one to the highest possible value of int16 to check if the number fits in that type. Does it make sense?

pkg/hintrunner/zero/hintparser.go
```diff
@@ -122,80 +148,179 @@ func (expression DerefCastExp) Evaluate() (hinter.Reference, error) {
...
+		}
+		return leftResult, nil
+	} else {
+		for _, term := range expression.AddExp {
+			rightExp, err := term.TermExp.Evaluate()
+			if err != nil {
+				return nil, err
+			}
+
+			op, err := parseOperator(term.Operator)
```
Sh0g0-1758, line 221: Can the operator never be "-" here ?

pkg/utils/math.go
Sh0g0-1758: Ah yes that makes sense now. Thanks.

Sh0g0-1758 (approved):
LGTM

pkg/hintrunner/zero/hintparser.go
```diff
@@ -122,80 +148,179 @@ func (expression DerefCastExp) Evaluate() (hinter.Reference, error) {
...
+			return nil, err
+		}
+		return hinter.Immediate(*new(fp.Element).SetBigInt(intExp)), nil
 	case expression.CellRefExp != nil:
 		return expression.CellRefExp.Evaluate()
 	case expression.DerefExp != nil:
 		return expression.DerefExp.Evaluate()
-	case expression.BinOpExp != nil:
-		return expression.BinOpExp.Evaluate()
+	case expression.DDerefExp != nil:
```
MaksymMalicki, line 288: I don't quite understand, why could we replace parsing Binary operator with double deref here?

MaksymMalicki (approved):

pkg/hintrunner/zero/hintparser.go
cicr99: So it's not exactly that one is being replaced by the other, the entire grammar changed, so now the Expression type takes the "atomic" expressions of operanders: Immediate, CellRef, Deref and DoubleDeref. For BinOp, notice that with this grammar we can now have several terms in a sumation instead of only 2, so we introduced the type ArithmeticExpression in the grammar, which can be evaluated into Binary Operations. Also, for allowing multiple terms, now the left and right hand side of the binary operation are resOperanders, which allows the structure to be recursive.
For example, the expression: `"cast([ap + 3] + [ap + (-4)] * 18 + 15, felt)"` would be evaluated as:
```
hinter.BinaryOp{
    Operator: hinter.Add,
    Lhs: hinter.BinaryOp{
        Operator: hinter.Add,
        Lhs:      hinter.Deref{
             Deref: hinter.ApCellRef(3),
        },
        Rhs: hinter.BinaryOp{
             Operator: hinter.Mul,
             Lhs: hinter.Deref{
                  Deref: hinter.ApCellRef(-4),
             },
            Rhs: hinter.Immediate(18),
        },
    },
    Rhs: hinter.Immediate(15),
}
```

PR #627, opened 2024-08-13
Implement hint map

MaksymMalicki:


cmd/cli/main.go
cicr99, line 1: I've created an issue for updating this file, once we have more information on our side https://github.com/NethermindEth/cairo-vm-go/issues/638. Let's leave it as this for now

cicr99 (commented):
Looks very good

cicr99 (approved):

cicr99:
One last thought. Maybe we should add some unit tests for this parser, similar to `pkg/hintrunner/zero/hintparser_test.go`

pkg/hintrunner/core/cairo_hintparser_test.go
cicr99, line 1: Nice that you added a lot of cases. I do have one comment though: 
Having all the cases in the same test means that if one fails, it's more difficult to know which one it was, specially when there are so many, or if you were to add some more cases. You want to test the parsing is being done correctly, so it's good if you provide each case separately and then check the result is as expected. Similar to the `hintparser_test` file for cairo zero where you provide the string and check the resulting operander. The way you are doing it right now is actually testing a lot more of the pipeline, which is good, but this would be tested on integration tests as well, when we add them. Any thoughts on this?
I'd leave like this rn, and change the tests (if we consider it's best) in a different PR. This PR is ready to be merged.

cicr99 (approved):

pkg/hintrunner/core/cairo_hintparser_test.go
MaksymMalicki: Yeah, thats probably a good idea. I was more focused on feeding big file to the test to simulate real case scenario, multiple instances of the same hint etc. But providing the smaller instances is obviously a good idea

Issue #628, opened 2024-08-13
Feat : Implement `TestLessThanOrEqualAddress` hint


Sh0g0-1758:

bitfalt:
Hi! I'm an active contributor to OnlyDust. I've done contributions using Cairo and recently started involving myself in Hackathons. I was wondering if this issue is open for contributors, if so I'd like to apply. I would do the following:
- Learn how other hints were implemented throughout the project
- Implement the hint in `pkg/hintrunner/hint.go`
- Document the hint implementation properly
- Make the proper tests to ensure proper behaviour of the hint.

bitfalt:
No worries, I'll be waiting for the other issues!

PR #629, opened 2024-08-14
Implemented `TestLessThanOrEqualAddress`

Sh0g0-1758:
Fixes: #628

### DIFF
--- a/pkg/hintrunner/core/hint.go
+++ b/pkg/hintrunner/core/hint.go
@@ -151,18 +151,18 @@ func (hint *TestLessThanOrEqualAddress) String() string {
 }
 
 func (hint *TestLessThanOrEqualAddress) Execute(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
-	lhsAddr, err := hint.lhs.GetAddress(vm)
+	lhsPtr, err := hinter.ResolveAsAddress(vm, hint.lhs)
 	if err != nil {
-		return fmt.Errorf("GetAddress lhs operand %s: %w", hint.lhs, err)
+		return fmt.Errorf("resolve lhs pointer: %w", err)
 	}
 
-	rhsAddr, err := hint.rhs.GetAddress(vm)
+	rhsPtr, err := hinter.ResolveAsAddress(vm, hint.rhs)
 	if err != nil {
-		return fmt.Errorf("GetAddress rhs operand %s: %w", hint.rhs, err)
+		return fmt.Errorf("resolve rhs pointer: %w", err)
 	}
 
 	resFelt := f.Element{}
-	if lhsAddr.SegmentIndex <= rhsAddr.SegmentIndex && lhsAddr.Offset <= rhsAddr.Offset {
+	if lhsPtr.SegmentIndex < rhsPtr.SegmentIndex || (lhsPtr.SegmentIndex == rhsPtr.SegmentIndex && lhsPtr.Offset <= rhsPtr.Offset) {
 		resFelt.SetOne()
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/hintrunner/core/hint.go
```diff
@@ -140,6 +140,46 @@ func (hint *TestLessThanOrEqual) Execute(vm *VM.VirtualMachine, _ *hinter.HintRu
...
+
+	rhsAddr, err := hint.rhs.GetAddress(vm)
+	if err != nil {
+		return fmt.Errorf("GetAddress rhs operand %s: %w", hint.rhs, err)
+	}
+
+	resFelt := f.Element{}
+	if lhsAddr.SegmentIndex <= rhsAddr.SegmentIndex && lhsAddr.Offset <= rhsAddr.Offset {
+		resFelt.SetOne()
+	}
```
cicr99, line 167: I think you still have to access the value in `lhs` and `rhs`, in this case, they should be treated as addresses, so probably using `ResolveAsAddress` method should work here. If you use `GetAddress` here, you'll be getting the pointer to the location in memory where the value of `rhs`(`lhs`) is, instead of the actual value

pkg/hintrunner/core/hint.go
```diff
@@ -140,6 +140,46 @@ func (hint *TestLessThanOrEqual) Execute(vm *VM.VirtualMachine, _ *hinter.HintRu
...
+	dst hinter.CellRefer
+	lhs hinter.ResOperander
+	rhs hinter.ResOperander
+}
+
+func (hint *TestLessThanOrEqualAddress) String() string {
+	return "TestLessThanOrEqualAddress"
+}
+
+func (hint *TestLessThanOrEqualAddress) Execute(vm *VM.VirtualMachine, _ *hinter.HintRunnerContext) error {
```
cicr99, line 153: it might be useful to check `TestLessThanOrEqual` hint implementation. I see in the rust vm the execution method is the same for both, not necessarily has to be the case for us, but it's definitely worth taking a look into it to see if we can reuse some of the existing code.

Sh0g0-1758: Right the implementation in rust vm is same because the function is only supposed to compare the values, the values that are passed in the hints are resolved as ints or as addresses before that. They have made structs for them `ResOperandAsAddressFormatter` and `ResOperandAsIntegerFormatter`. I think our method `ResolveAsAddress` should achieve the same functionality.

pkg/hintrunner/core/hint.go
Sh0g0-1758: Made the changes.

pkg/hintrunner/core/hint.go
```diff
@@ -140,6 +140,46 @@ func (hint *TestLessThanOrEqual) Execute(vm *VM.VirtualMachine, _ *hinter.HintRu
...
+
+	rhsPtr, err := hinter.ResolveAsAddress(vm, hint.rhs)
+	if err != nil {
+		return fmt.Errorf("resolve rhs pointer: %w", err)
+	}
+
+	resFelt := f.Element{}
+	if lhsPtr.SegmentIndex < rhsPtr.SegmentIndex || (lhsPtr.SegmentIndex == rhsPtr.SegmentIndex && lhsPtr.Offset <= rhsPtr.Offset) {
+		resFelt.SetOne()
+	}
```
cicr99, line 167: I was thinking it might be a good idea to add a Compare functionality to the MemoryAddress type. Then here you would only have to ask for the result of the comparison, which is more readable.

cicr99 (approved):
LGTM

pkg/hintrunner/core/hint.go
Sh0g0-1758: That's a nice suggestion. Made the changes.

PR #648, opened 2024-08-29
Add cairo1 runner

MaksymMalicki:


TAdev0:
@MaksymMalicki  how is it going with this?
I see you made modification to the zero.go file, whats your plan exactly?

cmd/cli/main.go
```diff
@@ -102,144 +102,125 @@ func main() {
...
-					// TODO: move this action's body to a separate function to decrease the
-					// code nesting significantly.
-
 					pathToFile := ctx.Args().Get(0)
 					if pathToFile == "" {
 						return fmt.Errorf("path to cairo file not set")
 					}
+
+					var hints map[uint64][]hinter.Hinter
+					var program *runner.Program
```
cicr99, line 111: Is there any reason to define the variables here now that you've removed the if statement below? I think in Go, it's preferred defining the variables at the point of initialization, unless you need them to be accessible outside the specific block, or if you need default values before conditional initialization, and such cases

MaksymMalicki: Yeah thats right

pkg/runner/runner.go
```diff
@@ -126,8 +139,24 @@ func (runner *ZeroRunner) initializeMainEntrypoint() (mem.MemoryAddress, error)
...
+		}
+		return runner.initializeEntrypoint(mainPCOffset, nil, &mvReturnFp, memory, stack, 0)
+	case ProofModeCairo1:
+		returnFp := memory.AllocateEmptySegment()
+		mvReturnFp := mem.MemoryValueFromMemoryAddress(&returnFp)
+		mainPCOffset, ok := runner.program.Entrypoints["main"]
+		if !ok {
+			return mem.UnknownAddress, errors.New("can't find an entrypoint for main")
+		}
+		return runner.initializeEntrypoint(mainPCOffset, nil, &mvReturnFp, memory, stack, 2)
```
cicr99, line 158: It's the same logic for both I see except for the `cairo1FpOffset` you are passing to initializeEntrypoint function, am I correct? Would it be better to unify these two and pass `cairo1FpOffset` as a parameter instead of the direct value?

MaksymMalicki: Yeah, thats good idea 💡

cicr99 (approved):

PR #649, opened 2024-09-05
feat: implement `dex` layout

TAdev0:
This PR implements `dex` layout, which is very similar to the small layout , only changes the `RcUnits` value for the layout, ie, 4 vs 16


har777 (approved):
lgtm.

PR #651, opened 2024-09-05
feat: implement `starknet` layout

TAdev0:
This PR implements the `starknet` layout, which is identical to starknet_with_keccak except there is no keccak builtin in it

Note that i used the suffix `starknet_without_keccak` , otherwise `strings.Contains()` would induce a conflict with `starknet_with_keccak` if we call it only `starknet`

### DIFF
--- a/integration_tests/cairozero_test.go
+++ b/integration_tests/cairozero_test.go
@@ -288,12 +288,12 @@ func runPythonVm(testFilename, path string) (time.Duration, string, string, erro
 	// A file without this suffix will use the default ("plain") layout.
 	if strings.HasSuffix(testFilename, ".small.cairo") {
 		args = append(args, "--layout", "small")
-	} else if strings.HasSuffix(testFilename, ".starknet_without_keccak.cairo") {
-		args = append(args, "--layout", "starknet")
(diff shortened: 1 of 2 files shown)


integration_tests/cairozero_test.go
```diff
@@ -321,6 +323,8 @@ func runVm(path string) (time.Duration, string, string, string, error) {
 	layout := "plain"
 	if strings.Contains(path, ".small") {
 		layout = "small"
+	} else if strings.Contains(path, ".starknet_without_keccak") {
+		layout = "starknet"
 	} else if strings.Contains(path, ".starknet_with_keccak") {
 		layout = "starknet_with_keccak"
 	}
```
har777, line 330: You can just check `.starknet_with_keccak` first and then check for `.starknet` to avoid conflicts. You don't need the `_without_keccak` suffix.

har777 (approved):
lgtm other than minor comment above.

integration_tests/cairozero_test.go
TAdev0: true! wanted to respect rust vm order but it strictly doesnt matter, will do that!

PR #652, opened 2024-09-05
feat: implement `recursive_large_output` layout

TAdev0:
This PR implements the `recursive_large_output` layout, which contains : 
- output
- pedersen with 128 ratio
- rangecheck default (ie 8 ratio)
- bitwise with 8 ratio
- poseidon with 8 ratio


har777 (approved):
lgtm.

PR #654, opened 2024-09-05
feat: reorganize `runner` package

TAdev0:
This is a proposition to reorganize the runner folder

Until now, we had this separation of Zero / Cairo1 , but actually most of the logic in zero.go  (all the checks, endRun etc) is common to both Zero and Cairo.

Feel free to leave comment and share your thoughts about it


har777 (approved):
broadly looks fine to me

PR #662, opened 2024-09-13
Feat built in name literals

### LINKED ISSUE #660: Builtin name literals
Currently there are several places in the code where the name of a builtin is used as a literal string. For example [here](https://github.com/NethermindEth/cairo-vm-go/blob/4d9322770b8b07cb1a5dba4bef6a0f647bc805ce/pkg/parsers/starknet/starknet.go#L61). Notice that having to write the string each time could cause unwanted typos, and lead to errors in the code.

All the builtin names should be defined as variables (in the `builtin` module most likely), and all the places where the literal strings are used should be replaced with such variables.

danielcdz:
## Summary 
- Closes #660 
- Use name variables
- Add missing names
- Group constants(suggested improvement)

### DIFF
--- a/pkg/hintrunner/zero/zerohint_math_test.go
+++ b/pkg/hintrunner/zero/zerohint_math_test.go
@@ -6,8 +6,8 @@ import (
 	"testing"
 
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
-	builtins "github.com/NethermindEth/cairo-vm-go/pkg/vm/builtins"
 	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
+	builtins "github.com/NethermindEth/cairo-vm-go/pkg/vm/builtins"
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
 )
 
@@ -445,8 +445,8 @@ func TestZeroHintMath(t *testing.T) {
 		"Assert250bits": {
 			{
 				operanders: []*hintOperander{
-					{Name: "low", Kind: reference, Value: addrBuiltin(builtins.RangeCheckEnum, 0)},
-					{Name: "high", Kind: reference, Value: addrBuiltin(builtins.RangeCheckEnum, 1)},
+					{Name: "low", Kind: reference, Value: addrBuiltin(builtins.RangeCheckType, 0)},
+					{Name: "high", Kind: reference, Value: addrBuiltin(builtins.RangeCheckType, 1)},
 					{Name: "value", Kind: apRelative, Value: feltInt64(3042)},
 				},
 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
@@ -459,8 +459,8 @@ func TestZeroHintMath(t *testing.T) {
 			},
 			{
 				operanders: []*hintOperander{
-					{Name: "low", Kind: reference, Value: addrBuiltin(builtins.RangeCheckEnum, 0)},
-					{Name: "high", Kind: reference, Value: addrBuiltin(builtins.RangeCheckEnum, 1)},
+					{Name: "low", Kind: reference, Value: addrBuiltin(builtins.RangeCheckType, 0)},
+					{Name: "high", Kind: reference, Value: addrBuiltin(builtins.RangeCheckType, 1)},
 					{Name: "value", Kind: fpRelative, Value: feltInt64(4938538853994)},
 				},
 				makeHinter: func(ctx *hintTestContext) hinter.Hinter {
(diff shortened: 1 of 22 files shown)


danielcdz:
I wanted to use the names in `starknet.go` but this causes an import cycle between this file and builtins, which can be the best approach here?

cicr99:
Hi @danielcdz . Sorry for the late reply!
I see what you mean with the cyclic dependency issue. The reason why builtin module uses the starnetParser module is just because of this "enum" for builtins that defined in `starknet.go` file. Our suggestion for this is to move that [definition](https://github.com/NethermindEth/cairo-vm-go/blob/e51153f64482d37837209c5c6c6c5f9496729aca/pkg/parsers/starknet/starknet.go#L12) to `pkg/vm/builtins/builtin_runner.go`, so all the logic regarding builtins is in the same place. I think it would be more organized that way and should remove the issue with the dependencies. Only think you'll have to do some adjustments in the imports for the rest of the code.

cicr99 (commented):
It looks good in general, I liked that you used the builtin module to add the names, seems like the best place to put them. 
Apart from the `switch case` cases where the strings are being used, bare in mind there are other places in the code where the string is being used directly and it would be good to change that too

danielcdz:
@cicr99 working on your suggestions!

danielcdz:
I'm not 100% sure if I replaced all the strings

pkg/runner/runner.go
```diff
@@ -6,10 +6,9 @@ import (
 
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner"
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
-	"github.com/NethermindEth/cairo-vm-go/pkg/parsers/starknet"
 	"github.com/NethermindEth/cairo-vm-go/pkg/utils"
 	"github.com/NethermindEth/cairo-vm-go/pkg/vm"
-	"github.com/NethermindEth/cairo-vm-go/pkg/vm/builtins"
+	builtins "github.com/NethermindEth/cairo-vm-go/pkg/vm/builtins"
```
cicr99, line 11: No need for an alias here, right? You can already refer to the package as `builtins` by default. You would only use an alias if there was a need to resolve a conflict or if you wanted to use a different name, but neither is the case here.

pkg/parsers/zero/zero.go
```diff
@@ -4,7 +4,8 @@ import (
 	"encoding/json"
 	"os"
 
-	starknetParser "github.com/NethermindEth/cairo-vm-go/pkg/parsers/starknet"
+	builtins "github.com/NethermindEth/cairo-vm-go/pkg/vm/builtins"
```
cicr99, line 7: Same comment about the alias here

pkg/hintrunner/zero/zerohint_signature_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"testing"
 
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
-	"github.com/NethermindEth/cairo-vm-go/pkg/parsers/starknet"
+	builtins "github.com/NethermindEth/cairo-vm-go/pkg/vm/builtins"
```
cicr99, line 8: Same comment about the alias here

pkg/hintrunner/zero/zerohint_math_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"testing"
 
 	"github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/hinter"
-	"github.com/NethermindEth/cairo-vm-go/pkg/parsers/starknet"
+	builtins "github.com/NethermindEth/cairo-vm-go/pkg/vm/builtins"
```
cicr99, line 9: Same comment about the alias here

pkg/parsers/zero/zero.go
danielcdz: Done! ✅

pkg/hintrunner/zero/zerohint_signature_test.go
danielcdz: Done! ✅

pkg/hintrunner/zero/zerohint_math_test.go
danielcdz: Done! ✅

pkg/runner/runner.go
danielcdz: Done! ✅

pkg/vm/builtins/builtin_runner.go
```diff
@@ -3,33 +3,48 @@ package builtins
...
+	RangeCheckEnum
+	PedersenEnum
+	ECDSAEnum
+	KeccakEnum
+	BitwiseEnum
+	ECOPEnum
+	PoseidonEnum
+	SegmentArenaEnum
+	RangeCheck96Enum
+)
```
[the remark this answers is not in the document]
danielcdz: Done! ✅

danielcdz:
@cicr99 all comments addresed, unit and integrations tests are passing, all possible references to builtin names changed to their variables instead of comparing the strings, I think this should be ready to go 👍

cicr99 (approved):
Awesome!

Issue #632, opened 2024-08-15
Cairo one Builtins


Sh0g0-1758:
Builtins required for Cairo 1 :

- [x] #636 
- [ ] #634
- [ ] #635

PR #666, opened 2024-09-19
Implement ModBuiltin

Sh0g0-1758:
Refer : https://github.dev/lambdaclass/cairo-vm/blob/main/vm/src/vm/runners/builtin_runner/modulo.rs
Fixes: #632 
Fixes: #634 
Fixes: #635


Sh0g0-1758:
@cicr99

pkg/vm/builtins/modulo_test.go
```diff
@@ -0,0 +1,51 @@
...
+	segment.WithBuiltinRunner(mod)
+
+	v0, _ := new(fp.Element).SetString("1")
+	v1, _ := new(fp.Element).SetString("2")
+	v2, _ := new(fp.Element).SetString("3")
+	v3, _ := new(fp.Element).SetString("4")
+	v4, _ := new(fp.Element).SetString("5")
+	v5, _ := new(fp.Element).SetString("9")
+	v6, _ := new(fp.Element).SetString("7")
+	v7, _ := new(fp.Element).SetString("8")
```
cicr99, line 23: Pls avoid using `SetString` if you could use `SetUint64` instead. Leave the string method for a big number (that doesn't fit in uint64)

cicr99:
There are conflicts after merging #662 that need to be fixed. This PR creates variables for the name of the builtins and replace all places in the code where the raw string is used with such variables. You'll have to do the same for the new builtin. Also there were some refactors done in the code that affect your branch

pkg/vm/builtins/modulo.go
```diff
@@ -0,0 +1,477 @@
...
+	batchSize      uint64
+	shift          big.Int
+	shiftPowers    [N_WORDS]big.Int
+}
+
+func NewModBuiltin(ratio uint64, modBuiltinType ModBuiltinType, wordBitLen uint64, batchSize uint64) *ModBuiltin {
+	shift := new(big.Int).Lsh(big.NewInt(1), uint(wordBitLen))
+	shiftPowers := [N_WORDS]big.Int{}
+	for i := 0; i < N_WORDS; i++ {
+		shiftPowers[i].Exp(shift, big.NewInt(int64(i)), nil)
```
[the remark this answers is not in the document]
Sh0g0-1758: ah yes indeed. I also thought of this yesterday when I was writing the tests.

pkg/vm/builtins/modulo.go
```diff
@@ -0,0 +1,542 @@
...
+	switch {
+	case a != nil && b != nil && c == nil:
+		var value big.Int
+		if op == Add {
+			value = *new(big.Int).Add(a, b)
+		} else {
+			value = *new(big.Int).Mul(a, b)
+		}
+		if new(big.Int).Sub(&value, new(big.Int).Mul((new(big.Int).Sub(kBound, big.NewInt(1))), &inputs.p)).Cmp(new(big.Int).Sub(intLim, big.NewInt(1))) == 1 {
+			return false, fmt.Errorf("%s builtin: Expected a %s b - %d * p <= %d", m.String(), m.modBuiltinType, kBound.Sub(kBound, big.NewInt(1)), intLim.Sub(intLim, big.NewInt(1)))
```
cicr99, line 365: omg this is so difficult to read 😅 . Is it possible to separate the values? I see you use some of them in the error message, perhaps have it as separate variables so the code is easier to understand

cicr99 (approved):

pkg/vm/builtins/modulo.go
Sh0g0-1758: But those variables won't really be used anywhere else. I suppose I can write a comment on top of it that represents the mathematical form of this expression.

Sh0g0-1758: Okay. Merging this now.

Issue #667, opened 2024-09-25
Implement tests for the mod builtin


Sh0g0-1758:
Modbuiltin is going to be added by this [PR](#666). The tests for the same have not yet been implemented. Refer cairo-lang for the python tests.

Issue #637, opened 2024-08-15
Implement `EvalCircuit` hint


MaksymMalicki:
Requires the basic implementation of `AddModBuiltin` and `MulModBuiltin`

https://github.com/starkware-libs/cairo/blob/95fd09bab0cc44c4f101987c62491454890e3ba4/crates/cairo-lang-runner/src/casm_run/mod.rs#L2087

https://github.com/starkware-libs/cairo/blob/54804c8730dabea31a2b80034fe2c3911e79adf5/crates/cairo-lang-casm/src/hints/mod.rs#L289

Sh0g0-1758:
refer : 

```
            CoreHint::EvalCircuit { n_add_mods, add_mod_builtin, n_mul_mods, mul_mod_builtin } => {
                let n_add_mods = ResOperandAsIntegerFormatter(n_add_mods);
                let add_mod_builtin = ResOperandAsAddressFormatter(add_mod_builtin);
                let n_mul_mods = ResOperandAsIntegerFormatter(n_mul_mods);
                let mul_mod_builtin = ResOperandAsAddressFormatter(mul_mod_builtin);
                formatdoc! {"

                from starkware.cairo.lang.builtins.modulo.mod_builtin_runner import ModBuiltinRunner

                ModBuiltinRunner.fill_memory(
                        memory=memory,
                        add_mod=({add_mod_builtin}, builtin_runners[\"add_mod_builtin\"], {n_add_mods}),
                        mul_mod=({mul_mod_builtin}, builtin_runners[\"mul_mod_builtin\"], {n_mul_mods}),
                )
                "}
            }
```

PR #669, opened 2024-10-01
Implemented EvalCircuit, Fixed bugs and wrote tests for Modbuiltin

Sh0g0-1758:
Fixes: #667 
Fixes: #637 
Fixes: #36

### DIFF
--- a/pkg/vm/builtins/modulo.go
+++ b/pkg/vm/builtins/modulo.go
@@ -573,7 +573,9 @@ func FillMemory(mem *memory.Memory, addModInputAddress memory.MemoryAddress, nAd
 					nComputedMulGates = nMulMods
 				}
 				mulModBuiltinInputs.n = nComputedMulGates
-				mulModBuiltinRunner.fillOffsets(mem, mulModBuiltinInputs.offsetsPtr, nMulMods, nComputedMulGates-nMulMods)
+				if err := mulModBuiltinRunner.fillOffsets(mem, mulModBuiltinInputs.offsetsPtr, nMulMods, nComputedMulGates-nMulMods); err != nil {
+					return err
+				}
 			} else {
 				if mulModBuiltinRunner.batchSize != 1 {
 					return fmt.Errorf("MulMod builtin: Inverse failure is supported only at batch_size == 1")
@@ -589,7 +591,9 @@ func FillMemory(mem *memory.Memory, addModInputAddress memory.MemoryAddress, nAd
 				return err
 			}
 
-			mulModBuiltinRunner.fillInputs(mem, mulModInputAddress, mulModBuiltinInputs)
+			if err := mulModBuiltinRunner.fillInputs(mem, mulModInputAddress, mulModBuiltinInputs); err != nil {
+				return err
+			}
 		}
 	}
 	return nil
--- a/pkg/vm/builtins/modulo.go
+++ b/pkg/vm/builtins/modulo.go
@@ -352,7 +352,6 @@ func (m *ModBuiltin) fillValue(mem *memory.Memory, inputs ModBuiltinInputs, inde
 	if kBound == nil {
 		kBound = new(big.Int).Set(intLim)
 	}
-
 	switch {
 	case a != nil && b != nil && c == nil:
 		var value big.Int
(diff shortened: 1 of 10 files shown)


pkg/vm/builtins/modulo.go
```diff
@@ -320,27 +321,32 @@ func (m *ModBuiltin) writeNWordsValue(mem *memory.Memory, addr memory.MemoryAddr
 // Returns true on success or if all values are already known.
```
cicr99, line 321: Shouldn't this comment change now that you changed the return value?

pkg/vm/builtins/modulo.go
```diff
@@ -320,27 +321,32 @@ func (m *ModBuiltin) writeNWordsValue(mem *memory.Memory, addr memory.MemoryAddr
...
+			return 0, err
 		}
 		offset := offsetFelt.Uint64()
 		addr, err = inputs.valuesPtr.AddOffset(int16(offset))
 		if err != nil {
-			return false, err
+			return 0, err
 		}
 		addresses = append(addresses, addr)
 		// do not check for error, as the value might not be in memory
```
cicr99, line 343: This comment should be removed now right?

pkg/vm/builtins/modulo.go
cicr99: I see you are returning different possible values, it'd be nice to explain what do they mean

pkg/vm/builtins/modulo.go
```diff
@@ -464,78 +479,122 @@ func (m *ModBuiltin) fillValue(mem *memory.Memory, inputs ModBuiltinInputs, inde
...
 		if err != nil {
 			return err
 		}
-		if ok {
-			addModIndex++
+		if err := addModBuiltinRunner.fillInputs(mem, addModInputAddress, addModBuiltinInputs); err != nil {
+			return err
+		}
+		if err := addModBuiltinRunner.fillOffsets(mem, addModBuiltinInputs.offsetsPtr, nAddMods, addModBuiltinInputs.n-nAddMods); err != nil {
+			return err
```
cicr99, line 531: Couldn't you add this logic in the first if? I mean, where you know that AddModBuiltinRunner is not nil. This is completely independent from the mulMod logic so I don't see why not

pkg/vm/builtins/modulo.go
```diff
@@ -464,78 +479,122 @@ func (m *ModBuiltin) fillValue(mem *memory.Memory, inputs ModBuiltinInputs, inde
...
 	}
 
-	for mulModIndex < nMulModsIndex {
-		ok, err = mulModBuiltinRunner.fillValue(mem, mulModBuiltinInputs, int(mulModIndex), Mul)
+	if mulModBuiltinRunner != nil {
+		mulModBuiltinInputs, err = mulModBuiltinRunner.readInputs(mem, mulModInputAddress, true)
 		if err != nil {
 			return err
 		}
-		if ok {
```
cicr99, line 534: Same comment for this one

pkg/vm/builtins/modulo.go
```diff
@@ -464,78 +479,122 @@ func (m *ModBuiltin) fillValue(mem *memory.Memory, inputs ModBuiltinInputs, inde
...
+			if res == 0 {
+				return fmt.Errorf("MulMod builtin: Could not fill the values table")
+			}
+			if res == 2 && nComputedMulGates == 0 {
+				nComputedMulGates = mulModIndex
+			}
 			mulModIndex++
 		}
-	}
-	// POTENTIALY: add n_computed_mul_gates features in the future
```
cicr99, line 538: I see that you had to add this feature. I suppose it was related to the bug you found. Could you add a comment explaining about why is needed or what it does?

pkg/vm/builtins/modulo.go
Sh0g0-1758: not exactly. Modified it.

pkg/vm/builtins/modulo.go
Sh0g0-1758: How do you mean ?

pkg/vm/builtins/modulo.go
Sh0g0-1758: Well no, this feature had to be implemented independently of that. Since this modifies the mulmod builtin memory segment. Though I don't understand why this was being done since there was not test for it.

cicr99 (approved):

PR #672, opened 2024-10-12
fix: fix slice init length

rusttech:
The purpose here should be to initialize a slice with a capacity of keccakStateSizeFeltsVal  and a length of 0, and then append it later. Instead of initializing the length to keccakStateSizeFeltsVal  from the beginning, the resulting slice will not meet expectations.

The online demo: https://go.dev/play/p/q1BcVCmvidW


Sh0g0-1758:
Hi, that is intentional. Please run unit tests with your changes before making a PR. If you ran them, you should have noticed that they fail with your changes which should have hinted that the current implementation in fact does meet expectations. 

Thank you for taking out your time to look into this project!

PR #680, opened 2025-01-10
Missing dict functionalities

MaksymMalicki:


pkg/hintrunner/core/hint.go
```diff
@@ -1143,7 +1143,8 @@ func (hint *Felt252DictEntryInit) Execute(vm *VM.VirtualMachine, ctx *hinter.Hin
 
 	prevValue, err := ctx.DictionaryManager.At(dictPtr, key)
 	if err != nil {
-		return fmt.Errorf("get dictionary entry: %w", err)
+		mv := mem.MemoryValueFromFieldElement(&utils.FeltZero)
+		prevValue = &mv
```
cicr99, line 1147: Why this change? Do you still need to write the zero value to memory?

pkg/hintrunner/core/hint.go
```diff
@@ -1237,7 +1238,8 @@ func (hint *InitSquashData) Execute(vm *VM.VirtualMachine, ctx *hinter.HintRunne
 	// todo(rodro): Don't know if it could be called multiple times, or
 	err := hinter.InitializeSquashedDictionaryManager(ctx)
 	if err != nil {
-		return err
+		ctx.SquashedDictionaryManager = hinter.SquashedDictionaryManager{}
+		_ = hinter.InitializeSquashedDictionaryManager(ctx)
```
cicr99, line 1242: I assume this means you can call this method even if the squashed dict has already been initialized in the past, right? Shouldn't the `todo` above be removed then?

pkg/runner/gas.go
```diff
@@ -51,7 +51,7 @@ func gasInitialization(memory *mem.Memory) error {
 		return err
 	}
 
-	preCostTokenTypes := []TokenGasCost{PedersenToken, BitwiseToken, EcOpToken, PoseidonToken, AddModToken, MulModToken}
+	preCostTokenTypes := []TokenGasCost{PedersenToken, PoseidonToken, BitwiseToken, EcOpToken, AddModToken, MulModToken}
```
cicr99, line 54: I'm assuming the order here is important since you are writing the costs to the memory, is that correct? Can you leave a comment here for it? Also with the source?

pkg/vm/builtins/builtin_runner.go
```diff
@@ -25,6 +25,7 @@ const (
 	AddModeType
 	MulModType
 	GasBuiltinType
+	SystemType
```
cicr99, line 28: 👀

pkg/hintrunner/core/hint.go
MaksymMalicki: this should overwrite the old manager. i agree that we can remove the todo

pkg/runner/gas.go
MaksymMalicki: yes, done

pkg/vm/builtins/builtin_runner.go
MaksymMalicki: removed

cicr99 (approved):

PR #682, opened 2025-01-17
Implement `proof_mode` functionalities

MaksymMalicki:


pkg/parsers/starknet/starknet.go
```diff
@@ -22,16 +22,22 @@ type EntryPointByType struct {
...
 
 type Arg struct {
-	GenericID string `json:"generic_id"`
+	GenericID      string         `json:"generic_id"`
+	Size           int            `json:"size"`
+	DebugName      string         `json:"debug_name"`
+	PanicInnerType PanicInnerType `json:"panic_inner_type"`
+}
+
+type PanicInnerType struct {
```
cicr99, line 31: Can we leave some kind of comment about this?

pkg/runner/runner.go
```diff
@@ -72,17 +73,22 @@ func getNewHintRunnerContext(program *Program, userArgs []starknet.CairoFuncArgs
 	//  * segment_arena_ptr + 3
 	// But the builtin itself shouldn't be included in len(builtins), therefore the writeApOffset should be increased by 3.
```
cicr99, line 74: Can you check if this comment needs to be updated after the changes in the code?

pkg/runner/runner.go
```diff
@@ -119,7 +133,21 @@ func AssembleProgram(cairoProgram *starknet.StarknetProgram, userArgs []starknet
...
-	entryCodeInstructions, entryCodeHints, err := GetEntryCodeInstructions(mainFunc)
+	entryCodeInstructions, entryCodeHints, currentCodeOffset, builtins, gotGasBuiltin, gotSegmentArena, err := GetEntryCodeInstructions(mainFunc, proofmode)
+
+	program.Builtins = builtins
+	program.GotGasBuiltin = gotGasBuiltin
+	program.GotSegmentArenaBuiltin = gotSegmentArena
+
+	if err != nil {
+		return Program{}, nil, nil, fmt.Errorf("cannot get entry code instructions: %w", err)
+	}
```
cicr99, line 144: You should do the check before setting `program` fields

pkg/runner/runner.go
```diff
@@ -221,21 +248,39 @@ func (runner *Runner) initializeMainEntrypoint() (mem.MemoryAddress, error) {
...
+		if runner.runnerMode == ExecutionModeCairo {
+			return runner.initializeEntrypoint(uint64(0), nil, &mvReturnFp, memory, stack)
 		} else {
 			mainPCOffset, ok := runner.program.Entrypoints["main"]
 			if !ok {
 				return mem.UnknownAddress, errors.New("can't find an entrypoint for main")
 			}
-			return runner.initializeEntrypoint(mainPCOffset, nil, &mvReturnFp, memory, stack, 0)
+			return runner.initializeEntrypoint(mainPCOffset, nil, &mvReturnFp, memory, stack)
+		}
```
cicr99, line 262: Perhaps it's better to separate these two, at the end you have the logic in different branches already. I think it will be more readable this way. Should we re-add the comment regarding `mainPCOffset` in Cairo?

pkg/runner/runner.go
```diff
@@ -661,17 +719,187 @@ func GetEntryCodeInstructions(function starknet.EntryPointByFunction) ([]*fp.Ele
...
+	}
+	for _, builtin := range function.Builtins {
+		adjustedRetOffset += 1
+		_, ok := builtinsOffsetsMap[builtin]
+		if ok {
+			builtinsOffsetsMap[builtin] = adjustedRetOffset
+		} else {
+			continue
+		}
+	}
```
cicr99, line 739:
```suggestion
	for _, builtin := range function.Builtins {
		adjustedRetOffset += 1
		if _, ok := builtinsOffsetsMap[builtin]; ok {
			builtinsOffsetsMap[builtin] = adjustedRetOffset
		} 
	}
```

pkg/runner/runner.go
```diff
@@ -661,17 +719,187 @@ func GetEntryCodeInstructions(function starknet.EntryPointByFunction) ([]*fp.Ele
...
+	if proofmode {
+		ctx.AddInlineCASM("jmp rel 0;")
+	} else {
+		ctx.AddInlineCASM("ret;")
 	}
-	totalSize := uint64(endInstructionsSize) + uint64(codeOffset)
-	ctx.AddInlineCASM(fmt.Sprintf("call rel %d; ret;", int(totalSize)))
-	return ctx.instructions, hints, nil
+	ctx.instructions[callRelArgLocation] = new(fp.Element).SetUint64(uint64(ctx.currentCodeOffset) + codeOffsetBeforeCallRel)
+	return ctx.instructions, hints, ctx.currentCodeOffset, programBuiltins, gotGasBuiltin, gotSegmentArena, nil
```
cicr99, line 871: Is this function returning an error ever?

pkg/runner/runner.go
```diff
@@ -661,17 +719,187 @@ func GetEntryCodeInstructions(function starknet.EntryPointByFunction) ([]*fp.Ele
...
+	isPanicResultFeltArray := false
+	if strings.Contains(mainFunc.ReturnArgs[0].DebugName, "core::panics::PanicResult::") {
+		isPanicResultFeltArray = strings.Contains(mainFunc.ReturnArgs[0].PanicInnerType.DebugName, "Array<felt252>")
+	}
+	isFeltArray := mainFunc.ReturnArgs[0].DebugName == "Array<felt252>" &&
+		mainFunc.ReturnArgs[0].Size == 2
+
+	if !isPanicResultFeltArray && !isFeltArray {
+		return fmt.Errorf("main function return argument should be either PanicResult of size 3 or Felt Array of size 2")
+	}
```
cicr99, line 903: I think there is a mismatch between the comment point 1 for PanicResult and the error at the end regarding the size: in the first case it says 2 but in the error it says 3, what's the correct one?

pkg/runner/runner.go
```diff
@@ -661,17 +719,187 @@ func GetEntryCodeInstructions(function starknet.EntryPointByFunction) ([]*fp.Ele
...
+	if mainFunc.InputArgs[0].Size != 2 || mainFunc.InputArgs[0].DebugName != "Array<felt252>" {
+		return fmt.Errorf("main function input argument should be Felt Array")
+	}
+
+	// Check if return type is either:
+	// 1. PanicResult with inner type of Array<felt252> with size 2
+	// 2. Array<felt252> with size 2
+	isPanicResultFeltArray := false
+	if strings.Contains(mainFunc.ReturnArgs[0].DebugName, "core::panics::PanicResult::") {
+		isPanicResultFeltArray = strings.Contains(mainFunc.ReturnArgs[0].PanicInnerType.DebugName, "Array<felt252>")
```
cicr99, line 896: Don't you have to check for the size here?

integration_tests/cairo1_test.go
```diff
@@ -190,22 +190,26 @@ func TestCairoFiles(t *testing.T) {
...
 					if strings.Contains(path, "proofmode") || strings.Contains(path, "serialized_output/with_input") {
 						runAndTestFile(t, path, name, benchmarkMap, *cairobench, errorExpected, inputArgs, true)
 					}
-					if !strings.Contains(path, "with_input") {
-						compareWithStarkwareRunner(t, path, errorExpected, inputArgs)
-					}
+					// compare program execution in Execution mode with starkware runner (with gas)
+					// if !strings.Contains(path, "with_input") {
+					// 	compareWithStarkwareRunner(t, path, errorExpected, inputArgs, true)
+					// }
```
cicr99, line 201: Why are these commented out?

integration_tests/cairo1_test.go
```diff
@@ -190,22 +190,26 @@ func TestCairoFiles(t *testing.T) {
...
 				if strings.Contains(path, "proofmode") || strings.Contains(path, "serialized_output/with_input") {
 					runAndTestFile(t, path, name, benchmarkMap, *cairobench, errorExpected, inputArgs, true)
 				}
-				if !strings.Contains(path, "with_input") {
-					compareWithStarkwareRunner(t, path, errorExpected, inputArgs)
-				}
+				// compare program execution in Execution mode with starkware runner (with gas)
+				// if !strings.Contains(path, "with_input") {
+				// 	compareWithStarkwareRunner(t, path, errorExpected, inputArgs, true)
+				// }
```
cicr99, line 212: Same question over here

pkg/runner/runner.go
```diff
@@ -728,6 +721,12 @@ func GetEntryCodeInstructions(function starknet.EntryPointByFunction, proofmode
 	for _, retArgs := range function.ReturnArgs {
 		adjustedRetOffset += retArgs.Size
 	}
+
+	// builtins have to be ordered by the highest id to generate proper offsets
+	for i, j := 0, len(function.Builtins)-1; i < j; i, j = i+1, j-1 {
+		function.Builtins[i], function.Builtins[j] = function.Builtins[j], function.Builtins[i]
+	}
```
cicr99, line 728: Use [`slice.Reverse`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/slices/slices.go#:~:text=493-,494,-495) instead. It does the exact same thing

MaksymMalicki: thanks! way cleaner

integration_tests/cairo1_test.go
MaksymMalicki: I wanted all the rustVM comparisons to pass first, so that after adding the builtins costs segment with condition, the output should match the starkware runner as well. But I might uncomment it as well.

integration_tests/cairo1_test.go
MaksymMalicki: I wanted all the rustVM comparisons to pass first, so that after adding the builtins costs segment with condition, the output should match the starkware runner as well. But I might uncomment it as well.

pkg/runner/runner.go
MaksymMalicki: yep, added checking the size. At the time of writing this function I was unsure if the panic size will be always 3, but turns out it indeed is.

pkg/runner/runner.go
MaksymMalicki: The correct one is 3 for panic, changed

pkg/vm/memory/memory.go
```diff
@@ -227,6 +229,15 @@ func (memory *Memory) AllocateEmptySegment() MemoryAddress {
 	}
 }
 
+// Allocates an empty temporary segment and returns its index
+func (memory *Memory) AllocateEmptyTemporarySegment() MemoryAddress {
+	memory.TemporarySegments[uint64(len(memory.TemporarySegments))] = EmptySegment()
```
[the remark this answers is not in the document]
MaksymMalicki: got it

pkg/vm/memory/memory.go
```diff
@@ -227,6 +229,15 @@ func (memory *Memory) AllocateEmptySegment() MemoryAddress {
...
+	return MemoryAddress{
+		SegmentIndex: uint64(len(memory.TemporarySegments)),
```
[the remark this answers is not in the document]
MaksymMalicki: Yep, now after changing to the array instead of map, the length should be ok. I think it is ok to have the first segment in temporary segments with index 0. But as mentioned today during the daily, this might change as the whole system requires changes to make it work.

pkg/hintrunner/hinter/dict.go
```diff
@@ -38,19 +38,27 @@ func (d *Dictionary) InitNumber() uint64 {
 type DictionaryManager struct {
 	// a map that links a segment index to a dictionary
 	dictionaries map[uint64]Dictionary
+	// useTemporarySegments is a flag that indicates if the dictionaries should located in temporary segments, and later relocated to memory segments
```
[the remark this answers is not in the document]
MaksymMalicki: changed

pkg/parsers/starknet/starknet.go
MaksymMalicki: yep, comment left

pkg/runner/runner.go
MaksymMalicki: after the changes the comment is irrelevant. we strip the builtins from gas, segment arena and system, and hold that information in program

pkg/runner/runner.go
MaksymMalicki: got it

pkg/runner/runner.go
MaksymMalicki: done 😅 this is very nice

pkg/runner/runner.go
MaksymMalicki: did a little cleanup, moved all the runner modes into separate switch branches. Also removed the assignment of "__start__" label in AssembleProgram (which is always 0 for cairo), and hardcoded the 0 value directly in the initialization.

pkg/runner/runner.go
MaksymMalicki: no haha, it doesn't. Good catch

integration_tests/cairo1_test.go
```diff
@@ -129,10 +129,11 @@ func TestCairoFiles(t *testing.T) {
...
 	roots := []string{
-		// "./cairo_1_programs/",
-		// "./cairo_1_programs/dict_non_squashed",
-		// "./cairo_1_programs/with_input",
-		// "./cairo_1_programs/serialized_output",
+		"./cairo_1_programs/",
+		"./cairo_1_programs/dict_non_squashed",
+		"./cairo_1_programs/with_input",
+		"./cairo_1_programs/serialized_output",
+		"./cairo_1_programs/serialized_output/with_input",
```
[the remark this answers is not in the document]
MaksymMalicki: Refactored

integration_tests/cairo1_test.go
cicr99: Understood, np

cicr99 (approved):

PR #684, opened 2025-01-25
Modify the integration tests structure

MaksymMalicki:


integration_tests/.env
```diff
@@ -1,2 +1,2 @@
 # Set to run some specific file tests (ex. fib.cairo,alloc.cairo)
-INTEGRATION_TESTS_FILTERS=
\ No newline at end of file
+INTEGRATION_TESTS_FILTERS=factorial__small.cairo
```
cicr99, line 2: Is this supposed to be set already?

integration_tests/cairo0_test.go
```diff
@@ -0,0 +1,318 @@
...
+			"%s %s: %w\n%s", cliCommand, path, err, string(res),
+		)
+	}
+
+	return compiledOutput, nil
+}
+
+// given a path to a compiled cairo zero file, execute it using the
+// python vm and returns the trace and memory files location
+func runPythonVmZero(path, layout string) (time.Duration, string, string, error) {
```
cicr99, line 214: Isn't the Python VM only used for Cairo zero? You could just call this `runPythonVM` instead

cicr99 (commented):
The new layout for the integration testing looks great. It's a lot more cleaner now!

integration_tests/.env
MaksymMalicki: removed

integration_tests/cairo0_test.go
MaksymMalicki: you are right, renamed

PR #692, opened 2025-02-18
Public memory

MaksymMalicki:


pkg/runner/runner.go
```diff
@@ -918,7 +937,7 @@ func (runner *Runner) GetAirMemorySegmentsAddresses() (map[string]AirMemorySegme
 	segmentsOffsets, _ := runner.vm.Memory.RelocationOffsets()
 	memorySegmentsAddresses := make(map[string]AirMemorySegmentEntry)
 	for segmentIndex, segment := range runner.vm.Memory.Segments {
-		if segment.BuiltinRunner == nil {
+		if segment.BuiltinRunner.String() == "no builtin" {
```
cicr99, line 940: Why this change?

pkg/vm/vm.go
```diff
@@ -680,14 +680,14 @@ func DecodeMemory(content []byte) []*f.Element {
 func (vm *VirtualMachine) BuiltinsFinalStackFromStackPointerDict(builtinNameToStackPointer map[builtins.BuiltinType]uint64) error {
 
 	for segmentIndex, segment := range vm.Memory.Segments {
-		if segment.BuiltinRunner == nil {
+		if segment.BuiltinRunner.String() == "no builtin" {
```
cicr99, line 683: Same question here

pkg/runner/runner.go
MaksymMalicki: because the default runner is &NoBuiltin, with `.String()` method returning "no builtin". Isn't it quicker to compare strings, rather then interfaces?

cicr99: You are correct, for some reason I had thought it was a pointer

cicr99 (approved):

